PR #602, opened 2019-01-16
fixes #589: back on select wallet screen closes app.

### LINKED ISSUE #589: Back on wallet select screen closes app
Repro
1) Open settings screen
2) Select `switch wallet`
3) Press back

**Expected result:** return to settings screen
**Actual result:** app is closed

crchemist-ip:


cool-user:
Hi, I tested the changes and found some problems were introduced:
- When you open the app, select wallet, go to settings, and tap a wallet you want to open, it won't navigate you to that wallet
- When you open the app, select wallet, go to settings, tap switch wallet and DO NOT select wallet, but navigate back to settings and then once back to wallet detail (just tap two times back button when on select wallet screen), instead of tx list, there is an empty screen with "please wait while wallet is initialized..." text. I will get back to this why this happens.

We believe, the app should behave like this: 
- When you press back button on wallet detail screen, it should return you back to wallet selection screen
- When you press back button on wallet selection screen, it should either do nothing (the Androids back button default behavior would be overridden to do nothing in this case), or display popup that you are gonna close the app, or even close the app. But it should not return you back to the wallet you had opened before - one of the reasons being that it would need a lot of code change.

Currently when you navigate to switch wallet screen, your current opened wallet is closed via closeWallet Redux action which calls walletManager.closeWallet(). It is important to properly close opened wallet via walletManager.closeWallet() before switching to new wallet.

So if you are going to implement the go-back from wallet detail in a way that it will return you to wallet selection screen then you need to override default Android back button behavior to first close the wallet and then to navigate to wallet selection screen.

v-almonacid:
Fixed in #1029

PR #630, opened 2019-02-06
Updated FirebaseCore version for ios. Updated BundleVersion to 10.0

crchemist-ip:


ios/Podfile
```diff
@@ -1,12 +1,12 @@
...
-platform :ios, '9.0'
+platform :ios, '10.0'
 
 target 'emurgo' do
   # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
   # use_frameworks!
 
   # Pods for emurgo
-  pod 'Firebase/Core', '~> 5.14.0'
+  pod 'Firebase/Core', '~> 5.2'
```
nicarq: @crchemist-ip @SebastienGllmt  this one is a little odd, because `5.2` is way older than `5.14` and the version that was installed (Podfile.lock similar to package-lock.json) was `5.16.0`

nicarq: oh nvm

nicarq: it’s confusing that there is `Firebase/Core` and `FirebaseCore`

PR #641, opened 2019-02-25
Misc small refactoring

SebastienGllmt:
- Change max line-length to 100 (matches Yoroi-frontend)
- Remove prettier (seems redundant with eslint which does the job better)
- Organized some files (split up files into named sections so it's easier to search)
- Made a generic `backButtonNavigatorOptions` (instead of having it copied in many places)
- Remove "complex" password check (now the only check is >12 words -- same as Yoroi frontend)
- Brought in some constants from Yoroi-frontend (as opposed to magic numbers previously used) including changing "micro ADA" to "lovelace"
- Fixed some type errors in `validator.js`


SebastienGllmt:
@nicarq ran prettier as you wanted 👍

PR #645, opened 2019-02-26
Some renaming

SebastienGllmt:
- Rename `numReceiveAddresses` to `numGeneratedAddresses` (the "receive" name comes from the fact these generated address are for the receive screen, but I think the name is kind of confusing since it sounds like you're counting addresses that have actually received something as opposed to counting addresses that have been generated)
- Rename `block` to `chunk`/`addressRequestSize` (block is a bad word since it sounds like it's a "block" in a blockchain as opposed to a chunk used for batching a backend request)

**Note**: This PR changes the serialization of `AddressChain` and `State`. This means this is a breaking change and some migration of the data structure has to happen. I open this as a draft PR since right now we have no convention for how to migrate this kind of data (easiest solution is just to ad-hoc do the migration inside individual `fromJSON` functions but this would require the `fromJSON` function have access to the version of the app that generated the JSON to know when&how to apply migrations). Thoughts?

### DIFF
--- a/src/actions/history.js
+++ b/src/actions/history.js
@@ -57,7 +57,7 @@ export const mirrorTxHistory = () => (dispatch: Dispatch<any>) => {
     externalAddresses,
     confirmationCounts,
     isUsedAddressIndex,
-    numReceiveAddresses,
+    numGeneratedAddresses,
     canGenerateNewReceiveAddress,
     isEasyConfirmationEnabled,
   } = walletManager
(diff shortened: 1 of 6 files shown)


SebastienGllmt:
I changed `numGeneratedAddresses` back to `numReceiveAddresses` since it's not a big deal and I don't want to have to write code to migrate all existing wallets to a new data format just for a variable name.

Changing `blockSize` to `addressRequestSize` should be safe since presumably every single wallet is running on the default value so it's not a big deal if you discard it.

PR #654, opened 2019-03-26
Fix Receive Screen

SebastienGllmt:
The old receive screen would only show you addresses up to the first unused address in your wallet.

Now we properly show you addresses you've used as long as the gap between them is less than the bip44 gap size.

This will automatically fix all wallets already existing in production also (no extra work was needed for this)


nicarq:
@SebastienGllmt rebuilding in Bitrise. I will review it today

PR #640, opened 2019-02-19
react-intl support

crchemist-ip:

PR #664, opened 2019-04-11
Update Russian language label in the language picker

ebeloded:
image


SebastienGllmt:
Good fix :)
However this will cause a merge conflict with #640 
I would rather we just fix the same issue inside #640 instead of merging this PR then dealing with the conflict

v-almonacid:
As far as I can tell this fix has already been integrated in #640 (see commit c875167b5afd0bdc229dabf1e5269b4db1cdc152) so I'm closing it.

PR #675, opened 2019-05-01
[1 of 2] Switch to Sentry instead of Crashlytics

nicarq:


ios/emurgo/AppDelegate.m
```diff
@@ -7,29 +7,33 @@
 
 #import "AppDelegate.h"
 
-#import <Firebase.h>
 #import <React/RCTBundleURLProvider.h>
 #import <React/RCTRootView.h>
+#if __has_include(<React/RNSentry.h>)
+#import <React/RNSentry.h> // This is used for versions of react >= 0.40
+#else
+#import "RNSentry.h" // This is used for versions of react < 0.40
```
SebastienGllmt: do we need this? Our version is >= 0.4

nicarq: we actually don’t need it. it was added automatically by the Sentry script. I will get rid of it.

PR #678, opened 2019-05-17
adds Spanish support

v-almonacid:
This should be OK for review and merge but notice that I just added a flag I created myself (for the meantime). Maybe you might want a better design.


src/i18n/index.js
```diff
@@ -68,7 +70,8 @@ const numberLocales = {
   // [LANGUAGES.CHINESE_TRADITIONAL]: customNumberFmt,
   [LANGUAGES.KOREAN]: defaultNumberFmt,
   [LANGUAGES.JAPANESE]: defaultNumberFmt,
-  [LANGUAGES.RUSSIAN]: russianNumberFmt,
+  [LANGUAGES.RUSSIAN]: decimalCommaNumberFmt,
+  [LANGUAGES.SPANISH]: decimalCommaNumberFmt,
```
nicarq: @v-almonacid this seems off

`groupSeparator: '  ',` it would look like: 1 000,20 and it should look like 1.000,20

nicarq:
@v-almonacid nice work 😄 when you have a PR ready please request a review.

src/i18n/index.js
v-almonacid: Thanks for the heads up! You are right, I used the same thousands separator we were using for Russian, but it's not really common in Spanish (except in Peru it seems :P).

PR #685, opened 2019-07-05
removes ToS from state

v-almonacid:
**Summary**

This is a simple refactoring that removes the whole Terms of Service text from the app state to instead retrieve it on-demand (through the component state) when the ToS component is instantiated.

**Discussion**

Removing the ToS from the app state is not straightforward from a design perspective because the ToS are retrieved from an asynchronous call. Asynchronous calls are typically associated to actions but actions are supposed to modify the state (and paradoxically, we want to remove the ToS from the state). Another alternative might have been to use a selector (of the form `languageCode => tos`) but selectors are not meant to be asynchronous (though this can be done using the `async-selector` package, we want to avoid adding new dependencies).
The approach here is simply to perform the asyn call within the component itself, in the `componentDidMount()` method which is where async calls are recommended to be placed ([1]), and save the ToS in the component state using `setState()`. Note that even if using `setState()` in `componentDidMount()` is traditionally not recommended because it triggers two consecutive `render()`, there seems to be consensus ([2]) in that this is acceptable (second rendering occurs before the browser updates the screen).

[1]: https://reactjs.org/docs/faq-ajax.html#how-can-i-make-an-ajax-call
[2]: https://github.com/airbnb/javascript/issues/684

PS: This solution may add a small lag before the ToS are loaded, but it may just be due to the fact I was testing on an emulator.

### DIFF
--- a/src/actions/language.js
+++ b/src/actions/language.js
@@ -1,9 +1,5 @@
 // @flow
-import {
-  writeAppSettings,
-  APP_SETTINGS_KEYS,
-  loadTOS,
-} from '../helpers/appSettings'
+import {writeAppSettings, APP_SETTINGS_KEYS} from '../helpers/appSettings'
 
 import {type Dispatch} from 'redux'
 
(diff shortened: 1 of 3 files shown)


zuzunker:
@wolverineks , take a look plz when you have time. We might get this up to date with the `develop` and get it in if looks reasonable

stackchain:
Solved in #1472, therefore closing.

PR #689, opened 2019-07-15
fixes incorrect pin confirmation crash

v-almonacid:
See: https://app.clubhouse.io/emurgo/story/1714/android-1-3-incorrect-pin-confirmation-crash


nicarq:
@PavelHaliakevich can you check out the artifact in the bitrise page and check that the fix fixes the bug?

v-almonacid:
@nicarq tested on both emulator and real device. But more testing is always welcome ;)

PR #690, opened 2019-07-15
Update Fastlane scripts

nicarq:
The scripts were missing the `track` parameter that allows to set the distribution type in Google Play.

`staging` -> internal. we don’t want to confuse users.
`mainet` -> rollout. we would need to actually push the release, which is better as a second confirmation for a new release.

### DIFF
--- a/android/fastlane/Fastfile
+++ b/android/fastlane/Fastfile
@@ -25,12 +25,12 @@ platform :android do
   desc "Deploy a new staging version to Google Play"
   lane :deploy_staging do
     gradle(task: "clean assembleDevRelease")
-    upload_to_play_store
+    upload_to_play_store(track: internal)
   end
 
   desc "Deploy a new version to Google Play"
   lane :deploy_mainnet do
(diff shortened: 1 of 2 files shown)


android/fastlane/Fastfile
```diff
@@ -25,12 +25,12 @@ platform :android do
...
     gradle(task: "clean assembleDevRelease")
-    upload_to_play_store
+    upload_to_play_store(track: internal)
   end
 
   desc "Deploy a new version to Google Play"
   lane :deploy_mainnet do
     gradle(task: "clean assembleMainRelease")
-    upload_to_play_store
+    upload_to_play_store(track: rollout)
```
v-almonacid: As I understand the `rollout` track is for releasing gradually, so maybe we'd need to add the `rollout` parameter as well (it has no default value apparently [1]).

[1]: https://docs.fastlane.tools/actions/supply/#parameters

nicarq: I think the documentation is wrong 🤔 it actually has a default of 10%

PR #691, opened 2019-07-18
Fix crash when no internet connection or wrong date/time

v-almonacid:
Apparently the crash is due to a rare bug in react-native that happens when logging an error through `console.error`. This fix prevents the crash but probably will not be needed once we upgrade to react native `0.59+`.

Consider merging in case the `0.59+` upgrade is delayed for some reason.


v-almonacid:
closing as this should be covered by #692

PR #692, opened 2019-07-19
Upgrade react native to 0.59.10

v-almonacid:
This upgrade should fix several bugs, including:
- Crash due to no internet connection or incorrect time/date setting
- Rare crash with  `NoSuchKeyException` error
- Presumably some issues in Samsung devices

Additionally, this PR should also generate ABI's compatible with 64 bit devices (currently all the libraries are included in a single `.apk.`).

Pending:
- [x] Fix flow issues
- [x] Merge changes in `react-native-cardano` and update that dependency here


v-almonacid:
@nicarq looks like this PR is finally ready for review :smile:

nicarq:
LGTM 🍕

PR #694, opened 2019-07-25
fixes vulnerability alerts

v-almonacid:
Note:
- Vulnerability in `mini_magick` has been addressed by bumping fastlane to version 2.127.2 (see https://github.com/fastlane/fastlane/issues/15044)
- `simple-markdown` is a nested dependency from [`react-native-simple-markdown`](https://github.com/TitanInvest/react-native-easy-markdown). Unfortunately we are already in the latest version so either we fork and update the dependency or we wait for the original repo to address the problem (not likely to be done soon since no activity for ~5 months)

pending:
- [x] simple-markdown
- [x] mini-magick

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 # Gemfile
 source "https://rubygems.org"
-gem 'fastlane'
+gem 'fastlane', '2.127.2'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -14,11 +14,11 @@ GEM
     declarative (0.0.10)
     declarative-option (0.1.0)
     digest-crc (0.4.1)
-    domain_name (0.5.20180417)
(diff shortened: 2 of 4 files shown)


v-almonacid:
Update: It doesn't look like the `react-native-simple-markdown` team is going to do something about it soon. I contacted one of the devs without response. Should we consider a fork @nicarq or ignore the issue?

SebastienGllmt:
I looks like the `simple-markdown` is specified using a caret so I think you should be able to just manually bump the version used inside our `yarn.lock` for this package and everything will work.

v-almonacid:
Actually what happens is that `yarn` will found out that `react-native-easy-markdown` is already in its latest version, so it won't look further for new releases in nested dependencies.

PR #696, opened 2019-07-26
Yoroi Mobile 1.4

nicarq:
Yoroi Mobile 1.4

Besides the bug fixes, the end user shouldn’t notice any difference between the previous version and this one.

- Upgrade to React Native 0.59.10
- Upgrade to Flow 0.92
- Fix crash when no internet connection or wrong date/time
- Fixes incorrect pin confirmation crash
- Adds 64bit
- Android Target Sdk Version 26 -> 27


nicarq:
@v-almonacid I uploaded some builds before for testing fastlane and the version numbers can’t be the same.

PR #708, opened 2019-08-13
Disable screenshots on non-sensitive screen

v-almonacid:
Screenshots can be disabled on any given component by adding something like:
```
componentDidMount() {
    FlagSecure.activate()
  }

  componentWillUnmount() {
    FlagSecure.deactivate()
  }
```

The helper functions `FlagSecure.activate()` and `FlagSecure.deactivate()` are very simple and are written in native code (currently using a fork from third-party package, but we may just rewrite them and include them directly in our codebase to avoid an additional dependency).

The only problem I see is that most components are written as functional components, in which case the lifecycle methods are not directly available; so some non-negligible refactoring might be needed.

Note: this is just a PoC and is not ready for merge. Currently only disables screenshots in send screen.

### DIFF
--- a/src/components/Send/SendScreen.js
+++ b/src/components/Send/SendScreen.js
@@ -50,6 +50,8 @@ import type {
 } from '../../utils/validators'
 import type {ComponentType} from 'react'
 
+import FlagSecure from 'react-native-flag-secure-android';
+
 const amountInputErrorMessages = defineMessages({
   INVALID_AMOUNT: {
     id: 'components.send.sendscreen.amountInput.error.INVALID_AMOUNT',
@@ -227,6 +229,7 @@ class SendScreen extends Component<Props, State> {
       this.handleAmountChange(CONFIG.DEBUG.SEND_AMOUNT)
     }
     this.props.navigation.setParams({onScanAddress: this.handleAddressChange})
+    FlagSecure.activate()
   }
 
   componentDidUpdate(prevProps, prevState) {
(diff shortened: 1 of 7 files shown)


src/components/Send/SendScreen.js
```diff
@@ -227,6 +229,7 @@ class SendScreen extends Component<Props, State> {
       this.handleAmountChange(CONFIG.DEBUG.SEND_AMOUNT)
     }
     this.props.navigation.setParams({onScanAddress: this.handleAddressChange})
+    FlagSecure.activate()
```
nicarq, line 232: I think this way may lead to a lot of bugs. As I understood `componentDidMount` is `is invoked immediately after a component is mounted (inserted into the tree).` If the component is not removed from the stack and another screen is pushed, it will not deactivate it. We basically want to activate/ deactivate screens based if they are been shown, rather than when they were loaded (sometimes is the same).

I think we want to resemble something similar to `viewWillAppear` or `viewDidAppear` from iOS.

v-almonacid: You are right! So I did further research on this and it seems we might achieve what you suggest through the `ReactNavigation` API: https://reactnavigation.org/docs/en/navigation-events.html

What do you think of this approach?

nicarq: I think it makes sense! I couldn’t verify what `willFocus` and `didFocus` do / behave. The links seem broken https://facebook.github.io/react-native/docs/0.43/navigator.md#onwillfocus 🤔

v-almonacid:
closing in favor of the approach in #1265

PR #709, opened 2019-08-14
add build version, network and commit hash in settings

v-almonacid:
This is currently how this PR look like (without much effort trying to match the specified design).

image

Notes for the reviewer:

-  In particular, please check the way the commit hash is retrieved (through the `get_commit.sh` script and then using ` react-native-config` to read the commit hash as a config variable . I'd have loved to find a cleaner way. Perhaps just hard coding it the same way the version number is specified?
- This PR adds an additional dependency (`react-native-device-info`) to get the version number.


nicarq (changes requested):
the UI looks good! but I would make it part of the building process, so `get_commit` is executed in every build. I maybe I could do the setup for iOS and you for Android?

package.json
```diff
@@ -3,6 +3,12 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
+    "android-mainnet-debug": "yarn run pre-packager && react-native run-android --variant=mainDebug",
```
nicarq: I wouldn’t use this, because it just works with this specific scripts :-/ for example we use fastlane for deployments and it wouldn’t work with that

package.json
```diff
@@ -3,6 +3,12 @@
...
+    "android-testnet-debug": "yarn run pre-packager && react-native run-android --variant=devDebug",
+    "android-release": "yarn run pre-packager && cd android && ./gradlew assembleMainRelease",
+    "ios-mainnet-debug": "yarn run pre-packager && react-native run-ios --scheme=emurgo-staging --configuration=Staging.Debug",
+    "ios-testnet-debug": "yarn run pre-packager && run-ios --scheme=emurgo --configuration=Debug",
+    "pre-packager": "./get_commit.sh",
```
nicarq: this should be part of `Gradle` and `Xcode` rather than `package.json`

package.json
nicarq: So basically, I would focus in making it work for the `fastlane` deployment scripts

v-almonacid:
Right! Actually I forgot to mention that this was only a proposal and that the releasing process was not addressed yet. All the part related to getting the commit hash seemed a bit too hacky and I wanted to have your feedback before going further.

v-almonacid:
@nicarq it took me a while to figure out but I think it's much better now. Basically every Gradle task of type release or debug install will be executed after a `:app:getCurrentCommitHash` task. So this  task will be executed before every release with fastlane and also before every `react-native run-android ...` build. What do yo think?

Let me know if you are going to take care of the iOS part, or I'll do it ;)

nicarq:
@v-almonacid that sounds pretty good! Could you take a look at iOS? I started doing some stuff, but I didn’t finish 😁 I can explain what I learn along the way

v-almonacid:
@nicarq this PR should be now ready for final review.

nicarq (approved):

PR #718, opened 2019-08-29
Security upgrades

v-almonacid:
- Bump `mixin-deep` from `1.3.1` to `1.3.2`
- Bump `set-value` from `2.0.0` to `2.0.1`

Note we had two versions of `set-value`, `2.0.0` and `0.4.3`, the latter was required by `union-value`, so upgrading this one resulted in only one version of `set-value` being required.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -8202,9 +8202,9 @@ set-value@^0.4.3:
     to-object-path "^0.3.0"
 
 set-value@^2.0.0:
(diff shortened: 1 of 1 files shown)


v-almonacid:
@SebastienGllmt sorry I mistakenly requested your review (but you are always welcome for reviewing anyway :smile: )

SebastienGllmt (approved):

PR #721, opened 2019-10-08
fix jest snapshot

v-almonacid:
Minor fix.

Context: Some tests are based on a snapshot of the Testnet which includes date fields that may slightly change in case the backend is restarted. These should be ignored.


rcmorano (approved):
LGTM!

PR #722, opened 2019-10-26
replace rust encryption/decryption by js module emip3js

v-almonacid:
The new Shelley rust libraries do not include encryption/decryption functions (due to design reasons), so these are now implemented in JS (see [emip3js](https://github.com/Emurgo/emip3js)). This PR just replaces the rust functions by the new JS functions.

Note: base branch is Shelley and not develop


src/crypto/customPin.js
```diff
@@ -2,10 +2,7 @@
 import {encryptData, decryptData} from './util'
 import {WrongPassword} from './errors'
 
-export const encryptCustomPin = async (
-  installationId: string,
-  pin: string,
-): Promise<void> => {
+export const encryptCustomPin = async (installationId: string, pin: string) => {
```
v-almonacid, line 5: (nothing changed here)

nicarq (changes requested):
this seems good to me, but I would like to be 100% that we are compatible with previous encryptions -- if not we would do a migration. Could u write a test to validate this?

v-almonacid:
@nicarq You are right, I checked this locally by compiling the rust implementation and checking if both JS and rust were compatible. Just added a unit test 😉

nicarq (approved):

PR #723, opened 2019-10-29
add react-native-chain-libs (new shelley rust wrappers)

v-almonacid:
This PR changes some setting in RN in order to support the new wrappers. Basically, we need to migrate to Android X using this tool https://www.npmjs.com/package/jetifier).


ios/Podfile
```diff
@@ -5,6 +5,8 @@ target 'emurgo' do
   # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
   # use_frameworks!
 
+  pod 'react-native-chain-libs', :path => '../node_modules/react-native-chain-libs'
```
nicarq: this shouldn’t work because it uses React as a dependency so it’s going to install React again using Cocopoads. The library should be added manually.

nicarq: this shouldn’t work because it uses React as a dependency so it’s going to install React again using Cocopoads. The library should be added manually.

nicarq (changes requested):

PR #724, opened 2019-10-30
Implementing BIP-21 (EmIP-02) standard QRcode scanning.

aihanrashidov:
If an amount is included in the QR code it will be automatically filled in the amount field.

### DIFF
--- a/src/components/Send/SendScreenNavigator.js
+++ b/src/components/Send/SendScreenNavigator.js
@@ -1,25 +1,42 @@
 // @flow
-
 import React from 'react'
-import {Button} from '../UiKit'
 import {createStackNavigator} from 'react-navigation'
+
+import {Button} from '../UiKit'
 import SendScreen from './SendScreen'
 import ConfirmScreen from './ConfirmScreen'
 import AddressReaderQR from './AddressReaderQR'
 import BiometricAuthScreen from './BiometricAuthScreen'
 import iconQR from '../../assets/img/qr_code.png'
-
+import {pastedFormatter} from './amountUtils'
 import HeaderBackButton from '../UiKit/HeaderBackButton'
 import {
   defaultNavigationOptions,
   defaultStackNavigatorOptions,
 } from '../../navigationOptions'
-
 import {SEND_ROUTES} from '../../RoutesList'
 
 import styles from './styles/QrButton.style'
 
-import {pastedFormatter} from './amountUtils'
+const getParams = (params) => {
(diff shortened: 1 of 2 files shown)


src/components/Send/SendScreenNavigator.js
```diff
@@ -30,9 +32,22 @@ const SendScreenNavigator = createStackNavigator(
             style={styles.qrButton}
             onPress={() =>
               navigation.navigate(SEND_ROUTES.ADDRESS_READER_QR, {
-                onSuccess: (address) => {
-                  const handler = navigation.getParam('onScanAddress')
-                  handler && handler(address)
+                onSuccess: (stringQR) => {
+                  const regex = /(cardano):([a-zA-Z1-9]\w+)\?/
```
v-almonacid: could you please change the URL to match our spec? It should be `web+cardano`, for more info please check: https://github.com/Emurgo/EmIPs/blob/master/specs/emip-002.md#core-implementation

v-almonacid: Also Byron cardano addresses are encoded in `base58` so it'd be better to check `[A-HJ-NP-Za-km-z1-9]`

src/components/Send/SendScreenNavigator.js
```diff
@@ -30,9 +32,22 @@ const SendScreenNavigator = createStackNavigator(
...
+
+                  if (regex.test(stringQR)) {
+                    const address = stringQR.match(regex)[2]
+                    const index = stringQR.indexOf('?')
+                    const params = getParams(stringQR.substr(index))
+                    if ('amount' in params) {
+                      setAddress(address, navigation)
```
v-almonacid: Before setting the address and amount here, could you please validate them using these? 🙏 
https://github.com/Emurgo/yoroi-mobile/blob/ead9ca1a6f75601a0fc4305b1c89a74fd8c97491/src/utils/validators.js#L135

v-almonacid (changes requested):

src/components/Send/SendScreenNavigator.js
SebastienGllmt: I would keep it `a-zA-Z1-9` and do validation of address-specific rules after this regex passes because v2 validation rules are different than Shelley validation rules so it would be weird to duplicate this logic inside the regex.

v-almonacid (approved):

v-almonacid (approved):

PR #725, opened 2019-10-31
Add shelley accounting transactions

v-almonacid:
This PR adds the minimum to create accounting transaction (I'll be trying to reuse code from yoroi-frontend as much as possible and this is an example).

Also, this PR adds `js-chain-libs-node` and sets jest to use it for unit testing.

Note: https://github.com/Emurgo/yoroi-mobile/pull/723 should be merged first


src/crypto/shelley/accountingTransactions.js
```diff
@@ -0,0 +1,136 @@
...
+  PrivateKey,
+  PublicKey,
+  Witness,
+  SpendingCounter,
+  Hash,
+  Account,
+  AuthenticatedTransaction,
+} from 'react-native-chain-libs'
+
+// TODO: move to external config file
```
nicarq, line 22: Agree. I would just use `config.js` because it also has the protocol magic

src/crypto/shelley/accountingTransactions.js
```diff
@@ -0,0 +1,136 @@
...
+  addressDiscrimination: {
+    production: '0',
+    test: '1',
+  },
+  genesisHash:
+    'adbdd5ede31637f6c9bad5c271eec0bc3d0cb9efb86a5b913bb55cba549d0770',
+}
+
+// TODO: define a transaction type
+export const buildTransaction = async (
```
nicarq, line 38: maybe we could be explicit about being Unsigned so it could be: `buildUnsignedTransaction`

src/crypto/shelley/accountingTransactions.js
```diff
@@ -0,0 +1,136 @@
...
+  sourceAccount.free()
+  return tx
+}
+
+export const signTransaction = async (
+  // unsignedTx: Transaction, // TODO
+  unsignedTx: any,
+  accountCounter: number,
+  accountPrivateKey: PrivateKey,
+): AuthenticatedTransaction => {
```
nicarq, line 126: Ugh I think `AuthenticatedTransaction` was changed already

nicarq: actually the PR is still open https://github.com/input-output-hk/chain-libs/pull/128

SebastienGllmt: yes the whole transaction format is undecided at this point

src/crypto/shelley/utils.js
```diff
@@ -0,0 +1,35 @@
...
+// import {Transaction} from 'react-native-chain-libs'
+import {BigNumber} from 'bignumber.js'
+
+export function getTxInputTotal(
+  // tx: Transaction,
+  tx: any,
+): BigNumber {
+  let sum = new BigNumber(0)
+
+  const inputs = tx.inputs()
```
nicarq, line 13: this would look much cleaner with `tx.inputs().reduce(...)`

src/crypto/shelley/utils.js
```diff
@@ -0,0 +1,35 @@
...
+  return sum
+}
+
+export function getTxOutputTotal(
+  // tx: Transaction,
+  tx: any,
+): BigNumber {
+  let sum = new BigNumber(0)
+
+  const outputs = tx.outputs()
```
nicarq, line 28: we could also use `reduce` here

src/crypto/shelley/utils.js
SebastienGllmt: you can't do this because `tx.inputs` is not iterable. It's a webassembly class with two methods: `size` and `get`.

src/crypto/shelley/accountingTransactions.js
v-almonacid: yep I'm aware of this. Hopefully the libs won't change too much and we can still continue to integrate code even if there will be some refactoring after

vincenthz: the transaction format is very much defined, this is not expected to change in any fundamental way apart from at the code level to finish to implement what was already documented in the format file. I'm not saying the format is stable just now, but really close to be.

nicarq (approved):

PR #730, opened 2019-11-13
initial balance check flow

v-almonacid:
This PR adds the following UIs:

image

image

image


src/i18n/locales/en-US.json
```diff
@@ -159,6 +159,14 @@
   "components.txhistory.txnavigationbuttons.receiveButton": "Receive",
   "components.txhistory.txnavigationbuttons.sendButton": "Send",
   "components.uikit.offlinebanner.offline": "You are offline. Please check settings on your device.",
+  "components.walletinit.balancecheck.balancecheckmodal.attention": "Attention",
+  "components.walletinit.balancecheck.balancecheckmodal.attentionDescription": "The balance check executed successfully, and we were able to match your wallet with the balance displayed below. Remember that the balance displayed should only match the one that you <strong>had on November 12th</strong>.",
```
nicarq, line 163: @v-almonacid it’s not displaying `html` maybe there is another way to do **bold**

src/components/WalletInit/BalanceCheck/BalanceCheckModal.js
```diff
@@ -0,0 +1,80 @@
...
+import {compose} from 'redux'
+import {withHandlers} from 'recompose'
+import {View} from 'react-native'
+import {injectIntl, defineMessages, intlShape} from 'react-intl'
+
+import type {Navigation} from '../../../types/navigation'
+import {Text, Button, Modal} from '../../UiKit'
+import {ROOT_ROUTES} from '../../../RoutesList'
+
+// import styles from './styles/AddressModal.style'
```
nicarq, line 13: we could just remove this

src/components/WalletInit/BalanceCheck/BalanceCheckModal.js
```diff
@@ -0,0 +1,80 @@
...
+  onRequestClose: () => any,
+  intl: intlShape,
+  navigation: Navigation,
+}
+
+export default injectIntl(
+  (compose(
+    withHandlers({
+      navigateWalletInit: ({navigation}) => (event) =>
+        navigation.navigate(ROOT_ROUTES.WALLET),
```
nicarq, line 77: there is a crash here but also we shouldn’t navigate to a previous location, but rather we should pop as many times as necessary to take screens out of the navigation stack -- rather than add more to it

src/i18n/locales/en-US.json
v-almonacid: @nicarq yep, resolved in next PR

src/components/WalletInit/BalanceCheck/BalanceCheckModal.js
v-almonacid: @nicarq got it !

v-almonacid:
I will just merge this as it is an intermediary draft PR.

PR #739, opened 2019-11-27
fix general error handling

v-almonacid:
Many times when an `onunhandledrejection` event occurs (there is a listener registered at app startup in `src/index.js`) and the app crashes, it is not easy to debug because the error message, both in console logs and the user interface, is hidden in a pile of `'intl.formatMessage is not a function (...)`. This is because of a bug in the function that reports the error, `handleGeneralError`, which was calling `showErrorDialog()` without providing the `intl` object.

To fix this, we may just call `showErrorDialog()` with `intl=null`, as providing a valid object is not possible at the stage in which the `onunhandledrejection` event listener is registered.

Now, after that is fixed, the app crashes again because apparently that was the intended behaviour.   `handleGeneralError` calls `Sentry.crash()`, which throws, but the problem is that it keeps throwing forever, polluting once again the logs.

So **unless there is a good reason for that, I propose we remove this behaviour.** If this fix is accepted, then I also propose to apply it on `shelley`, as that may help for debugging unexpected errors.

### Reproduction steps
Simply call `handleGeneralError()` somewhere:
```
const e = {title: 'title', message: 'message'}
handleGeneralError('Could not submit transaction', e)
```

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -385,9 +385,10 @@ const showDialog = (translations: DialogOptions): Promise<DialogButton> =>
   })
 
 export const showErrorDialog = (
-  dialog: DialogOptions,
+  dialog: Object,
   intl: ?intlShape,
-  msgOptions?: any,
+  msgOptions?: Object,
+  // $FlowFixMe
 ): Promise<DialogButton> => {
   let title, message, yesButton
   if (intl != null) {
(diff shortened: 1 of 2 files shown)


nicarq:
@v-almonacid aha now I see what you meant. This is awesome because it will *help* a lot with debugging :-)

nicarq (approved):

PR #744, opened 2019-11-29
add sendAllUnsignedTx function

v-almonacid:


src/crypto/shelley/transactions/utxoTransactions.js
```diff
@@ -271,36 +272,126 @@ async function utxoToTxInput(utxo: RawUtxo): Input {
...
     witnesses.add(witness)
   }
   return await builderSetWitnesses.set_witnesses(witnesses)
 }
+
+export const sendAllUnsignedTxFromUtxo = async (
+  receiver: string,
+  allUtxos: Array<RawUtxo>,
+): Promise<V3UnsignedTxData> => {
+  const totalBalance = allUtxos
```
nicarq, line 314: @v-almonacid probably we should throw if there isn’t any UTXO too. I don’t know if that could happen

nicarq (approved):
Looks good. I only left that one comment that could be helpful as a precaution

src/crypto/shelley/transactions/utxoTransactions.js
v-almonacid: you mean if `allUtxos = []`? in that case `totalBalance` should be zero.

PR #747, opened 2019-12-02
V almonacid/ch2304/bump rn chain libs and fix utxo txs

v-almonacid:
This PR mainly includes additions from https://github.com/Emurgo/yoroi-frontend/pull/1147/files
but also a few things from:
- https://github.com/Emurgo/yoroi-frontend/pull/1144
- https://github.com/Emurgo/yoroi-frontend/pull/1177


src/crypto/byron/util.test.js
```diff
@@ -273,8 +273,8 @@ describe('signTransaction', () => {
     // If this fails, it means we let this bug open for too long time.
     // Try updating rust library and seeing if it disappears
     // Note(v-almonacid): I increased this timer. We'll hopefully update the
-    // bindings within the the next months and this issue should be fixed by then©
-    expect(moment().isBefore('2019-12-01')).toBeTruthy()
+    // bindings within the the next months and this issue should be fixed by then
+    expect(moment().isBefore('2019-12-12')).toBeTruthy()
```
nicarq: @v-almonacid I would aim to 15 or something like that

src/crypto/shelley/transactions/utils.js
```diff
@@ -1,10 +1,22 @@
...
+} from 'react-native-chain-libs'
+import type {BaseSignRequest} from '../../../types/HistoryTransaction'
+import {CONFIG} from '../../../config'
+
 import {BigNumber} from 'bignumber.js'
 
-export const getTxInputTotal = async (IOs: InputOutput): Promise<BigNumber> => {
+export const getTxInputTotal = async (
+  IOs: InputOutput,
+  shift: boolean,
```
nicarq: I’m not in love with the naming here. I would make it more obvious as passing an enum to specify `ADA` or `Lovelace` the shift is too low level

src/crypto/shelley/transactions/utils.js
```diff
@@ -27,19 +43,119 @@ export const getTxOutputTotal = async (
...
+      if (changeAddrs.has(addr)) {
+        continue
+      }
+    }
+    receivers.push(addr)
+  }
+  return receivers
+}
+
+export const shelleyTxEqual = async (
```
nicarq: is this actually used somewhere? if it’s, also I would change the name to `IsShelleyTxEqual` or something like that

src/crypto/shelley/transactions/utils.js
```diff
@@ -27,19 +43,119 @@ export const getTxOutputTotal = async (
...
+  switch (await certificate.get_type()) {
+    // TODO: maybe should be `await CertificateKind.StakeDelegation`
+    case StakeDelegation: {
+      return await PayloadAuthData.for_stake_delegation(
+        await StakeDelegationAuthData.new(bindingSignature),
+      )
+    }
+    default:
+      throw new Error(
+        `generateAuthData unexptected cert type ${await certificate.get_type()}`,
```
nicarq: small typo

src/crypto/shelley/transactions/utils.js
SebastienGllmt: it comes from Yoroi-Frontend where we need to compare if two txs for some functionality that doesn't exist in Yoroi Mobile I think

src/crypto/byron/util.test.js
v-almonacid: Actually it's going to be more because at the time I wrote that comment we were thinking in going from v1 to v2 but now there is no much sense on doing that, and I assume we'll keep using v1 and v3 and eventually just v3.

src/crypto/shelley/transactions/utils.js
v-almonacid: Good point. We don't use, I ported this code from frontend and thought we might eventually need it as well. I'm removing it for now.

nicarq (approved):

PR #761, opened 2020-01-27
connect staking UI to API

v-almonacid:


src/actions/account.js
```diff
@@ -0,0 +1,88 @@
...
+  }
+  dispatch(_clearAccountState())
+  dispatch(_startFetching())
+  try {
+    // $FlowFixMe (v-almonacid): not sure how to correctly handle union types
+    const accountStateResp = await walletManager.fetchAccountState()
+    const accountState = Object.keys(accountStateResp).map(
+      (key) => accountStateResp[key],
+    )[0]
+    if ('error' in accountState) throw new Error(accountState.error)
```
SebastienGllmt: Careful: in case of an error, the value for the <key, value> pair can return `""` or `"Account does not exist`" 

https://github.com/Emurgo/yoroi-frontend/pull/1388

src/actions/pools.js
```diff
@@ -0,0 +1,77 @@
...
+  try {
+    const pools = getState().accountState.delegation.pools
+    // $FlowFixMe
+    const poolInfoResp = await walletManager.fetchPoolInfo({
+      ids: pools.map((pool) => pool[0]),
+    })
+    const poolInfo = Object.keys(poolInfoResp).map(
+      (key) => poolInfoResp[key],
+    )[0]
+    if ('error' in poolInfo) throw new Error(poolInfo.error)
```
SebastienGllmt: This fixes the Flow issue so you can remove the `// $FlowFixMe`
```suggestion
    if (poolInfo.error != null) throw new Error(poolInfo.error)
```

src/actions/account.js
SebastienGllmt: This fixes the Flow issue so you can remove the `// $FlowFixMe`
```suggestion
    if (accountState.error != null) throw new Error(accountState.error)
```

src/components/Delegation/DelegationCenter.js
```diff
@@ -35,28 +38,27 @@ const prepareStakingURL = (
 ): null | string => {
   // eslint-disable-next-line max-len
   // Refer: https://github.com/Emurgo/yoroi-frontend/blob/2f06f7afa5283365f1070b6a042bcfedba51646f/app/containers/wallet/staking/StakingPage.js#L60
-
   // source=mobile is constant and already included
-  const finalURL = CARDANO_CONFIG.SHELLEY.SEIZA_STAKING_SIMPLE(userAda)
-  // finalURL += `&delegated=${encodeURIComponent(JSON.stringify(poolList))}`
-
+  let finalURL = CARDANO_CONFIG.SHELLEY.SEIZA_STAKING_SIMPLE(userAda)
+  if (poolList != null) {
```
SebastienGllmt, line 50: Although it doesn't work right now, we will add a `locale` parameter later. Maybe add a TODO for this?

src/components/Delegation/DelegationConfirmation.js
```diff
@@ -23,14 +45,155 @@ const messages = defineMessages({
...
+  },
+  amountLabel: {
+    // TODO: add to global messages
+    id: 'components.send.amountfield.label',
+    defaultMessage: '!!!Amount',
+    description: 'some desc',
+  },
+  feeLabel: {
+    // TODO: add to global messages
+    id: 'components.send.amountfield.label',
```
SebastienGllmt: Both ID and defaultMessage here are wrong (copy paste mistake). It should be `fee`  since it's `feeLabel`

src/components/Delegation/DelegationSummary.js
```diff
@@ -61,75 +78,228 @@ const SyncErrorBanner = injectIntl(({intl, showRefresh}) => (
...
+    ) {
+      this.props.fetchPoolInfo()
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.intervalId != null) clearInterval(this.intervalId)
+  }
+
+  intervalId: void | IntervalID
```
SebastienGllmt, line 122: FYI in the extension I moved all of this out of the UI and moved it to a `TimeStore` (https://github.com/Emurgo/yoroi-frontend/blob/shelley/app/stores/ada/TimeStore.js)

The benefit of putting it in the store is that now the UI can react to the currentTime, `currentEpoch` and `currentSlot` changing. Since only the epoch countdown needs to know the current time, it can react to that separately and the rest of the page can react to the epoch only. The downside with this `intervalId` solution was that it causes a re-draw of the whole component every second

src/components/Delegation/DelegationConfirmation.js
v-almonacid: actually that message is not being used, deleting...

src/components/Delegation/DelegationSummary.js
```diff
@@ -61,75 +78,228 @@ const SyncErrorBanner = injectIntl(({intl, showRefresh}) => (
...
+          <OfflineBanner />
+          {isOnline &&
+            lastSyncError && <SyncErrorBanner showRefresh={!isSyncing} />}
+
+          <ScrollView style={styles.inner}>
+            {poolInfo == null && <NotDelegatedInfo />}
+            {/* eslint-disable indent */
+            poolInfo != null &&
+              false && (
+                /* TODO */
```
SebastienGllmt: TODO

src/state.js
```diff
@@ -31,6 +36,16 @@ export type State = {
...
   },
+  accountState: {
+    isFetching: boolean,
+    lastFetchingError: any,
+    ...AccountState,
+  },
+  poolInfo: {
+    isFetching: boolean,
+    lastFetchingError: any,
+    meta: any, // TODO(v-almonacid): type me
```
SebastienGllmt, line 49: TODO (although I guess we can do this later)

SebastienGllmt (approved):

PR #770, opened 2020-02-13
add warning modal for debug wallets

v-almonacid:
Modal is added right after a wallet is selected, within the TxHistory screen. Accepting the modal sends user to WalletSelectionScreen.

Edit: modal is also added in Delegation dashboard. In both cases it's triggered if wallet corresponds to any of the two default wallets used for debugging.


nicarq (approved):
🙌

src/crypto/wallet.js
```diff
@@ -1172,6 +1172,36 @@ class WalletManager {
...
+
+    for (const mnemonic of mnemonics) {
+      Logger.debug('WalletManager::checkForFlawedWallets mnemonic:', mnemonic)
+      const flawedAddresses = await util.getAddressesFromMnemonics(
+        mnemonic,
+        'External',
+        [0, 1, 2, 3, 4],
+        CARDANO_CONFIG.MAINNET,
+      )
+      for (let i = 0; i < flawedAddresses.length; i++) {
```
SebastienGllmt: is this not a little excessive? Can't you just do
```javascript
    if (this._wallet == null) throw new WalletClosed()
    if (this._wallet._externalChain.isMyAddress(flawedAddresses[0])) {
        return true;
    }
```

src/crypto/wallet.js
```diff
@@ -1172,6 +1172,36 @@ class WalletManager {
       : CONFIG.CARDANO
     return await api.getPoolInfo(pool, config)
   }
+
+  async checkForFlawedWallets() {
```
SebastienGllmt: Remember to add type annotations for function return type

src/crypto/wallet.js
v-almonacid: Yes we can

nicarq (approved):

PR #774, opened 2020-02-17
fix delegation confirmation (wrong password) error handling

v-almonacid:
Previously when a user entered a wrong password to confirm a delegation tx, the app responded with a generic "Unexpected error: request operation failed... could not submit transaction", triggering a `console.error` and reporting a crash to sentry.

This PR fixes this and now the app should show the following dialog instead:

Screenshot 2020-02-17 at 09 43 05

(note: this commit also prettifies a few unrelated lines of code)

### DIFF
--- a/src/components/Delegation/DelegationConfirmation.js
+++ b/src/components/Delegation/DelegationConfirmation.js
@@ -29,6 +29,7 @@ import {NetworkError, ApiError} from '../../api/errors'
 import {WrongPassword} from '../../crypto/errors'
 import walletManager, {SystemAuthDisabled} from '../../crypto/wallet'
 import KeyStore from '../../crypto/KeyStore'
+import {KNOWN_ERROR_MSG} from '../../crypto/byron/util'
 
 import styles from './styles/DelegationConfirmation.style'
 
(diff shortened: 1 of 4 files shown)


src/components/Delegation/DelegationConfirmation.js
```diff
@@ -74,7 +75,7 @@ const messages = defineMessages({
     description: 'some desc',
   },
   delegationTxSignError: {
-    id: 'components.stakingcenter.delegationTxSignError',
```
SebastienGllmt, line 77: why are you changing this? It breaks all translations

SebastienGllmt (changes requested):

src/components/Delegation/DelegationConfirmation.js
v-almonacid: https://github.com/Emurgo/yoroi-mobile/blob/ea6ee0f74b34edf55ad9ef1081462a731cf54395/src/i18n/locales/en-US.json#L148

SebastienGllmt (approved):

PR #775, opened 2020-02-17
fix wallet upgrade

v-almonacid:
This should hopefully fix the bug when sending the wallet transfer tx.

note: Ignore flow and lint errors, should be fixed in previous PR.

### DIFF
--- a/src/components/WalletInit/RestoreWallet/WalletCredentialsScreen.js
+++ b/src/components/WalletInit/RestoreWallet/WalletCredentialsScreen.js
@@ -90,6 +90,7 @@ type Props = {
   intl: any,
   navigation: Navigation,
   createWallet: (string, string, string, boolean) => any,
+  submitShelleyTx: (Uint8Array) => any,
 }
 
 type State = {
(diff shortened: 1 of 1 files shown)


src/components/WalletInit/RestoreWallet/WalletCredentialsScreen.js
```diff
@@ -273,9 +274,8 @@ class WalletCredentialsScreen extends React.Component<Props, State> {
       this.setState({isProcessing: false})
       navigation.navigate(ROOT_ROUTES.SHELLEY_WALLET)
     } catch (e) {
-      handleApiError(e, intl, 'Could not upgrade wallet')
-    } finally {
       this.setState({isProcessing: false})
+      handleApiError(e, intl, 'Could not upgrade wallet')
```
v-almonacid, line 278: This is a small side fix. I just removed the `finally` statement because when there is no error, the component is unmounted and `setState` would try to update the state of an unmounted component.

SebastienGllmt (approved):

nicarq (commented):
@v-almonacid I didn’t understand why `submitShelleyTx` is required. It’s added but it doesn’t seem to be used 🤔

PR #780, opened 2020-02-19
refactor delegation summary data fetching and remove please wait modal

v-almonacid:
I'm temporarily removing the "please wait modal" which is used in iOS during wallet loading.

The spinner is still setup and works in some scenarios (eg from settings -> back button or after pull to refresh), but doesn't seem to work after wallet selection or wallet switch.


v-almonacid:
closing this one as will be replaced by https://github.com/Emurgo/yoroi-mobile/pull/782

PR #782, opened 2020-02-20
remove please wait dialog plus a few minor modifications

v-almonacid:
As discussed, we will just remove the dialog for iOS.

There are some other minor modifications though (things that I saved from #780, which is deprecated in favor of this PR).


src/components/Delegation/DelegationSummary.js
```diff
@@ -162,16 +162,16 @@ class DelegationSummary extends React.Component<Props, State> {
...
     if (this._firstFocus) {
       this._firstFocus = false
       // skip first focus to avoid
-      // didMount -> refetch -> done -> didFocus -> refetch
+      // didMount -> fetchPoolInfo -> done -> didFocus -> fetchPoolInfo
       // blinking
       return
     }
     this.props.checkForFlawedWallets()
-    this.props.fetchPoolInfo()
```
v-almonacid, line 174: this is just redundant, it's triggered in didUpdate

src/components/Delegation/DelegationSummary.js
```diff
@@ -257,7 +256,10 @@ class DelegationSummary extends React.Component<Props, State> {
             style={styles.inner}
             refreshControl={
               <RefreshControl
-                onRefresh={fetchAccountState}
+                onRefresh={() => {
+                  fetchUTXOs()
```
v-almonacid: Here I had forgot to fetch utxos as well when user pulls to update

SebastienGllmt: is it okay not to block on these calls? I assume yes but just want to confirm you didn't forget to await

v-almonacid: yes they can be async

nicarq: I was going to ask the same thing. That I thought that maybe this could be an await. Maybe the issues with the loading dialog could be related to multiple calls turning it off / on?

nicarq (approved):

src/components/Delegation/DelegationSummary.js
v-almonacid: No these are the calls generated when the user pulls to refresh, in which case the spinner works just fine for both Android and iOS.

It just doesn't work in the initial loading, in which case the calls are triggered inside a `ComponentDidMount` or after the component is focused.

SebastienGllmt: Okay maybe in that case you can do something like
```
onRefresh={() => Promise.all(fetchUTXOs(), fetchAccountState())}
```
should behave exactly the same way but it won't be considered an error if ever `onRefresh` is refactored to have special handling for async functions

v-almonacid: Ok I added this too (it was failing but it's because `Promise.all` requires an iterable)

PR #785, opened 2020-02-24
add nightly flavor

v-almonacid:
This PR simply adds a new product flavor for Android called _nightly_ 

Recall: Android distinguishes _product flavours_ from _build types_. We have 3 product flavours (mainnet, dev and now nightly) and 2 build types (the default ones: release and debug).

To build the app using the new nightly product flavor and the debug build type:
```
react-native run-android --variant=nightlyDebug
```

## Small UI change
To distinguish between different flavours installed in your phone I changed the app name to _yoroi-nightly_ as well as the icon color and the initial screen color (what you see right after opening the app). I only changed this because changing the whole UI theme requires more work.
 
Screenshot 2020-02-24 at 21 36 44

Screenshot 2020-02-24 at 21 13 35


android/app/build.gradle
```diff
@@ -165,6 +166,15 @@ android {
             targetSdkVersion 29
             resValue "string", "build_config_package", "com.emurgo"
         }
+        nightly {
+            dimension "version"
+            applicationIdSuffix ".nightly"
+            applicationId 'com.emurgo'
```
SebastienGllmt, line 172: Needs a different application ID to be uploaded as a separate app in the play store

android/app/build.gradle
```diff
@@ -165,6 +166,15 @@ android {
...
+            minSdkVersion 21
+            compileSdkVersion = 29
+            targetSdkVersion 29
+            resValue "string", "build_config_package", "com.emurgo"
```
SebastienGllmt, line 176: not sure if `com.emurgo` needs to be changed here also

SebastienGllmt (changes requested):

android/app/build.gradle
v-almonacid: the resulting applicationId should be `com.emurgo.nightly`

android/app/build.gradle
v-almonacid: nope https://github.com/luggit/react-native-config#advanced-android-setup

SebastienGllmt (approved):

PR #787, opened 2020-02-25
improve chain libs memory management

v-almonacid:
In `js-chain-libs` some methods use a local copies of data while others operate on the input memory directly. As a consequence, it's difficul tot know when it's ok to free the pointers.

Fortunately, we can safely free everything in `react-native-chain-libs`, because freeing a null pointer doest not throw any error.

So I added a `.free()` to all objects once they are not used anymore.


v-almonacid:
Closing since Jormungandr support was removed.

PR #795, opened 2020-02-27
fix network info in settings

v-almonacid:
Currently we can only have builds that support both the Byron mainnet and the shelley testnet, so I'm hardcoding this information in the "network" label displayed in the settings.

This should be fixed in the future once the `.env` and `CONFIG.js` files are fixed.

Result:

Screenshot 2020-02-27 at 13 41 43

### DIFF
--- a/src/components/Settings/ApplicationSettingsScreen.js
+++ b/src/components/Settings/ApplicationSettingsScreen.js
@@ -235,7 +235,7 @@ const ApplicationSettingsScreen = ({
 
       <SettingsBuildItem
         label={intl.formatMessage(messages.network)}
-        value={CONFIG.NETWORK}
+        value={'Byron Mainnet / Shelley Testnet'}
       />
 
       <SettingsBuildItem


src/components/Settings/ApplicationSettingsScreen.js
```diff
@@ -235,7 +235,7 @@ const ApplicationSettingsScreen = ({
 
       <SettingsBuildItem
         label={intl.formatMessage(messages.network)}
-        value={CONFIG.NETWORK}
+        value={'Byron Mainnet / Shelley Testnet'}
```
SebastienGllmt, line 238: How temporary are these? If we think this will stick around for a while, maybe it's better to make this a translatable string? Or maybe we already have strings for "byron" and "shelley we can re-use to construct this string?

v-almonacid: We need some refactoring to support both Shelley testnet and mainnet. It's probably a good idea to do it before working on new features (like tx history, etc). So maybe next week?

I think it's safe to add that only to patch the coming release.

SebastienGllmt (approved):

PR #804, opened 2020-03-02
add detox dependency and  basic setup

v-almonacid:
After some research it seems that the `detox` package is the more widely used e2e testing framework for `react-native`. This PR does the following:

- adds the detox dependency
- setups dextox for both iOS and Android
- adds a test that setups a fresh wallet (as shown in gif below)
- adds some related documentation in the README.

Note: I wasn't able to get my test to work on iOS properly (in a reasonable time). It gets stuck after confirming the password. Tried many many hacks but nothing worked, so I give up.

ezgif-3-76679b525836


package.json
```diff
@@ -125,5 +134,33 @@
...
+  "detox": {
+    "test-runner": "jest",
+    "configurations": {
+      "ios.release": {
+        "binaryPath": "./ios/build/Build/Products/Release-iphonesimulator/emurgo.app",
+        "build": "xcodebuild -workspace ios/emurgo.xcworkspace -configuration release -scheme emurgo -sdk iphonesimulator -derivedDataPath ios/build",
+        "type": "ios.simulator",
+        "name": "iPhone 11"
+      },
+      "android.emu.mainnet.debug": {
```
SebastienGllmt, line 147: I don't think the debug Android build is used anywhere. Is there a benefit to keeping this as an option for the e2e test? For example maybe it's easier to debug an E2E test issue with this build?

v-almonacid: I don't think we should use it for CI but it can be practical in a situation in which somebody wants to run the test but doesn't have the signing keys needed for a release build. That's why I just added the setup config but didn't bother to add a `yarn` shortcut.

SebastienGllmt (approved):

PR #805, opened 2020-03-02
modify delegation delay explanation string

v-almonacid:
As requested, changed

> "Remember that your first reward will be allocated two epochs after the end of the epoch in which you have submitted your delegation transaction" 

by

>~~"Note that after you delegate to a stake pool, there will be a delay of two epochs before you start receiving rewards".~~

> Note that after you delegate to a stake pool, you will need to wait until the end of the epoch, plus two additional epochs, before starting receiving rewards.

### DIFF
--- a/src/i18n/locales/en-US.json
+++ b/src/i18n/locales/en-US.json
@@ -27,7 +27,7 @@
   "components.delegationsummary.notDelegatedInfo.secondLine": "Go to Staking center to choose which stake pool you want to delegate in. Note, you may delegate only to one stake pool in this Tesnnet.",
   "components.delegationsummary.upcomingReward.followingLabel": "Following reward",
   "components.delegationsummary.upcomingReward.nextLabel": "Next reward",
-  "components.delegationsummary.upcomingReward.rewardDisclaimerText": "Remember that your first reward will be allocated two epochs after the end of the epoch in which you have submitted your delegation transaction.",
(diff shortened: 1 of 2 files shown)


src/i18n/locales/en-US.json
```diff
@@ -27,7 +27,7 @@
   "components.delegationsummary.notDelegatedInfo.secondLine": "Go to Staking center to choose which stake pool you want to delegate in. Note, you may delegate only to one stake pool in this Tesnnet.",
   "components.delegationsummary.upcomingReward.followingLabel": "Following reward",
   "components.delegationsummary.upcomingReward.nextLabel": "Next reward",
-  "components.delegationsummary.upcomingReward.rewardDisclaimerText": "Remember that your first reward will be allocated two epochs after the end of the epoch in which you have submitted your delegation transaction.",
+  "components.delegationsummary.upcomingReward.rewardDisclaimerText": "Note that after you delegate to a stake pool, there will be a delay of two epochs before you start receiving rewards.",
```
SebastienGllmt: Are we okay not mentioning it's 2 epochs after the end of the epoch where your delegation preference was changed?

v-almonacid: I pointed this out to Leo too.

What about something like: "_Note that after you delegate to a stake pool, you will need to wait until the end of the epoch, plus two additional epochs, before starting receiving rewards._"?

SebastienGllmt: yes I think that wording is fine. In the Yoroi extension we use this wording for undelegating: _"It will take 2 epochs after the end of the current epoch for undelegation to take effect"_

SebastienGllmt (approved):

PR #806, opened 2020-03-06
warning if delegating to a pool in the blacklist

v-almonacid:
This PR integrates the reputation endpoint defined in EmIP6 (see Yoroi frontend's PR for reference https://github.com/Emurgo/yoroi-frontend/pull/1329)

In terms of design I just reused a typical "warning modal" that we have used for similar purposes.

Example for a pool with `node_flags = 1`:

Screenshot 2020-03-06 at 19 25 49

Clicking "I understand" sends the user to the "delegation confirmation" screen so their can still change their mind and choose not to send the delegation tx.


src/components/Delegation/DelegationCenter.js
```diff
@@ -91,40 +126,39 @@ export default injectIntl(
...
-      handleOnMessage: ({navigation, setBusy}) => async (event, intl) => {
-        setBusy(true)
+      navigateToDelegationConfirm: ({intl, navigation, setBusy}) => async (
+        selectedPools,
+      ) => {
         try {
-          const pools: Array<SelectedPool> = JSON.parse(
-            decodeURI(event.nativeEvent.data),
-          )
+          setBusy(true) // trigger "please wait" modal
```
SebastienGllmt: not specific to just here, but instead of doing setBusy with `finally`s everywhere which is error prone during refactoring, maybe it would have made more sense to make setBusy take an argument that's a lambda which guarantees setBusy is always set to false after the lambda is done running. Something like

```
wrapBusy(func: void => Promise<void>): Promise<void> {
    setBusy(true);
    try {
        await func();
    } finally {
        setBusy(false);
    }
}
```

src/components/Delegation/PoolWarningModal.js
```diff
@@ -0,0 +1,127 @@
...
+    <Text>
+      {'\u2022'} {textRow}
+    </Text>
+  )
+}
+
+type Props = {
+  intl: intlShape,
+  visible: boolean,
+  onPress: () => any,
```
SebastienGllmt: avoid using `any`

SebastienGllmt (approved):

src/components/Delegation/DelegationCenter.js
v-almonacid: @SebastienGllmt I added a wrapper as you suggested. I see the point on doing this but we must be careful when using this pattern with a `navigation.navigate()` call. If after navigating the origin component is unmounted, then `setBusy` will try to update the State of an unmounted component (which in this case will cause the please wait modal getting stuck forever). In this case it's safe doing it because the origin component is kept in the stack.

I'm leaving this PR unmerged for a while in case you want to review the last changes.

SebastienGllmt (approved):

PR #807, opened 2020-03-10
add nightly builds for iOS

v-almonacid:
In order to have a separate app with a different icon I needed to add:
- A new target (emurgo-nightly)
- A new scheme (emurgo-nightly)

The new scheme is set to use the new app id defined in the new target. So now nightly builds can also be installed in parallel on iOS:

Screenshot 2020-03-09 at 20 05 44


ios/Staging-Info.plist
```diff
@@ -21,7 +21,7 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>$(CURRENT_PROJECT_VERSION)</string>
+	<string>47</string>
```
SebastienGllmt, line 24: Why change this to a hard-coded number?

v-almonacid: I haven't touched that field, this was done by Xcode internally. Actually I only modified two files directly: `Fastfile` and fastlane's README. All other modifications are the result of creating and configuring a new *scheme* and a new *target* in Xcode.

v-almonacid: I was wrong, this was done automatically by fastlane when a new build is pushed. I'm setting up all this again in a new branch to get a cleaner result.

v-almonacid:
Deprecated in favor of https://github.com/Emurgo/yoroi-mobile/pull/808

PR #808, opened 2020-03-11
add nightly builds for ios (cleaner setup)

v-almonacid:
This PR replaces #807 and does exactly the same but with a cleaner setup (which also means a cleaner git diff).

One problem with Xcode is that many things are changed under the hood. Changing some parameter, and undoing that change, creates artifacts that pollute the project, even though nothing was really changed. Etc...

Also the last PR included changes done by fastlane when pushing to TestFlight. These are removed here.


ios/emurgo.xcodeproj/project.pbxproj
```diff
@@ -977,6 +1056,8 @@
 				2D16E6871FA4F8E400B85C8A /* Frameworks */,
 				F7FBD967219C5A6D003FE9E2 /* Recovered References */,
 				661CFB1B626FF527BA54DEF4 /* Pods */,
+				E9A1C61824183FB700151CFA /* emurgo copy2-Info.plist */,
+				E9A1C80B24185CCA00151CFA /* emurgo copy3-Info.plist */,
```
nicarq, line 1060: @v-almonacid I’m not sure if it’s unused or just stuff that didn’t get rename 🤔

v-almonacid: @nicarq I believe that the `emurgo copyX.plist` files were created when new targets were added. When I created the nightly target I noticed `emurgo copy3.plist` was created. But I don't remember if I created the `Nightly-Info.plist` manually or a new file was created automatically after I configured the target. You can check which plist file is used by each target in `target > Build Settings > Info.plist File`.
So it should be:
- emurgo -> emurgo/info.plist
- emurgo-staging -> Staging-info.plist
- emurgo-nightly -> Nightly-Info.plist

I assumed it was safe to remove the other ones, and so I did in the last commit. But I didn't realise there were still some references in the project. I don't think they are used though.

SebastienGllmt (approved):

PR #813, opened 2020-03-25
fix vulnerability in minimist

v-almonacid:
As I mentioned in the ticket description, fixing this requires: 

1) update react-native (on going)
2) bump eslint to version >7.0.0
3) add `"minimist": "1.2.3"` in resolutions entry in `package.json`, effectively forcing all dependencies to rely on 1.2.3.

This PR only does 3). 

Doing 2) breaks our configuration and also, 7.0.0 is still in alpha. IMO, we can skip it because forcing resolution on 1.2.3 doesn't seem to break anything in eslint.

I'll keep this as a draft until 1) is completed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,8 @@
     "react-test-renderer": "16.8.3"
   },
   "resolutions": {
-    "base-x": "3.0.4"
+    "base-x": "3.0.4",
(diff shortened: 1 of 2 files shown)


v-almonacid:
Closing since this doesn't seem to be an issue anymore

PR #816, opened 2020-04-02
storybook proof of concept for yoroi mobile

v-almonacid:
For instructions on how to setup and run, see README included in this PR.

## Summary

Add storybook in yoroi-mobile debug builds (technically, stories are embedded in all builds but are only available in debug builds).

## Some technical details
By default, stories have access to the initial state of redux's Store (ie. the global State seen by the app at start up). Also, any action that does not require a network call or an opened wallet should work by default (this means, that for instance, changing the language through the `LanguagePickerScreen` story will actually change the app's language---unless this action is mocked up.
Other actions, like those related to delegation or sending a tx, or generating new addresses, won't be available.

### Providing mock props to Redux's `connect()`
Stories don't have access to any Store prop that is updated from a network call; they only see the initial State. Thus, we may need to mockup some props for Screens and components that need network data. For simple components that don't access the Store this is straight forward (we just pass the prop directly like `<Component propX={something}>`.)

For Screen/components using redux's `connect` we need to do a small refactoring and add a third argument to `connect()`, which is the [`mergeProps`](https://react-redux.js.org/api/connect#mergeprops-stateprops-dispatchprops-ownprops-object) function. It can be implemented it as follows:

```
(state, dispatchProps, ownProps) => ({
  ...state,
  ...dispatchProps,
  ...ownProps,
}),
```

This way, the props passed to the component when instantiated inJSX will override those derived from the Store. 

You can see an example of this approach for mocking Store data in `src/components/Delegation/DelegationSummary.stories.js`  and `src/components/Delegation/DelegationSummary.js`.

Note that this changes the default behaviour of `mergeProps`, which is to return `{ ...ownProps, ...stateProps, ...dispatchProps }`.

## Preview
Screenshot 2020-04-02 at 19 36 06


src/components/Delegation/DelegationCenter.stories.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+
+import React from 'react'
+
+import {storiesOf} from '@storybook/react-native'
+
+import DelegationCenter from './DelegationCenter'
+
+storiesOf('DelegationCenter', module).add(
```
SebastienGllmt, line 9: FYI they suggest moving away from `storiesOf` and using their new "CSP" pattern instead when possible

SebastienGllmt (approved):

src/components/Delegation/DelegationCenter.stories.js
v-almonacid: Yep, but it's not available for react-native yet.

PR #817, opened 2020-04-08
Ledger Nano X integration

v-almonacid:
## Tx Signing with ledger using V1 rust bindings (hack needed)
As suggested by Sebastian, I used a hack based on [this PR](https://github.com/Emurgo/yoroi-frontend/pull/160) by Ruslan. In short, a tx is first signed using rust v1 bindings, then it is parsed in JS using `cbor`. The inputs and outputs from rust are provided to Ledger, and finally the witnesses of the original rust tx are replaced by the witnesses given by Ledger. The tx is then re-encoded in JS using `cbor`.
The logic for encoding/decoding rust txs is located in ` src/crypto/byron/util.js`

Note: the `encodeTxAsRust` function doesn't generate *exactly* the same bytes as rust (it differs in a few bytes of metadata I guess). However, the backend seems to decode it correctly.

## Some observed issues
Trying to connect to a device that has been idle for some time seems to fail, and a restart is needed. The best is to do all operations right after inserting the PIN and opening the ADA app (and actually, waiting for at least ~5 seconds after opening the app)

## New dependencies
- `@cardano-foundation/ledgerjs-hw-app-cardano` 1.0.8: Ledger operations
- `@ledgerhq/react-native-hw-transport-ble` 5.13.0: Ledger connection
- `react-native-ble-plx` 1.1.1: bluetooth connection. **An older version of this library was installed** because latest version (2.x) uses RN 0.60+. We'll need to upgrade this library once RN version is upgraded too.
- `cbor` 5.0.2: Needed for hack explained above
- `jsc-android`: Required by `@ledgerhq/react-native-hw-transport-ble`

## Features for follow up PR
- Address verification (UI needed)
- USB-cable connection support (this can only be supported for Android though)
- perhaps some minor UI improvements
- The "learn more about using yoroi with ledger" must point to some relevant link (video or tutorial), which currently doesn't exist now.

### Add wallet screen

image

### Send screen

Only added a few instructions, but might require some improvements.

image

### Check Nano X screen

image

#### Scrolling down shows more text in Android

image

### Connect Nano X screen

image

image

image

### Save Nano X screen 

image


ios/dummy_ble.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  dummy_ble.swift
+//  emurgo
+//
+// why do we need this file? see https://github.com/Polidea/react-native-ble-plx#ios-react-native--060-example-setup
+//
+//  NOTE: after upgrading to RN 0.60+ this file can be safely deleted
+//
+//  Created by Vicente Almonacid on 08/04/2020.
+//  Copyright © 2020 Facebook. All rights reserved.
```
SebastienGllmt: I think this copyright is incorrect

ios/emurgo-Bridging-Header.h
```diff
@@ -0,0 +1,4 @@
+//
+//  Use this file to import your target's public headers that you would like to expose to Swift.
```
SebastienGllmt, line 2: Do we still need these two header files?

package.json
```diff
@@ -26,6 +26,8 @@
     "@babel/core": "7.4.5",
     "@babel/node": "7.0.0",
     "@babel/runtime": "7.4.5",
+    "@cardano-foundation/ledgerjs-hw-app-cardano": "^1.0.8",
```
SebastienGllmt: Are we okay with non-strict dependencies for all of these?

src/__mocks__/config.js
```diff
@@ -10,29 +10,37 @@ import env from '../env'
...
     PROTOCOL_MAGIC: 764824073,
     API_ROOT: 'https://iohk-mainnet.yoroiwallet.com/api',
     EXPLORER_URL_FOR_TX: (tx: string) => `https://cardanoexplorer.com/tx/${tx}`,
   },
   SHELLEY: {
+    IS_SHELLEY: true,
+    NETWORK: 'Testnet', // for now, assume Shelley testnet by default
     PROTOCOL_MAGIC: 764824073,
     API_ROOT: 'https://shelley-itn-yoroi-backend.yoroiwallet.com/api',
+    SEIZA_STAKING_SIMPLE: (ADA: string) =>
```
SebastienGllmt, line 41: I'm confused why this file was modified in this PR. Additionally, why are these part of the mock but removed from the config.js in the same PR?

src/components/Send/ConfirmScreen.js
```diff
@@ -78,6 +81,32 @@ const handleOnConfirm = async (
     }
   }
 
+  // TODO: add a wrapper for setSendingTransaction()
```
SebastienGllmt: multiple TODOs

src/components/WalletInit/ConnectNanoX/CheckNanoXScreen.js
```diff
@@ -0,0 +1,104 @@
...
+    id: 'components.walletinit.connectnanox.checknanoxscreen.line3',
+    defaultMessage: '!!!Your smartphone is connected to the Internet.',
+  },
+  learnMore: {
+    id: 'components.walletinit.connectnanox.checknanoxscreen.learnMore',
+    defaultMessage: '!!!Learn more about using Yoroi with Ledger',
+  },
+})
+
+// TODO
```
SebastienGllmt, line 39: TODO

src/components/WalletInit/ConnectNanoX/ConnectNanoXScreen.js
```diff
@@ -0,0 +1,288 @@
...
+
+const deviceAddition = (device) => ({devices}) => ({
+  devices: devices.some((i) => i.id === device.id)
+    ? devices
+    : devices.concat(device),
+})
+
+type Props = {
+  intl: intlShape,
+  defaultDevices: ?Array<Device>, // for UI design prototyping
```
SebastienGllmt: ?

src/components/WalletInit/ConnectNanoX/ConnectNanoXScreen.js
```diff
@@ -0,0 +1,288 @@
...
+      await PermissionsAndroid.request(
+        PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION,
+      )
+    }
+    // check if bluetooth is available
+    let previousAvailable = false
+    TransportBLE.observeState({
+      next: (e) => {
+        if (this.bluetoothEnabled == null && !e.available) {
+          // consider adding a specific string since this is a common error
```
SebastienGllmt: ?

src/components/WalletInit/ConnectNanoX/ConnectNanoXScreen.js
```diff
@@ -0,0 +1,288 @@
...
+            this.bluetoothEnabled = e.available
+            this.reload()
+          }
+        }
+      },
+    })
+    this.startScan()
+  }
+
+  // componentDidUpdate = (prevProps) => {
```
SebastienGllmt: can we remove this?

src/components/WalletInit/ConnectNanoX/ConnectNanoXScreen.js
```diff
@@ -0,0 +1,288 @@
...
+          this.setState(deviceAddition(e.descriptor))
+        }
+      },
+      error: (error) => {
+        this.setState({error, refreshing: false})
+      },
+    })
+  }
+
+  // TODO
```
SebastienGllmt: TODO

src/components/WalletInit/ConnectNanoX/ConnectNanoXScreen.js
```diff
@@ -0,0 +1,288 @@
...
+    if (this.subscriptions != null) this.subscriptions.unsubscribe()
+    this.setState({
+      devices: [],
+      error: null,
+      refreshing: false,
+    })
+    this.startScan()
+  }
+
+  // TODO
```
SebastienGllmt: TODO

src/components/WalletInit/ConnectNanoX/DeviceItem.js
```diff
@@ -0,0 +1,71 @@
...
+  deviceName: {
+    fontSize: 20,
+    lineHeight: 24,
+    fontWeight: 'bold',
+    color: COLORS.LIGHT_POSITIVE_GREEN,
+  },
+})
+
+type Props = {
+  onSelect: (device: any) => any, // TODO
```
SebastienGllmt: add types and maybe make `Props` exact

src/crypto/byron/ledgerUtils.js
```diff
@@ -0,0 +1,331 @@
...
+
+//
+// ============== transaction logic ==================
+//
+
+/** Generate a payload for Ledger SignTx */
+export const createLedgerSignTxPayload = async (
+  unsignedTx: PreparedTransactionData,
+  txsBodiesMap: TxBodiesResponse,
+  // addressedUtxos: Array<AddressedUtxo>,
```
SebastienGllmt: Presumably you will need this to know the derivation path for the key of each input

src/crypto/byron/ledgerUtils.js
```diff
@@ -0,0 +1,331 @@
...
+}
+
+
+function _transformToLedgerInputs(
+  inputs: Array<TransactionInput>,
+  txDataHexMap: { [key: string]:string, ... }
+): Array<InputTypeUTxO> {
+  return inputs.map((input) => {
+    // TODO
+    // verifyAccountLevel(addressingInfo)
```
SebastienGllmt: I'm not sure you can check this in Yoroi Mobile since we don't have anything other than accounts

src/crypto/byron/ledgerUtils.js
```diff
@@ -0,0 +1,331 @@
...
+  change: string,
+): Array<OutputTypeAddress | OutputTypeChange> {
+  // note(v-almonacid): As opposed to yoroi frontend, which uses cardano_wallet
+  // bindings, in mobile the txOutputs array won't contain the change address
+  let outputs = []
+  outputs = txOutputs.map((txOutput) => ({
+    address58: txOutput.address,
+    amountStr: txOutput.value,
+  }))
+  // TODO: verify if order is important
```
SebastienGllmt: Just needs to match the same order what the Ledger signs. Protocol itself doesn't care about order of outputs

src/crypto/byron/ledgerUtils.test.js
```diff
@@ -0,0 +1,99 @@
...
+// import type {CryptoAccount} from './util'
+
+// jest.unmock('./util.js')
+// const myModule = require('myModule');
+// myModule.foo = jest.fn();
+
+jestSetup.setup()
+// jest.setTimeout(30 * 1000)
+
+describe('prepare HW tx', () => {
```
SebastienGllmt, line 10: I created some Jest tests in my latest merged PR in the Yoroi extension that you might be interested in

src/crypto/wallet.js
```diff
@@ -266,71 +305,124 @@ export class Wallet {
...
+            return Buffer.from(
+              await chimericAccountAddrPtr.as_bytes(),
+            ).toString('hex')
+          }
+          this._chimericAccountAddress = await retrieveChimericAddr(data)
+        }
+      }
+    } catch (e) {
+      Logger.error('wallet::_integrityCheck', e)
+      // TODO: open general error dialog. Use intl
```
SebastienGllmt: TODO

ios/dummy_ble.swift
v-almonacid: That was automatically generated by Xcode (I only added the note). What Copyright should be used?

ios/emurgo-Bridging-Header.h
v-almonacid: yep

src/__mocks__/config.js
v-almonacid: This is not really relevant. I just updated the mock config. I did a small change that broke some test, fixed it and changed a few extra things.

SebastienGllmt:
> Trying to connect to a device that has been idle for some time seems to fail, and a restart is needed. The best is to do all operations right after inserting the PIN and opening the ADA app (and actually, waiting for at least ~5 seconds after opening the app)

In the browser extension, I think we had to increase the timeout, but there is a browser limit so after the timeout we automatically fail the prompt. This is different from what you're doing (different transport mode and not a browser), but just so you're aware.

package.json
```diff
@@ -59,6 +63,7 @@
     "react-markdown": "^4.0.6",
     "react-native": "0.59.10",
     "react-native-background-timer": "^2.1.1",
+    "react-native-ble-plx": "^1.1.1",
```
SebastienGllmt: Are you okay with loose version number?

src/components/WalletInit/ConnectNanoX/ConnectNanoXScreen.js
```diff
@@ -0,0 +1,257 @@
...
+    : devices.concat(device),
+})
+
+type Props = {
+  intl: intlShape,
+  defaultDevices: ?Array<Device>, // for UI design prototyping
+  navigation: Navigation,
+}
+
+type State = {
```
SebastienGllmt: prefer exact type

src/components/WalletInit/ConnectNanoX/SaveNanoXScreen.js
```diff
@@ -0,0 +1,138 @@
...
+          onPress={onPress}
+          title={intl.formatMessage(messages.save)}
+          style={styles.button}
+        />
+      </View>
+    </SafeAreaView>
+  )
+}
+
+type ExternalProps = {
```
SebastienGllmt: prefer exact types

src/crypto/byron/ledgerUtils.js
```diff
@@ -0,0 +1,351 @@
...
+  OutputTypeAddress,
+  OutputTypeChange,
+  SignTransactionResponse,
+  Witness,
+} from '@cardano-foundation/ledgerjs-hw-app-cardano'
+import type {TxWitness} from './util'
+import type BluetoothTransport from '@ledgerhq/react-native-hw-transport-ble'
+
+// these are defined in LedgerConnectStore.js in yoroi-frontend
+type LedgerConnectionResponse = {
```
SebastienGllmt: prefer strict for all of these

src/crypto/byron/ledgerUtils.js
```diff
@@ -0,0 +1,351 @@
...
+}
+
+type LedgerSignTxPayload = {
+  inputs: Array<InputTypeUTxO>,
+  outputs: Array<OutputTypeAddress | OutputTypeChange>,
+}
+
+// Hardware wallet device Features object
+// borrowed from HWConnectStoreTypes.js in yoroi-frontend
+export type HWFeatures = {
```
SebastienGllmt: I removed a lot of these from the Yoroi Extension. Notably, I moved it down to 

```
export type HWFeatures = {|
  Vendor: string,
  Model: string,
  DeviceId: string,
|};
```
but I also use Label for the default name.
However, not that almost all of these are Trezor only (I think `vendor` is the only one provided by Ledger)

v-almonacid: Agree. I'll simplify. But why removing the app version? I think it's important if we need to debug any issue eventually (even though there should be very few version upgrades, I guess)

SebastienGllmt: The app version doesn't really make sense to store because it can change at any time. Somebody can start Yoroi, then while the app is still running, do an update of their Ledger device. If you ever need to do something based off the version, you would need to query it at the time

src/crypto/byron/ledgerUtils.js
```diff
@@ -0,0 +1,414 @@
...
+    e.message.includes('DisconnectedDevice') ||
+    e.name.includes('DisconnectedDevice') ||
+    e.message.includes('not found')
+  ) {
+    return true
+  }
+  return false
+}
+
+// note: e.statusCode === ErrorCodes.ERR_CLA_NOT_SUPPORTED probably means
```
SebastienGllmt: > probably

Have we confirmed this, or are we leaving it as "probably" for now?

src/components/WalletInit/ConnectNanoX/CheckNanoXScreen.js
```diff
@@ -0,0 +1,103 @@
...
+    defaultMessage: '!!!Before continuing, please make sure that:',
+  },
+  learnMore: {
+    id: 'components.walletinit.connectnanox.checknanoxscreen.learnMore',
+    defaultMessage: '!!!Learn more about using Yoroi with Ledger',
+  },
+})
+
+// TODO
+const url = '' // 'https://yoroi-wallet.com/...'
```
SebastienGllmt, line 40: Why is both this and `ledgerMessages.internetConnected` commented out?

src/i18n/global-messages.js
```diff
@@ -79,6 +79,60 @@ export const txLabels = defineMessages({
...
+      '!!!Cardano ADA app must remain open on your Ledger Nano X device.',
+  },
+  internetConnected: {
+    id: 'global.ledgerMessages.internetConnected',
+    defaultMessage: '!!!Your smartphone is connected to the Internet.',
+  },
+  // connection requisites
+  enableTransport: {
+    id: 'global.ledgerMessages.enableTransport',
+    defaultMessage: '!!!Enable bluetooth and location services (Android only).',
```
SebastienGllmt: Do we want to specify "Android only" in the string instead of only displaying this on Android devices?

SebastienGllmt:
Feedback:
> remove wallet no longer relevant message asking you to make sure you write down your 15 words.
> "change password" no longer relevant with Ledger confirmation for this kind of wallet
> easy transaction confirmation" no longer relevant with Ledger confirmation for this kind of wallet

src/components/WalletInit/ConnectNanoX/CheckNanoXScreen.js
v-almonacid: @SebastienGllmt `url` should be set once we have a video ready, we said we will add it once the video is ready.

About the `ledgerMessages.internetConnected` string, would you agree if we just remove it? I think I initially added it because we had it in the extension, but in the extension you really need internet to communicate with the http bridge. In mobile you don't strictly need internet.

src/crypto/byron/ledgerUtils.js
v-almonacid: Yes, "probably" is not precisely what I meant. Theoretically that error rises when the wrong app is opened in ledger, but I've observed it a few times when no app is opened at all and a tx is requested. So I said "probably" because the error is "probably" due to the user not having opened the ADA app instead of having opened the wrong app.

src/components/WalletInit/ConnectNanoX/ConnectNanoXScreen.js
v-almonacid: This is used by storybook.

SebastienGllmt (approved):

PR #818, opened 2020-04-15
fix copy button in pool info component

v-almonacid:
This one was pending for a while. Animation works well on both Android and iOS.

out

### DIFF
--- a/src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
+++ b/src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
@@ -12,6 +12,7 @@ import {
   Animated,
 } from 'react-native'
 import {injectIntl, defineMessages, intlShape} from 'react-intl'
+import _ from 'lodash'
 
 import {Text, TitledCard, Button} from '../../UiKit'
 import {formatStakepoolNameWithTicker} from '../../../utils/format'
@@ -39,6 +40,8 @@ const messages = defineMessages({
     defaultMessage: '!!!Copied!',
   },
 })
+
+const COPY_NOTIFICATION_TIME = 5000 // show 'copied' notification for 5 s
 
 const FadeOutView = (props) => {
   const [fadeAnim] = useState(new Animated.Value(1))
@@ -98,7 +101,9 @@ const DelegatedStakepoolInfo = ({
           </Text>
           <TouchableOpacity
             activeOpacity={0.5}
-            onPress={copyPoolHash}
+            onPress={_.debounce(copyPoolHash, COPY_NOTIFICATION_TIME, {
+              leading: true,
+            })}
             style={styles.spacedElem}
           >
             <Image source={copyIcon} width={24} />
@@ -132,7 +137,7 @@ export default injectIntl(
         showCopyNotif: false,
       },
       {
-        setshowCopyNotif: () => (showCopyNotif) => ({showCopyNotif}),
+        setShowCopyNotif: () => (showCopyNotif) => ({showCopyNotif}),
       },
     ),
     withHandlers({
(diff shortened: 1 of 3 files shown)


src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
```diff
@@ -80,12 +127,25 @@ const DelegatedStakepoolInfo = ({
...
     withHandlers({
       openExternalURL: ({poolURL}) => () => {
         if (poolURL) {
           Linking.openURL(poolURL)
         }
       },
+      copyPoolHash: ({poolHash, setshowCopyNotif}) => () => {
+        Clipboard.setString(poolHash)
+        setshowCopyNotif(true)
+        setTimeout(() => setshowCopyNotif(false), 10000)
```
SebastienGllmt: Why 10 seconds here?

v-almonacid: To avoid undesired rendering. Basically any additional tap within those 10 s won't generate an additional rendering of 'copied!' element. It'll be displayed 5 s and only after 5 more seconds a subsequent tab will render the message again.

SebastienGllmt: Is this the recommended way to do it over throttle/debounce?

nicarq: good catch Seba. this feels like a classic debounce case tho so maybe something like this? https://medium.com/@devmrin/debouncing-touch-events-in-react-native-prevent-navigating-twice-or-more-times-when-button-is-90687e4a8113

src/i18n/locales/en-US.json
```diff
@@ -14,6 +14,7 @@
   "components.common.languagepicker.russian": "Русский",
   "components.common.languagepicker.spanish": "Español",
   "components.delegation.delegationnavigationbuttons.stakingCenterButton": "Go to Staking Center",
+  "components.delegationsummary.delegatedStakepoolInfo.copied": "copied!",
```
nicarq: (the placeholder is correct) It should be: Copied!

src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
```diff
@@ -27,8 +34,35 @@ const messages = defineMessages({
       'components.delegationsummary.delegatedStakepoolInfo.fullDescriptionButtonLabel',
     defaultMessage: '!!!Go to website',
   },
+  copied: {
+    id: 'components.delegationsummary.delegatedStakepoolInfo.copied',
+    defaultMessage: '!!!Copied!',
+  },
 })
 
+const FadeOutView = (props) => {
```
nicarq, line 47: Cool to have animations :-)

src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
v-almonacid: Ah! I wasn't aware of those patterns (apart from intuition), nor that there are well-known implementations, thanks for the heads up! After reading a bit about it, I agree, it should be called with debounce and with the `leading` flag set. However, I still need that `setInterval` so that the component is actually re-rendered after a state change. Pushed changes.

src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
```diff
@@ -12,6 +12,7 @@ import {
   Animated,
 } from 'react-native'
 import {injectIntl, defineMessages, intlShape} from 'react-intl'
+import _ from 'lodash'
```
SebastienGllmt: prefer to import specific functions when possible to make better use of tree shaking

src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
```diff
@@ -141,10 +146,10 @@ export default injectIntl(
           Linking.openURL(poolURL)
         }
       },
-      copyPoolHash: ({poolHash, setshowCopyNotif}) => () => {
+      copyPoolHash: ({poolHash, setShowCopyNotif}) => () => {
         Clipboard.setString(poolHash)
-        setshowCopyNotif(true)
-        setTimeout(() => setshowCopyNotif(false), 10000)
+        setShowCopyNotif(true)
+        setTimeout(() => setShowCopyNotif(false), COPY_NOTIFICATION_TIME)
```
SebastienGllmt: Remove timeout on unmount

SebastienGllmt (approved):

PR #824, opened 2020-05-04
POC typed actions

SebastienGllmt:
I think we should be able to avoid ever having to do `Dispatch<any>` using this trick

### DIFF
--- a/src/actions/utxo.js
+++ b/src/actions/utxo.js
@@ -2,40 +2,70 @@
 
 import walletManager from '../crypto/wallet'
 
+import type {RawUtxo} from '../types/HistoryTransaction'
+
 import type {Dispatch} from 'redux'
 
 import type {State} from '../state'
 
+declare type ExtractReturnType = <R>((...arg: any) => R) => R;
+declare type ReturnType<Func> = $Call<ExtractReturnType, Func>;
+
 // start fetching utxo
-const _startFetching = () => ({
+const _startFetching: void => {|
+  type: 'START_FETCHING_BALANCE',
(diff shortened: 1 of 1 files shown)


src/actions/utxo.js
```diff
@@ -2,48 +2,84 @@
 
 import walletManager from '../crypto/wallet'
 
+import type {RawUtxo} from '../types/HistoryTransaction'
+
 import type {Dispatch} from 'redux'
 
 import type {State} from '../state'
 
+declare type ExtractReturnType = <R>((...arg: any) => R) => R;
```
SebastienGllmt, line 11: These two types shouldn't go here -- it's just convenient to place them here for this example.

They should go in a Flow utils folder and then added to the `[libs]` section of `.flowconfig` like we have in the Yoroi extension

src/actions/utxo.js
```diff
@@ -2,48 +2,84 @@
...
+|} = () => ({
   type: 'END_FETCHING_BALANCE',
   path: ['balance', 'isFetching'],
   payload: null,
   reducer: (state, isFetching) => false,
 })
 
-const _setUTXOs = (rawUTXOs) => ({
+const _setUTXOs: Array<RawUtxo> => {|
+  type: 'SET_UTXOS',
```
SebastienGllmt, line 40: note: we specify the type here as a constant string instead of just `string`. Flow can handle creating types out of constants and it needs this to be able to differentiate between the different types. Otherwise, if all types are just "string", Flow doesn't know what type to infer for the different possible dispatches.

PR #825, opened 2020-05-05
Lower password size to 10

SebastienGllmt:
This matches both Daedalus and the Yoroi extension which both only require 10


src/utils/validators.js
```diff
@@ -74,7 +74,7 @@ export const getPasswordStrength = (password: string): PasswordStrength => {
     return {isStrong: false}
   }
 
-  if (password.length >= 12) {
+  if (password.length >= 10) {
     return {isStrong: true, hasTwelveCharacters: true}
```
v-almonacid: Maybe rename `hasTwelveCharacters` by something more generic like `hasMinimumLen` ?

src/components/WalletInit/PasswordStrengthIndicator.js
```diff
@@ -21,34 +20,14 @@ const messages = defineMessages({
...
     defaultMessage: '!!!Continue',
     description: 'some desc',
   },
   passwordBigLength: {
     id: 'components.walletinit.passwordstrengthindicator.passwordBigLength',
-    defaultMessage: '!!!12 characters',
+    defaultMessage: '!!!10 characters',
     description: 'some desc',
   },
   or: {
```
v-almonacid: This one also seems unnecessary now (sorry 😬)

SebastienGllmt: How is it unnecessary? It's used to show the user the current password requirement is 10 characters.

v-almonacid: I mean the "or" string, I believe it's not used anymore.

SebastienGllmt: A good point. Removed some more dead code

v-almonacid (approved):
LGTM

PR #826, opened 2020-05-14
implement app disabled screen

v-almonacid:
When `isMaintenance==true`, the app navigates from the splash screen right to the newly added "maintenance screen". The only option available at this point is to close the app. 

Here is how this would look:

image


src/components/MaintenanceScreen.js
```diff
@@ -0,0 +1,84 @@
...
+  },
+  learnMore: {
+    id: 'components.maintenancemodal.learnMore',
+    defaultMessage: '!!!Learn more',
+  },
+})
+
+const URL = 'https://twitter.com/YoroiWallet'
+
+type Props = {
```
SebastienGllmt: prefer strict type

src/components/MaintenanceScreen.js
```diff
@@ -0,0 +1,84 @@
...
+      </View>
+    </Modal>
+  )
+}
+
+const MaintenanceScreen = ({intl}) => (
+  <MaintenanceModal
+    visible
+    onRequestClose={
+      /* eslint-disable-next-line */
```
SebastienGllmt: prefer to disable a single rule instead of all rules for a line

src/i18n/locales/en-US.json
```diff
@@ -42,6 +42,10 @@
   "components.firstrun.custompinscreen.title": "Set PIN",
   "components.login.appstartscreen.loginButton": "Login",
   "components.login.custompinlogin.title": "Enter PIN",
+  "components.maintenancemodal.attention": "ATTENTION:",
```
SebastienGllmt: Usually better to avoid repeating strings that are already present in the app since it reduces the amount of translation work required

SebastienGllmt (approved):

nicarq:
@SebastienGllmt @v-almonacid what do you guys think about adding a Try Again button (open to another text). Closing an App to retry doesn’t feel like the best UX -- some people may not even know how to do it

v-almonacid:
> @SebastienGllmt @v-almonacid what do you guys think about adding a Try Again button (open to another text). Closing an App to retry doesn’t feel like the best UX -- some people may not even know how to do it

mhm but the probability of an incident like this being resolved right before the user "tries again" is very low IMO. It is much more likely that the issue is resolved some time after the user closes the app.

nicarq:
> > @SebastienGllmt @v-almonacid what do you guys think about adding a Try Again button (open to another text). Closing an App to retry doesn’t feel like the best UX -- some people may not even know how to do it
> 
> mhm but the probability of an incident like this being resolved right before the user "tries again" is very low IMO. It is much more likely that the issue is resolved some time after the user closes the app.

if you leave the App and you come back, you didn’t close the App. For example, you enter the App and it was in maintain mode, you are like okay I will try later. The next day, you come back, you *still will* see the maintenance mode (!). Solution:

a) Whenever you come back to this screen, we re-check with the server is maintenance is over.
b) Add a retry button.

I really think we should do a) and b)

v-almonacid:
UPDATE:

As requested, I added a "try again" button. Once the error dialog appears for the first time, the `initApp` action —which checks the server status and decides what is the first screen a user will see— will be run every time the app comes to the foreground or when the user hits "try again".

With the button:

image

cc @nicarq @SebastienGllmt

src/i18n/locales/en-US.json
```diff
@@ -328,6 +326,7 @@
   "global.actions.dialogs.walletStateInvalid.title": "Invalid wallet state",
   "global.actions.dialogs.wrongPinError.message": "PIN is incorrect.",
   "global.actions.dialogs.wrongPinError.title": "Invalid PIN",
+  "global.attention": "Attention",
```
SebastienGllmt: FYI changing the key causes you to have to re-translate everything. For the extension I just keep the old key name if we ever have to rename something.

src/components/MaintenanceScreen.js
```diff
@@ -35,44 +38,92 @@ type Props = {|
...
+
+  componentDidMount() {
+    AppState.addEventListener('change', this._handleAppStateChange)
+  }
+
+  componentWillUnmount() {
+    AppState.removeEventListener('change', this._handleAppStateChange)
+  }
+
+  _handleAppStateChange = (nextAppState): void => {
```
SebastienGllmt: `_handleAppStateChange: State => void`

src/i18n/locales/en-US.json
v-almonacid: mhm I know, but I thought "it's just 1 word, and this way code will be cleaner". I'll just keep the old key ;)

PR #829, opened 2020-05-20
add address verification with ledger

v-almonacid:
## UI modifications/additions:
- the copy button icon, which opens the address modal, is replaced with an "info" icon, to be more generic: we can now copy or verify an address when tapping on an address in the receive screen.
- Ledger transport switch component: a simple dialog to pick between USB or Bluetooth for ledger connection. This component can be reused in other flows were connection with ledger is required.
- The address modal is extended for hardware wallets, to support address verification. But normal wallets won't see any difference.

address-verif


src/components/Ledger/LedgerTransportSwitchModal.js
```diff
@@ -0,0 +1,107 @@
...
+    id: 'components.ledger.ledgertransportswitchmodal.bluetoothButton',
+    defaultMessage: '!!!Connect with Bluetooth',
+  },
+})
+
+type Props = {|
+  intl: intlShape,
+  visible: boolean,
+  onSelectUSB: () => any,
+  onSelectBLE: () => any,
```
v-almonacid, line 48: @SebastienGllmt I know you always tell me to avoid `any` but these could be used in different contexts, so not sure if I can be more precise.

SebastienGllmt (approved):

PR #837, opened 2020-05-29
various minor en-US text revisions

Leo-Emurgo:
Some edits to en-US strings for improved readability and consistency


v-almonacid (approved):

src/i18n/locales/en-US.json
```diff
@@ -67,7 +67,7 @@
   "components.send.amountfield.label": "Amount",
   "components.send.biometricauthscreen.DECRYPTION_FAILED": "Fingerprint login failed. Please use an alternate login method.",
   "components.send.biometricauthscreen.NOT_RECOGNIZED": "Fingerprint was not recognized. Try again",
-  "components.send.biometricauthscreen.SENSOR_LOCKOUT": "You used too many fingers. Sensor is disabled",
```
SebastienGllmt, line 70: Are we sure this changed meaning is accurate?

src/i18n/locales/en-US.json
```diff
@@ -20,19 +20,19 @@
   "components.delegationsummary.delegatedStakepoolInfo.warning": "If you just delegated to a new stake pool, it may take a couple of minutes for the network to process your request.",
   "components.delegationsummary.epochProgress.endsIn": "Ends in",
   "components.delegationsummary.epochProgress.title": "Epoch progress",
-  "components.delegationsummary.failedwalletupgrademodal.title": "Heads up!",
-  "components.delegationsummary.failedwalletupgrademodal.explanation1": "Some users experienced problems while upgrading their wallets in the Shelley testnet.",
```
SebastienGllmt: I'm not sure it's worth changing these "failedwalletupgrademodal" strings since probably nobody sees them anymore so it's not worth forcing everybody to re-translate these

src/i18n/locales/en-US.json
```diff
@@ -84,9 +84,9 @@
   "components.send.confirmscreen.receiver": "Receiver",
   "components.send.confirmscreen.sendingModalTitle": "Submitting transaction",
   "components.send.confirmscreen.title": "Send",
-  "components.send.sendscreen.addressInputErrorInvalidAddress": "Please enter valid address",
+  "components.send.sendscreen.addressInputErrorInvalidAddress": "Please enter a valid address",
   "components.send.sendscreen.addressInputLabel": "Address",
-  "components.send.sendscreen.amountInput.error.INVALID_AMOUNT": "Please enter valid amount",
+  "components.send.sendscreen.amountInput.error.INVALID_AMOUNT": "Please enter a valid amount",
```
SebastienGllmt, line 89: there are a few other strings that don't have an "a". If you search for "Please enter" you can find and correct them all

src/i18n/locales/en-US.json
v-almonacid: Yes, he checked with me before changing it.

src/i18n/locales/en-US.json
v-almonacid: yeah probably not

Leo-Emurgo: Makes sense. I have no idea how to remove just these changes from the PR though, short of manually editing the file back to how it was. 

Incidentally, the meaning hasn't really changed so most languages would be able to use the existing translations without modification. Translators would just have to go back and re-confirm the translation on CrowdIn.

src/i18n/locales/en-US.json
Leo-Emurgo: You're right. I went and got the rest. 
I left instances of "Enter PIN" as they are since that doesn't need to be changed.

src/i18n/locales/en-US.json
v-almonacid: @Leo-Emurgo don't worry I'll take care of it.

SebastienGllmt:
@v-almonacid did you make the changes you wanted? Can we get this merged?

PR #839, opened 2020-06-07
review all floating promises

SebastienGllmt:
I reviewed all the floating promises in Yoroi Mobile and tried to decide if we should be awaiting on them, or explicitly mention that we do not want to await

**NOTE**: some of these I just took a guess. Please carefully review all of these since a wrongly added await can also break stuff


v-almonacid:
superseded by #1016

PR #843, opened 2020-06-25
bump version to 2.2.1

v-almonacid:


android/gradle.properties
```diff
@@ -16,6 +16,7 @@
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
+org.gradle.jvmargs=-Xmx2048M -Dkotlin.daemon.jvm.options\="-Xmx2048M"
```
SebastienGllmt: Why is this necessary?

v-almonacid: Good point. That line allows to increase the Java heap size, without it, it's impossible to build a release build. It's a known problem (they even put that line commented out by default as you can see above).

I'll just remove it because the problem may not arise in all environments.

SebastienGllmt (approved):

PR #845, opened 2020-07-08
Sorin/ch4072/implement wallet checksum ui

SorinC6:
- modify wallet selection screen(UI)
- added verify restored wallet screen(UI) (checksum)

Screenshot 2020-07-08 at 15 14 58
Screenshot 2020-07-08 at 15 15 13


v-almonacid:
Please check linting 🙏  Hint: you can run `yarn prettify`, then check if there are still some errors with `yarn lint`.

src/components/WalletSelection/WalletListItem.js
```diff
@@ -1,53 +1,62 @@
...
-import type {ComponentType} from 'react'
 
 type ExternalProps = {
   wallet: Wallet,
   onPress: (Wallet) => any,
 }
 
-const WalletListItem = ({wallet, onPress}) => (
+type CustomIconProps = {
+  isShelley: boolean,
```
SebastienGllmt: I would use the term "isJormungandr" instead of "isShelley"

src/components/WalletInit/RestoreWallet/VerifyRestoredWallet.js
```diff
@@ -0,0 +1,156 @@
...
+export default injectIntl(
+  (compose(
+    connect((state) => ({})),
+    withNavigation,
+    withNavigationTitle(({intl}) => intl.formatMessage(messages.title)),
+    withStateHandlers(
+      {
+        checksum: 'BONE-0770',
+        walletNumber: '123455', // added wallet number
+        // not sure about the phrase from restore wallet screen
```
SebastienGllmt: not sure what this comment means. However, it probably shouldn't be hard-coded here.

src/components/WalletInit/RestoreWallet/VerifyRestoredWallet.js
```diff
@@ -0,0 +1,156 @@
...
+})
+
+const walletRestaurationIntruction = (formatMessage) => [
+  formatMessage(messages.instructions1),
+  formatMessage(messages.instructions2),
+  formatMessage(messages.instructions3),
+]
+
+const walletAddresses = [
+  'DdfFfGgXxCcXcV … hHhJKKHgfdaa',
```
SebastienGllmt: Probably this shouldn't be hard-coded

src/components/WalletSelection/WalletSelectionScreen.js
```diff
@@ -104,10 +108,20 @@ const walletsListSelector = (state) => Object.values(state.wallets)
...
   (compose(
     connect(
-      (state: State) => ({
-        wallets: walletsListSelector(state),
-        currentVersion: currentVersionSelector(state),
-      }),
+      (state: State) => {
+        const newWalletsData = walletsListSelector(state).map((obj) => ({
+          ...obj,
+          avatar: walletAvatar, // hardcoded mock image
```
SebastienGllmt: remember to change this before merging

src/components/WalletInit/RestoreWallet/VerifyRestoredWallet.js
v-almonacid: it's ok, he only implemented the UI for now

src/components/WalletInit/RestoreWallet/VerifyRestoredWallet.js
SorinC6: Hi 😃 , I was not sure about this - the current screen (VerifyRestoreVallet) is placed between Restored wallet (the one in which user enter a 15-word recovery phrase) and the Wallet credentials (where user inputs wallet name and the passwoard ) . I'm not sure about how the integration with the backend will be and  if the wallet credential screen should have the recovery phrase or not (if I should pass it to it). Not sure exactly what data will need to generate the checksum and wallet image

SebastienGllmt: https://github.com/cardano-foundation/CIPs/blob/21d1a289efea4b8bef179ca75d0e61ef377bf22c/CIP4/CIP4.md this document describes exactly how it works and it has a reference implementation linked at the bottom. You can see how we use CIP4 in the Yoroi extension to see how we generate the image from the JS library.

package.json
```diff
@@ -37,8 +37,10 @@
     "base-x": "3.0.4",
     "bignumber.js": "^7.2.1",
     "bip39": "^2.5.0",
+    "blake2b": "^2.1.3",
```
v-almonacid: please use exact versions in all added dependencies

src/i18n/locales/en-US.json
```diff
@@ -293,6 +293,16 @@
   "components.walletinit.walletinitscreen.restoreWalletButton": "Restore wallet",
   "components.walletinit.walletinitscreen.restoreShelleyWalletButton": "Restore wallet (Shelley Testnet)",
   "components.walletinit.walletinitscreen.title": "Add wallet",
+
```
v-almonacid: there shouldn't be blank lines here (may be it won't generate an error in crowdin, but let's keep consistency)

v-almonacid:
I added all the missing logic in this PR. However, I still can't generate the same wallet icons (blockies) that the extension generates, even though both the extension and the mobile app use a library based on the same original library.

@SorinC6 if you have some free time, could you please try some other solution? I know there are a couple react-native compatible libraries that we can test.

v-almonacid:
After some pain I was able to generate the same blockies that the extension generates. For that, I [forked](https://github.com/Emurgo/react-native-blockies-svg) and fixed one implementation I've found.

Here's an example:

## Shelley:
image

## Byron
image

## Wallet selection
image

## What the extension shows
Screenshot 2020-08-21 at 21 03 05

Does this look good for you guys? @SebastienGllmt @nicarq

SebastienGllmt:
Awesome!

v-almonacid (approved):

PR #846, opened 2020-07-10
implement v2 endpoint

v-almonacid:


src/crypto/transactionCache.js
```diff
@@ -252,7 +332,15 @@ export class TransactionCache {
 
   static fromJSON(data: TransactionCacheState) {
     const cache = new TransactionCache()
-    cache.updateState(data)
+    // if cache is deprecated it means it was obtained from old history endpoint.
+    // in this case, we do not load the data and start from a fresh object.
+    const isDeprecatedCache = ObjectValues(data.perAddressSyncMetadata).some(
+      // $FlowFixMe
```
SebastienGllmt: // $FlowFixMe ?
Maybe a comment why this is necessary? I'm assuming it's because this field doesn't exist anymore after the migration to v2?

src/api/facade.js
```diff
@@ -57,64 +47,52 @@ export const checkValidHash = (data: string) => {
...
-    tx.tx_state !== 'Successful' || checkNonNegativeInt(tx.block_num),
-    'Invalid block_num',
-    tx.block_num,
-  )
-
-  const blockNum = tx.block_num ? parseInt(tx.block_num, 10) : null
-
-  tx.inputs_amount.forEach((amount) => {
-    assert.assert(checkNonNegativeInt(amount), 'Invalid input amount', amount)
+  // const blockNum = tx.block_num != null ? parseInt(tx.block_num, 10) : null
```
SebastienGllmt: Why are these commented out?

src/api/facade.js
```diff
@@ -57,64 +47,52 @@ export const checkValidHash = (data: string) => {
...
         'Invalid output address',
-        address,
+        output.address,
       )
     }),
   )
 
-  assert.assert(checkISO8601Date(tx.time), 'Invalid time', tx.time)
+  assert.assert(
+    tx.tx_state !== 'Successful' || checkISO8601Date(tx.time),
```
SebastienGllmt, line 101: Maybe add a comment saying this field is only present if the tx is successful which is why we need this `||`?

src/api/facade.js
```diff
@@ -57,64 +47,52 @@ export const checkValidHash = (data: string) => {
 
 export const checkAndFacadeTransactionAsync = async (
   tx: RawTransaction,
+  bestBlockNum: number,
```
SebastienGllmt: Do we want to  continue including `bestBlockNum` as part of the `Transaction` type?

src/crypto/transactionCache.js
```diff
@@ -168,7 +234,7 @@ export class TransactionCache {
 
   // Returns number of updated transactions
   _checkUpdatedTransactions(transactions: Array<Transaction>): number {
-    // Logger.debug('_updateTransactions', transactions)
+    Logger.debug('_updateTransactions', transactions)
```
SebastienGllmt: presumably this should be `_checkUpdatedTransactions` instead of `_updateTransactions` in the log statement

src/crypto/transactionCache.js
```diff
@@ -132,24 +171,51 @@ export class TransactionCache {
...
+  _buildTxHistoryRequest(
+    addresses: Array<string>,
+    metadata: SyncMetadata,
+    currentBestBlockHash: ?string,
+  ): TxHistoryRequest {
+    assert.assert(
+      currentBestBlockHash != null,
+      'buildTxHistoryRequest: bestBlock not null',
+    )
+    /* :: if (currentBestBlockHash == null) throw 'assert' */
```
SebastienGllmt, line 205: remove  comment?

src/api/facade.js
v-almonacid: before `tx.block_num` was string so it needed to be parsed. Both are ints. Just forgot to remove those comments

src/crypto/transactionCache.js
v-almonacid: This one is flow comment. We use it in situations in which, even though there is an assert, flow fails to check the type of the test variable.

src/api/facade.js
v-almonacid: hmm maybe we can get rid of it. I left it for backwards compatibility but I guess I can workaround the fact of not having it directly in the tx

SebastienGllmt (approved):

PR #848, opened 2020-07-13
Sorin/warning banner in txhistory

SorinC6:
Add  `WarningBanner`  used to display a warning message in TxHistory screen (UI)

test


v-almonacid:
I removed the upgrade button since the upgrade feature will probably be implemented in Shelley wallets.

Also, the banner should only be displayed in Byron wallets. Since Shelley wallets were introduced after this PR, I'll take care of this issue in a follow up PR because rebasing and fixing conflicts will take longer.

v-almonacid (approved):

PR #849, opened 2020-07-15
hack to fix tx sending with ledger after api upgrade

v-almonacid:


SebastienGllmt (approved):
no real way of verifying this works, but worst case if there is a case with a subtle bug, the transaction will just be rejected.
Since this will have to be entirely re-written to Shelley anyway, I suggest we just push this out to unblock the average user.

PR #859, opened 2020-07-29
add support for multiple networks

v-almonacid:
Previously we supported only two networks (Byron and jormun). The idea now is to have a more flexible store and api so that we can support several networks. For that, we need to decouple several things, namely, network configuration parameters, API library, crypto libraries, etc.

This PR is a first step in this direction, but more work will be required.

In the following PR I pretend to decouple `Wallet` class from `WalletManager` (currently they both live in `wallet.js`), and create separate `Wallet` classes for each supported network. Plus, include some Haskell-Shelley logic.


src/config/networks.js
```diff
@@ -0,0 +1,78 @@
...
+      API_ROOT: 'https://iohk-mainnet.yoroiwallet.com/api',
+      ..._DEFAULT_BACKEND_RULES,
+    },
+    COIN_TYPE: NUMBERS.COIN_TYPES.CARDANO,
+  },
+  HASKELL_SHELLEY: {
+    NETWORK_ID: NETWORK_REGISTRY.HASKELL_SHELLEY,
+    IS_MAINNET: true,
+    BACKEND: {
+      API_ROOT: 'TODO',
```
SebastienGllmt, line 30: URL stays the same as in Byron since the Byron endpoint will no longer exist.

SebastienGllmt (approved):

PR #873, opened 2020-08-25
add delegation tx functions

v-almonacid:
This PR just ports some delegation tx functions/utilities from the extension.

I decided to include the `HaskellShelleyTxSignRequest` class, though I can't use it yet cause some bindings are missing.


src/crypto/shelley/delegationUtils.js
```diff
@@ -0,0 +1,298 @@
...
+      }
+    }
+  }
+
+  return sumOutForKey.minus(sumInForKey)
+}
+
+export type CreateDelegationTxRequest = {|
+  absSlotNumber: BigNumber,
+  // computeRegistrationStatus: (void) => Promise<boolean>,
```
v-almonacid, line 176: One thing I still need to figure out is how to get and store certificate registration history. I think that I'll omit that part for now and just use the `accountState` endpoint to determine whether a given key has a registered certificate or not.

SebastienGllmt: >accountState

the account state endpoint doesn't tell you this FYI. The registration history endpoint (not deployed yet) and the tx history endpoint can tell you this however

src/crypto/shelley/transactionUtils.js
```diff
@@ -74,6 +74,7 @@ export const createUnsignedTx = async (
         absSlotNumber,
         protocolParams,
         [],
+        // [], // TODO
```
SebastienGllmt, line 77: TODO

SebastienGllmt (approved):

PR #970, opened 2020-09-08
update haskell shelley tx functions and tests

v-almonacid:
This PR:
- Updates the tx functions and tests to match those currently used in the Yoroi extension
- bump Haskell Shelley libs
- Fix/remove some old tests (now all tests should pass again)

~TODO before merging: implement `Vkeywitness::from_bytes()`~ DONE


src/api/fetch.js
```diff
@@ -9,22 +9,20 @@ import type {NetworkConfig} from '../config/types'
 
 type RequestMethod = 'POST' | 'GET'
 
-const _checkResponse = async (response, requestPayload) => {
-  if (response.status === 404) {
-    const responseBody = await response.json()
+const _checkResponse = (status, responseBody, requestPayload) => {
+  // TODO: verify that this is the correct behaviour
```
SebastienGllmt, line 13: looks fine to me. In Yoroi these error specifically trigger a rollback, but in mobile it can just be a regular error like any other.

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -507,11 +509,27 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
...
 
+    const wits = new Set()
+
+    const txBuilder: TransactionBuilder = signRequest.unsignedTx
+    // prettier-ignore
+    wits.add(
+      Buffer.from(
+        await (await make_vkey_witness(
+          await hash_transaction(
+            // $FlowFixMe
```
SebastienGllmt: Why do you need this FlowFixMe?

src/crypto/shelley/transactionUtils.js
```diff
@@ -74,7 +74,8 @@ export const createUnsignedTx = async (
         absSlotNumber,
         protocolParams,
         [],
-        // [], // TODO
+        [],
+        true,
```
SebastienGllmt: did you decide you want to allow 0-output transactions?

src/api/fetch.js
v-almonacid: What we do in mobile is we reset the tx cache and trigger a resynch https://github.com/Emurgo/yoroi-mobile/blob/1ad978e91fd74e56ea41e81cc8b9339f35a13e9b/src/crypto/shelley/transactionCache.js#L360

src/crypto/shelley/ShelleyWallet.js
v-almonacid: Flow can't get the types of TransactionBuilder right. It complains saying `build()` not found in `TransactionBuilder`.  I suspect that when a type is used as input in a polymorfic type, in this case `BaseSignRequest<TransactionBuilder>`, for some reason Flow fails to check the types.

src/crypto/shelley/transactionUtils.js
v-almonacid: Ah good point, no not definitely. But I think I'll just keep it consistent with the extension's behaviour.

SebastienGllmt: Okay. Just FYI the extension disabled 0-output transactions in all cases now.

src/crypto/shelley/ShelleyWallet.js
SebastienGllmt: Yeah, you have to use instanceof to cast it. I'd like to get rid of `BaseSignRequest` eventually since as you've probably notice, it doesn't work that well. The TxSignRequest makes a lot more logical sense (although it's questionable that it implements an interface and I may have to remove that some day)

SebastienGllmt (approved):

PR #972, opened 2020-09-11
add support for ledger in shelley

v-almonacid:
TODO:
- [x] Shelley and Byron wallet connection 
- [x] Tx sending
- [x] Delegation txs
- [x] Address verification

Note: Withdrawals will be implemented in a separate PR.


android/app/src/dev/res/xml/network_security_config.xml
```diff
@@ -4,5 +4,6 @@
         <domain includeSubdomains="false">localhost</domain>
         <domain includeSubdomains="false">10.0.2.2</domain>
         <domain includeSubdomains="false">10.0.3.2</domain>
+        <domain includeSubdomains="true">192.168.1.74</domain>
```
SebastienGllmt: ?

package.json
```diff
@@ -77,7 +76,7 @@
     "react-native-device-info": "6.0.1",
     "react-native-easy-markdown": "1.4.1",
     "react-native-fs": "^2.11.18",
-    "react-native-haskell-shelley": "Emurgo/react-native-haskell-shelley#ff2b8cb9fa8655c65464403a9afd5e8a52134d20",
+    "react-native-haskell-shelley": "Emurgo/react-native-haskell-shelley#0.0.6",
```
SebastienGllmt: maybe it's time we push this to NPM?

src/crypto/shelley/ledgerUtils.js
```diff
@@ -0,0 +1,905 @@
...
+  deviceId: ?DeviceId,
+  deviceObj: ?DeviceObj,
+  useUSB?: boolean = false,
+): Promise<HWDeviceInfo> => {
+  try {
+    Logger.debug('ledgerUtils::getHWDeviceInfo called')
+
+    const transport = await connectionHandler(deviceId, deviceObj, useUSB)
+    const appAda = new AppAda(transport)
+    const versionResp: GetVersionResponse = await appAda.getVersion()
```
SebastienGllmt: Keep in mind we need version >= 2.0.4 of the Cardano app

src/crypto/shelley/ledgerUtils.js
```diff
@@ -0,0 +1,905 @@
...
+          await baseAddr.stake_cred(),
+        )
+        const addressPayload = Buffer.from(
+          await (await rewardAddr.to_address()).to_bytes(),
+        ).toString('hex')
+        stakingKeyAddressing[addressPayload] = {
+          path: [
+            CONFIG.NUMBERS.WALLET_TYPE_PURPOSE.CIP1852,
+            CONFIG.NUMBERS.COIN_TYPES.CARDANO,
+            CONFIG.NUMBERS.ACCOUNT_INDEX + CONFIG.NUMBERS.HARD_DERIVATION_START,
```
SebastienGllmt, line 435: Just FYI we'll have to support multi-account for Ledger this year most likely, so be careful of hardcoding stuff like this

SebastienGllmt:
Is the DeviceId you get from `@ledgerhq/react-native-hw-transport-ble` the same as what you get from the call to `getSerial` in the  Shelley Cardano app?
The idea is we want to save the result of `getSerial` going forward so that when we have multi-account support, we can tell the user which accounts all belong to the same device.

android/app/src/dev/res/xml/network_security_config.xml
v-almonacid: oh I needed to whitelist that IP address to debug over wifi. That's the debug network config so it doesn't affect production builds, but I remove it anyway.

package.json
v-almonacid: mhm I'd like to but it's still a bit incomplete. Among other things, I still need to write bindings for Enterprise and pointer addresses, tx metadata, etc. I don't have rights to push to NPM so I prefer not to bother you anytime I need to push an update.

v-almonacid:
> Is the DeviceId you get from `@ledgerhq/react-native-hw-transport-ble` the same as what you get from the call to `getSerial` in the Shelley Cardano app?
> The idea is we want to save the result of `getSerial` going forward so that when we have multi-account support, we can tell the user which accounts all belong to the same device.

They are different. Also, we have a `DeviceId` for handling bluetooth connections and `DeviceObj` for USB. The 3 of them are different.

SebastienGllmt (approved):

PR #980, opened 2020-09-22
implement "send all" feature

v-almonacid:
## Important things to take into account (to discuss)

- I set a `step=10` for the slider (0, 10%..., 100%). 
- once the user moves the slider, the amount text input is automatically disabled. The reason for this is that if user moves the slider, and then sets the amount manually in the text input, we should then set the slider to reflect this, and this value might probably be not a multiple of `step`, and that could be confusing.
- If the slider is set back to 0%, I re-enable the text input. Not sure if this is the best approach.
- When the slider is set to 100%, user triggers a `sendAll` transaction. This will update the amount in the screen according to the fee.

## Demo
slider-demo

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -38,6 +38,9 @@ node_modules/warning/.*
 ; Ignore @emurgo/cardano-serialization-lib-nodejs
 .*/node_modules/@emurgo/cardano-serialization-lib-nodejs/cardano_serialization_lib.js.flow
 
+; Ignore @react-native-community/checkbox
+.*/node_modules/@react-native-community/checkbox/.*/.*
+
 [include]
 
 [libs]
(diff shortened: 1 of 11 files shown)


v-almonacid:
Removed the slider and used a checkbox instead, as requested. Ready for review @SebastienGllmt

.flowconfig
```diff
@@ -38,6 +38,9 @@ node_modules/warning/.*
 ; Ignore @emurgo/cardano-serialization-lib-nodejs
 .*/node_modules/@emurgo/cardano-serialization-lib-nodejs/cardano_serialization_lib.js.flow
 
+; Ignore @react-native-community/checkbox
+.*/node_modules/@react-native-community/checkbox/.*/.*
```
SebastienGllmt: Do we still need this after switching to our own checkbox?

SebastienGllmt (approved):

PR #1013, opened 2020-10-02
fix pool name displayed when there is no pool name data from the server

v-almonacid:
Story details: https://app.clubhouse.io/emurgo/story/6212

### DIFF
--- a/src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
+++ b/src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
@@ -41,7 +41,7 @@ const messages = defineMessages({
   },
   unknownPool: {
     id: 'components.delegationsummary.delegatedStakepoolInfo.unknownPool',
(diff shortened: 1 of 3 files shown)


src/components/Delegation/dashboard/DelegatedStakepoolInfo.js
```diff
@@ -40,8 +39,24 @@ const messages = defineMessages({
     id: 'components.delegationsummary.delegatedStakepoolInfo.copied',
     defaultMessage: '!!!Copied!',
   },
+  unknownPool: {
+    id: 'components.delegationsummary.delegatedStakepoolInfo.unknownPool',
+    defaultMessage: '!!!Unknown pool!',
```
SebastienGllmt: There's an `!` here whereas there isn't one in `en-US.json`. Probably we don't want the exclamation mark

SebastienGllmt (approved):

PR #1014, opened 2020-10-07
add support for 24 word wallets

v-almonacid:
This PR adds support for 24-word Shelley wallets. It also adds a reusable `ExpandableItem` (aka accordion), here used for the "learn more" label.

restore


src/components/WalletInit/WalletInitScreen.js
```diff
@@ -36,12 +38,40 @@ const messages = defineMessages({
...
+      'words generated when you created a Yoroi Wallet, choose this option ' +
+      'to restore your wallet.',
+  },
+  restore24WordWalletLabel: {
+    id: 'components.walletinit.walletinitscreen.restore24WordWalletLabel',
+    defaultMessage: '!!!24-word Wallet',
+  },
+  restore24WordWalletExplanation: {
+    id: 'components.walletinit.walletinitscreen.restore24WordWalletExplanation',
+    defaultMessage:
```
SebastienGllmt, line 58: It's probably better to make the use something like
```
If you have a recovery phrase consisting of {wordLength} words, choose this option to restore your wallet.
```
Since then you can re-use the same translation string for all languages (even if we add  support for 12-word and 18-word in the future)

src/config/types.js
```diff
@@ -19,7 +19,7 @@ export type DerivationType = $Values<typeof DERIVATION_TYPES>
 export const WALLET_IMPLEMENTATION_REGISTRY = {
   HASKELL_BYRON: 'haskell-byron', // bip44
   HASKELL_SHELLEY: 'haskell-shelley', // cip1852
-  BYRON: 'byron', // deprecated
+  DAEDALUS_HASKELL_SHELLEY: 'daedalus-haskell-shelley', // cip1852
```
SebastienGllmt: I'm not sure if calling the "Daedalus Shelley" makes sense since any wallet can add support for 24.

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -267,6 +245,7 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
     this.version = data.version
     this.internalChain = AddressChain.fromJSON(data.internalChain)
     this.externalChain = AddressChain.fromJSON(data.externalChain)
+    this.rewardAddressHex = await this.internalChain.getRewardAddressHex()
```
SebastienGllmt: I'm confused. How is `getRewardAddressHex` is related to `internalChain`

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -595,10 +572,8 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
       withdrawals: [
         {
           addressing: this._getRewardAddressAddressing(),
-          rewardAddress: Buffer.from(
-            await (await this.getRewardAddress()).to_bytes(),
-            'hex',
-          ).toString('hex'),
+          // $FlowFixMe flow thinks rewardAddressHex can be null
```
SebastienGllmt: You can usually fix this by doing something like
```javascript
const { rewardAddressHex} = this;
if (rewardAddressHex== null) throw new Error('reward address is null')
```

the fact it's non-null should propagate correctly in that case.

src/crypto/transactionUtils.js
```diff
@@ -56,6 +59,32 @@ export const processTxHistoryData = (
...
   const isMultiParty =
     ownInputs.length > 0 && ownInputs.length !== tx.inputs.length
+
+  // check if tx includes a deregistration (which means output may contain
+  // a deposit to this wallet)
+  const hasKeyDeregistration = (() => {
+    let _hasKeyDeregistration = false
+    for (const cert of tx.certificates) {
+      if (
+        cert.kind === CERTIFICATE_KIND.STAKE_DEREGISTRATION &&
```
SebastienGllmt: Pool retirement also has a refund FYI.

The way you're implementing this function also differs from the Yoroi Extension. I suggest you look at the call to `getFromUserPerspective` in `updateTransactions.js` in the Yoroi Extension along with related comments.

src/config/types.js
v-almonacid: The main reason why I wanted a new wallet ID was basically to have a bit of information on whether a wallet might have been restored from a Daedalus seed (mostly for debugging purposes). But since you mention we may add support for 12 or 18-word or whatever, I reckon it doesn't make much sense to store this information, specially since from the app perspective all wallets correspond to the same implementation. I'll remove that and just keep `HASKELL_SHELLEY`

UPDATE: I ended up just renaming as `HASKELL_SHELLEY_24`.

src/crypto/shelley/ShelleyWallet.js
v-almonacid: both the internal and external chains cache the public key and can derive the reward address. But probably it'll be less confusing to derive it right there explicitly.

src/crypto/transactionUtils.js
v-almonacid: Thanks, I'll take a look 👍 

Also note that we have a difference in the way amounts are computed. The extension adds the fee to the amount while mobile doesn't.

v-almonacid: @SebastienGllmt I refactored the code and tried to match it to the Yoroi extension (last 2 commits).

Also all other comments have been addressed.

src/components/WalletInit/WalletInitScreen.js
```diff
@@ -36,12 +38,40 @@ const messages = defineMessages({
...
     defaultMessage: '!!!Restore wallet',
   },
+  restoreNormalWalletLabel: {
+    id: 'components.walletinit.walletinitscreen.restoreNormalWalletLabel',
+    defaultMessage: '!!!15-word Wallet',
+  },
+  restoreNormalWalletExplanation: {
+    id: 'components.walletinit.walletinitscreen.restoreNormalWalletExplanation',
+    defaultMessage:
+      '!!!If you have a Yoroi recovery phrase consisting of 15 ' +
```
SebastienGllmt, line 48: I noticed you changed `components.walletinit.walletinitscreen.restoreNWordWalletExplanation` to use a variable, but the other instances in this file don't use a variable.

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -153,32 +143,26 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
...
       Buffer.from(masterKey, 'hex'),
     )
-    const accountKey = await (await (await masterKeyPtr.derive(
-      _purpose,
-    )).derive(CONFIG.NUMBERS.COIN_TYPES.CARDANO)).derive(
-      0 + CONFIG.NUMBERS.HARD_DERIVATION_START,
-    )
+    const accountKey = await (await (await masterKeyPtr.derive(purpose)).derive(
+      CONFIG.NUMBERS.COIN_TYPES.CARDANO,
+    )).derive(0 + CONFIG.NUMBERS.HARD_DERIVATION_START)
```
SebastienGllmt: be careful about hard-coding the 0th account, since we'll have to support multiple accounts per wallet some time this year.

SebastienGllmt (approved):

src/components/WalletInit/WalletInitScreen.js
v-almonacid: I thought I'd keep that one for now because 1) the message is slightly different, it refers explicitly to wallets created using yoroi (which is not a requirement but may be a bit more reassuring for users) and 2) changing it would break translations. I can change it in the future to use the same generic message, once we have more translations.

PR #1015, opened 2020-10-08
attempt to prevent null deviceId error

v-almonacid:
It is not clear why a few users get `deviceId=null` error since:

- right after opening a wallet there is an integrity check that verifies that `HWDeviceInfo != null` (it doesn't check for `deviceId` though, which is a field of `HWDeviceInfo`.

- If for some reason `deviceId=null` (it's the case for a user who connected via USB and then decided to send a tx using BLE), then a connection dialog is displayed to retrieve the `deviceId`.

This PR tries to prevent this issue, and if detected, it suggests the user to create a new wallet (see pic below. Note: changed "device data" by "device metadata" to cause less panic).

image

### DIFF
--- a/src/components/WalletInit/ConnectNanoX/SaveNanoXScreen.js
+++ b/src/components/WalletInit/ConnectNanoX/SaveNanoXScreen.js
@@ -136,8 +136,8 @@ export default injectIntl(
           walletImplementationId,
           hwDeviceInfo,
         )
-        navigation.navigate(ROOT_ROUTES.WALLET)
         saveHW(hwDeviceInfo)
+        navigation.navigate(ROOT_ROUTES.WALLET)
       },
     }),
     withHandlers({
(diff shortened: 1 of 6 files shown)


src/components/WalletInit/ConnectNanoX/SaveNanoXScreen.js
```diff
@@ -136,8 +136,8 @@ export default injectIntl(
           walletImplementationId,
           hwDeviceInfo,
         )
-        navigation.navigate(ROOT_ROUTES.WALLET)
         saveHW(hwDeviceInfo)
+        navigation.navigate(ROOT_ROUTES.WALLET)
```
v-almonacid, line 140: it makes more sense to save the device info before navigating but technically I think this doesn't make any difference, so the issue doesn't come from here.

SebastienGllmt (approved):

PR #1022, opened 2020-10-17
bump version to 3.3.0

v-almonacid:


android/gradle.properties
```diff
@@ -11,6 +11,7 @@
 # The setting is particularly useful for tweaking memory settings.
 # Default value: -Xmx10248m -XX:MaxPermSize=256m
 # org.gradle.jvmargs=-Xmx2048m -XX:MaxPermSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
+org.gradle.jvmargs=-Xmx2048M -Dkotlin.daemon.jvm.options\="-Xmx2048M
```
v-almonacid, line 14: This is intended (I think it's already the default config recommended by Android studio)

PR #1024, opened 2020-10-20
fix displayed balance in tx history

v-almonacid:
Currently the wallet balance shown in the tx history screen is computed from the local tx cache. However, this doesn't work when there are intra-wallet txs with implicit outputs.

This PR fixes this by computing the balance directly from the UTXO set given by the server. Not an ideal solution IMHO but it fixes the issue.


SebastienGllmt:
Why can't you just sum up the MIR certificates and the withdrawals in the transaction instead of doing this?

SebastienGllmt:
>However, this doesn't work when there are intra-wallet txs with implicit outputs

Sorry, I don't understand why this would be a problem

SebastienGllmt (approved):

PR #1059, opened 2020-11-13
add support for read-only wallets

v-almonacid:
read-only


src/components/WalletInit/RestoreWallet/ImportReadOnlyWalletScreen.js
```diff
@@ -0,0 +1,150 @@
...
+    id: 'components.walletinit.importreadonlywalletscreen.line2',
+    defaultMessage:
+      '!!!Click on the QR code icon of the wallet you want to ' +
+      'import in the mobile app.',
+  },
+})
+
+let scannerRef // reference to QR code sanner to re-activate if required
+
+// TODO(v-almonacid): move these validators somewhere else
```
SebastienGllmt: Did you want to handle this TODO in a different PR?

SebastienGllmt (commented):
Can we get a E2E test for this also? Maybe hard with the QR code involved.

src/components/WalletInit/RestoreWallet/ImportReadOnlyWalletScreen.js
```diff
@@ -0,0 +1,148 @@
...
+      'extension, you will need to:',
+  },
+  line1: {
+    id: 'components.walletinit.importreadonlywalletscreen.line1',
+    defaultMessage: '!!!Open "My wallets" page in the Yoroi extension.',
+  },
+  line2: {
+    id: 'components.walletinit.importreadonlywalletscreen.line2',
+    defaultMessage:
+      '!!!Click on the QR code icon of the wallet you want to ' +
```
SebastienGllmt: As a reminder, I don't think we will use a QR code icon for this button in the Yoroi Extension since we still want in the future a way to export the wallet as file or export the private key (which would probably be from the same menu)

SebastienGllmt (approved):

PR #1161, opened 2021-01-26
API support for Mary

v-almonacid:
~~There is still some work pending, but I'm leaving it for a follow up PR. In particular, probably we need to wrap utxos and account balances (which currently are used directly in the format  given by the backend) in new types that use the multi token class. And then refactor a few functions accordingly.~~

I ended up adding more changes to this PR, basically all the needed functionalities required for Mary (except the UI, which is being working on in a separate PR).

In summary, this PR does the following:
- Introduces the `MultiToken` class for representing amounts (ported from the extension)
- Updates all the transaction API, trying to stay in synch with the extension
- Gets rid of `BaseSignRequest`
- Move to ledgerjs 2.2.0
- Bump Cardano libs

TODO:
- [x] move to new backend API
- [x] test code that uses the new bindings (recall: jest uses the node version)
- [x] flow


src/crypto/types.js
```diff
@@ -14,12 +16,8 @@ export type Address = {|
 |}
 
 export type Value = {|
-  /**
-   * note: an undefined value is different than a value of 0
-   * since you can have a UTXO with a value of 0
-   * which is different from having no UTXO at all
-   */
   +value: void | BigNumber,
+  // +values: MultiToken, // TODO(multi-asset)
```
SebastienGllmt: TODO?

src/components/TxHistory/TxHistoryListItem.js
```diff
@@ -135,10 +144,24 @@ class TxHistoryListItem extends Component<Props> {
...
+
+    const amount: BigNumber = transaction.amount.getDefault()
+    const amountDefaultAsset: ?Token =
+      availableAssets[transaction.amount.getDefaultId()]
+
+    const defaultAsset = amountDefaultAsset || defaultNetworkAsset
+
+    // if we don't have a symbol for this asset, default to ticker first and
+    // then to identifier
+    const assetSymbol = defaultAsset.metadata.ticker
```
SebastienGllmt: Probably this should be a utility function instead?

src/crypto/types.js
v-almonacid: this is okay, it'll be handled in a follow up PR.

v-almonacid:
Note: this shouldn't be merged yet because I'd like to release a hot fix for the ledger issue, and I'd prefer to keep this out.

src/components/Send/SendScreen.js
```diff
@@ -186,36 +227,39 @@ const recomputeAll = async ({amount, address, utxos, sendAll}) => {
...
         const unsignedTx = await getTransactionData(
           utxos,
           address,
           amount,
           sendAll,
+          defaultAsset,
         )
-        _fee = await unsignedTx.fee(false) // in lovelaces
+        _fee = await unsignedTx.fee()
+        // TODO(multi-asset): format according to asset metadata
```
SebastienGllmt, line 240: TODO?

src/config/networks.js
```diff
@@ -201,29 +203,32 @@ export const PRIMARY_ASSET_CONSTANTS = {
...
-          IDENTIFIER: PRIMARY_ASSET_CONSTANTS.CARDANO,
-          IS_DEFAULT: true,
-          METADATA: {
-            TYPE: 'Cardano',
-            POLICY_ID: PRIMARY_ASSET_CONSTANTS.CARDANO,
-            ASSET_NAME: PRIMARY_ASSET_CONSTANTS.CARDANO,
-            TICKER: network.IS_MAINNET ? 'ADA' : 'TADA',
-            LONG_NAME: null,
-            NUMBER_OF_DECIMALS: 6,
+// returns Array<Object>
```
SebastienGllmt, line 206: not sure why this is a comment instead of a Flow type
If the file doesn't

src/crypto/processTransactions.js
```diff
@@ -154,51 +177,57 @@ export const processTxHistoryData = (
...
     )
-    assert.assert(brutto.gte(0), 'Received negative funds')
     direction = TRANSACTION_DIRECTION.RECEIVED
     amount = brutto
     fee = null
   }
 
   const assurance = getTransactionAssurance(tx.status, confirmations)
 
+  // TODO(multi-asset): register token metadata of tokens observed in this tx
```
SebastienGllmt, line 215: TODO?

SebastienGllmt (approved):

v-almonacid:
Note: PR feedback will be addressed in the following PR.

PR #1185, opened 2021-02-02
update readme and remove precompiled targets

v-almonacid:
I also remove the precompiled targets since they are outdated and only include v1 bindings (Byron era).


README.md
```diff
@@ -99,13 +85,14 @@ Make sure the rust targets for the platform you will work on (android/iOS) have
 
 1. `yarn install`
 1. `yarn setup_configs` - links libraries to ios testnet build configurations
+1. When building on iOS: `cd ios && pod install`
```
nicarq, line 88: I would keep this in the iOS. we still mention about installing dependences which is what we do with `pod install`

ashisherc: `pod install` requires `yarn install` to be done first. so we should add a setup step, which does common things at the first

nicarq: ah great point!

nicarq (approved):

nicarq (changes requested):
.

nicarq (approved):

PR #1187, opened 2021-02-03
Update config.yml

nicarq:


v-almonacid:
This produces an error with the v1 bindings when running tests (which uses the node version of the library):

```
The module 'node_modules/node-cardano-wallet/native/index.node'
    was compiled against a different Node.js version using
    NODE_MODULE_VERSION 64. This version of Node.js requires
    NODE_MODULE_VERSION 72. Please try re-compiling or re-installing
```
If I remember correctly those bindings are precompiled in node, causing this issue.

There is also an error building one dependency of emip3js, but at least that can be just removed since we don't use it anymore.

PR #1188, opened 2021-02-03
Additional ledger fixes

v-almonacid:
There was still some legacy ledgerjs code used that was renamed in 2.1.0, and this PR fixes it.

Other changes include:

- Enforce 2.1.0 as the minimum ledger ADA app
- If user doesn't meet the above, show the minimum required version in the error message
- Disable type checking only for `ledgerjs-hw-app-cardano/lib/utils.js.flow` (instead of `Ada.js.flow`)

## On type checking of ledgerjs

Unfortunately, flow types of ledgerjs were included in the [untyped] list, and that's why these errors were not caught. But enabling type checking for this library results in some flow errors which we cannot fix without bumping our flow version. This, in turn, cannot be done without upgrading react native. 
For now, I just put in [untyped] the specific file that is causing issues, and that allow us to enable type checking for most of the relevant API.

### DIFF
--- a/src/crypto/shelley/ledgerUtils.test.js
+++ b/src/crypto/shelley/ledgerUtils.test.js
@@ -1,0 +1,22 @@
+// @flow
+import jestSetup from '../../jestSetup'
+import {checkDeviceVersion, DeprecatedAdaAppError} from './ledgerUtils'
+
+import type {GetVersionResponse} from '@cardano-foundation/ledgerjs-hw-app-cardano'
+
+jestSetup.setup()
+
+describe('encryption/decryption', () => {
+  it('should throw on outdated ledger Ada app', () => {
+    expect.assertions(1)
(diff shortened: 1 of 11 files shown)


src/crypto/shelley/ledgerUtils.test.js
```diff
@@ -0,0 +1,22 @@
+// @flow
+import jestSetup from '../../jestSetup'
+import {checkDeviceVersion, DeprecatedAdaAppError} from './ledgerUtils'
+
+import type {GetVersionResponse} from '@cardano-foundation/ledgerjs-hw-app-cardano'
+
+jestSetup.setup()
+
+describe('encryption/decryption', () => {
```
v-almonacid, line 9: tests still fail because of the node version, but that's fixed in #1161

SebastienGllmt (approved):

PR #1189, opened 2021-02-04
Disable biometric auth for android 10+

v-almonacid:
This PR disables the biometric authentication feature for users on Android 10+ (Android SDK 29+). Note that:

- This should affect fresh installs mainly, as some users may already be using this feature on Android 10+ without problems
- If a user is on Android 10+ and *is not* using biometric auth, the option is disabled so that they cannot opt-in. This is to prevent a scenario in which a user is using pin-based login and then wants to switch to biometrics, and then is unable to access the wallet because the prompt does not show up.
- ⚠️ A consequence of the above is that if a user has no problems using biometric auth right now, and then for some reason they want to opt-out, then they won't be able to opt-in again (at least until we fix this definitively).

Please review carefully.


src/components/Login/styles/AppStartScreen.style.js
```diff
@@ -1,29 +0,0 @@
-// @flow
```
v-almonacid, line 1: this is unrelated, just a file I forgot to remove.

src/config/config.js
```diff
@@ -153,6 +153,7 @@ export const CONFIG = {
   APP_LOCK_TIMEOUT: 120 * 1000,
   LOG_LEVEL: _LOG_LEVEL,
   COMMIT: _COMMIT,
+  ANDROID_BIO_AUTH_EXCLUDED_SDK: [29, 30],
```
SebastienGllmt, line 156: Why do we specify two versions here if everywhere we just do
```
    Platform.Version >= CONFIG.ANDROID_BIO_AUTH_EXCLUDED_SDK[0]
```

SebastienGllmt (approved):

ashisherc (approved):

PR #1198, opened 2021-02-24
Update README.md

nicarq:


README.md
```diff
@@ -162,6 +162,44 @@ file by running command:
 
 ```ln -sf /usr/local/share/automake-<version>/test-driver <path_to_repo>/third-party/glog-0.3.5/test-driver```
 
+# Troubleshooting
+
+## Checking that Rust is well setup (iOS & Android)
+
+After following the instructions for iOS and Android -- it is not required to install both, `rustup show` show return something like this:
```
v-almonacid: >  `rustup show` show return

_should_ return?

v-almonacid (approved):

PR #1202, opened 2021-02-26
add activity indicator in send screen

v-almonacid:
Computing the fee, which requires to create a transaction, is taking more time after the new upgrade in the Cardano libs.

So I added an `<ActivityIndicator>` in the send screen to signal users the fee is being computed. The button is also disabled while the fee is computed.

activity

### DIFF
--- a/src/components/Send/SendScreen.js
+++ b/src/components/Send/SendScreen.js
@@ -4,7 +4,7 @@ import React, {Component} from 'react'
 import {BigNumber} from 'bignumber.js'
 import {compose} from 'redux'
 import {connect} from 'react-redux'
-import {ScrollView, View} from 'react-native'
+import {ScrollView, View, ActivityIndicator} from 'react-native'
 import _ from 'lodash'
 import SafeAreaView from 'react-native-safe-area-view'
 import {injectIntl, defineMessages} from 'react-intl'
@@ -453,6 +453,10 @@ class SendScreen extends Component<Props, State> {
   }
 
   async revalidate({utxos, address, amount, sendAll, selectedAsset}) {
+    this.setState({
+      fee: null,
+      balanceAfter: null,
+    })
     const {defaultAsset, availableAssets, tokenBalance} = this.props
     if (availableAssets[selectedAsset.identifier] == null) {
       throw new Error(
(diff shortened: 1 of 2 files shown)


src/components/Send/SendScreen.js
```diff
@@ -750,12 +754,14 @@ class SendScreen extends Component<Props, State> {
             assetsMetadata={availableAssets}
             unselectEnabled={false}
           />
+          {this.state.fee == null &&
```
nicarq, line 757: @v-almonacid wouldn't the fee change if we change the amount and it requires more utxos?

nicarq: (meaning that we need to display the indicator again)

nicarq, line 763: If the fee is already set, it doesn't show the indicator (from what I see). But if we change the amount it could be changing the number of utxos used so it would require to re-compute and we need to show the indicator

src/components/Send/SendScreen.js
```diff
@@ -750,12 +754,14 @@ class SendScreen extends Component<Props, State> {
...
+            !!this.state.amount && <ActivityIndicator />}
```
nicarq: design wise, can we make the indicator double the size and have an equal margin to its size from the input above?

nicarq: https://reactnative.dev/docs/activityindicator so let's do large

src/components/Send/SendScreen.js
v-almonacid: any time the amount change it recomputes the fee and shows the indicator. The utxos only change when you receive a new utxo, they are not the utxos used for building the tx.

nicarq: @v-almonacid 

a) I didn't see where this nullifies the fee if the amount to send is different
b) About the utxos, when you are going to send X amount you pick UTXOs and you not necessarily use all of your utxos, if you change the amount it may require more UTXOs and that would make the tx bigger in size therefore more expensive

v-almonacid: a) is reset in `revalidate()`

b) Yes. That's the behaviour. Am I missing something?

ashisherc: I just verified that it recomputes the fees as well and nullifies the fees after changing the amount @nicarq

nicarq: @v-almonacid 
a) okay, great
b) I wanted to clarify that statement `The utxos only change when you receive a new utxo`, but if a) is true it should be covered!

v-almonacid: ah ok! I thought you were referring to `this.props.utxos` which is the full utxo set

src/components/Send/SendScreen.js
v-almonacid: done!

ashisherc (approved):

PR #1243, opened 2021-03-08
catalyst support

ashisherc:


src/crypto/shelley/catalystUtils.js
```diff
@@ -0,0 +1,84 @@
...
+  GeneralTransactionMetadata,
+  BigNum,
+} from '@emurgo/react-native-haskell-shelley'
+import {generateMnemonic, mnemonicToEntropy} from 'bip39'
+
+export const CatalystLabels = Object.freeze({
+  DATA: 61284,
+  SIG: 61285,
+})
+export function generateRegistration(request: {|
```
v-almonacid: Heads-up: all the functions here should be `async` and should return `Promise<T>`. Basically all calls to `'@emurgo/react-native-haskell-shelley` API are async, so you'll need to add a few awaits in here. Note also that since the library is incomplete, we don't have flow types (we could, but flow would throw many errors for the types that are missing).

src/crypto/shelley/metadataUtils.js
```diff
@@ -0,0 +1,48 @@
...
+  BigNum,
+  TransactionMetadata,
+} from '@emurgo/react-native-haskell-shelley'
+
+export type JSONMetadata = {|
+  label: string,
+  data: {...},
+|}
+
+export function createMetadata(
```
v-almonacid: The same applies in this module.

src/crypto/shelley/metadataUtils.js
```diff
@@ -0,0 +1,48 @@
...
+  metadata: Array<JSONMetadata>
+): TransactionMetadata {
+  const transactionMetadata = GeneralTransactionMetadata.new()
+
+  metadata.forEach((meta: JSONMetadata) => {
+    const metadatum = encode_json_str_to_metadatum(
```
v-almonacid: Also note that since `encode_json_str_to_metadatum()` is an async call, a `forEach` wouldn't work.

src/crypto/shelley/catalystUtils.js
```diff
@@ -0,0 +1,84 @@
...
+      }),
+      MetadataJsonSchema.BasicConversions,
+    ),
+  )
+
+  return generalMetadata
+}
+
+export function generatePrivateKeyForCatalyst(): Bip32PrivateKey {
+  const mnemonic = generateMnemonic(160)
```
v-almonacid: shouldn't we provide a random seed to this function?

We have a utility function in `crypto/byron/util.js` that's always initialised with a random seed. Also, in general I'd prefer that we don't use hardcoded numbers and instead we use config objects (for this one we have `CONFIG.MNEMONIC_STRENGTH`)

src/components/Catalyst/Step1.js
```diff
@@ -0,0 +1,90 @@
+// @flow
+
+/**
+ * Step 1 for the Catalyst registration - landing
```
v-almonacid, line 4: Just a suggestion for the future: I think naming files/steps with numbers is not always a good a idea because steps can be introduced or removed at random places in time, which messes all the naming up. Better to use a name like `Landing.js` or something more descriptive.

package.json
```diff
@@ -100,7 +101,6 @@
     "redux-logger": "^3.0.6",
     "redux-thunk": "^2.3.0",
     "reselect": "^4.0.0",
-    "safe-buffer": "^5.2.1",
```
v-almonacid: why removing this?

src/components/Catalyst/Step1.js
```diff
@@ -0,0 +1,90 @@
...
+
+const messages = defineMessages({
+  subTitle: {
+    id: 'components.catalyst.step1.subTitle',
+    defaultMessage:
+      '!!!Before you begin, make sure to download the Catalyst Voting App.',
+  },
+})
+
+const Step1 = ({intl, generateVotingKeys, navigation, fetchUTXOs}) => {
```
v-almonacid, line 43: Please add types for all component props. Historically we've been lazy here and sometimes only typed external props, without including all the props that end up injected in the component (in particular, those that come from selectors).

src/components/Catalyst/Step3.js
```diff
@@ -0,0 +1,119 @@
...
+        navigation.navigate(CATALYST_ROUTES.STEP4)
+      } else {
+        showErrorDialog(errorMessages.incorrectPin, intl)
+      }
+    }
+  }
+  const pinCards = (
+    <View style={styles.pinContainer}>
+      {_.range(0, PIN_LENGTH).map((value, index) => {
+        // eslint-disable-next-line react/no-array-index-key
```
v-almonacid: Just curious, why do we need to remove this rule every time?

src/components/Catalyst/Step4.js
```diff
@@ -0,0 +1,217 @@
...
+ * ### HW is NOT supported yet - validation is done on first screen itself###
+ */
+
+import _ from 'lodash'
+import React, {useState} from 'react'
+import {View, SafeAreaView} from 'react-native'
+import {injectIntl, defineMessages} from 'react-intl'
+import {connect} from 'react-redux'
+
+import type {ComponentType} from 'react'
```
v-almonacid: Please keep types grouped at the end, according to our styling rules 🙏

src/components/Catalyst/Step4.js
```diff
@@ -0,0 +1,217 @@
...
+          await walletManager.closeWallet()
+          await showErrorDialog(errorMessages.enableSystemAuthFirst, intl)
+          navigation.navigate(WALLET_ROOT_ROUTES.WALLET_SELECTION)
+
+          return
+        } else {
+          setErrorData((_state) => ({
+            ..._state,
+            showErrorDialog: true,
+            // errorMessage: intl.formatMessage(
```
v-almonacid: TODO?

src/components/Catalyst/Step5.js
```diff
@@ -0,0 +1,235 @@
...
+    unSignedTx.fee().then((o) => {
+      setFees(o.getDefault())
+    })
+  }, [])
+
+  const isConfirmationDisabled = !isEasyConfirmationEnabled && !password
+
+  const submitTrx = async (decryptedKey) => {
+    setSendingTransaction(true)
+    await submitTransaction(unSignedTx, decryptedKey)
```
v-almonacid: if this call fails we can get stuck with `sendingTransaction = true`. Usually it's wrapped in a try-finally.

src/actions/voting.js
```diff
@@ -0,0 +1,73 @@
...
+
+import {encryptWithPassword} from '../crypto/catalystCipher'
+import {generatePrivateKeyForCatalyst} from '../crypto/shelley/catalystUtils'
+import walletManager from '../crypto/walletManager'
+import {Logger} from '../utils/logging'
+import type {RawUtxo} from '../api/types'
+
+import type {State} from '../state'
+
+const _setCatalystKeys = (voting) => ({
```
v-almonacid, line 14: please add types for the payload here (and everywhere in general)

src/components/Catalyst/CatalystNavigator.js
```diff
@@ -0,0 +1,93 @@
...
+      })}
+    />
+    <Stack.Screen
+      name={TX_HISTORY_ROUTES.TX_DETAIL}
+      component={TxDetails}
+      options={({route}) => ({
+        ...defaultNavigationOptions,
+      })}
+    /> */}
+    {/* <Stack.Screen name={CATALYST_ROUTES.STEP1} component={CatalystStep1} /> */}
```
v-almonacid: remove commented code if it's not going to be used

src/components/Catalyst/Step1.js
```diff
@@ -0,0 +1,90 @@
...
+          <Text style={styles.text}>
+            {intl.formatMessage(messages.subTitle)}
+          </Text>
+        </View>
+        <View style={styles.images}>
+          <View style={styles.mb40}>
+            <Image source={AppDownload} />
+          </View>
+          <View style={styles.buttons}>
+            <Image style={styles.iOS} source={appstoreBadge} />
```
v-almonacid: shouldn't these be actual links?

src/components/Catalyst/Step3.js
```diff
@@ -0,0 +1,119 @@
...
+const PIN_LENGTH = 4
+
+const Step3 = ({intl, pin, navigation}) => {
+  const [confirmPin, setPin] = useState('')
+
+  useEffect(() => {
+    generateVotingKeys()
+  }, [])
+
+  const pinChange = (enteredPin) => {
```
v-almonacid, line 51: again explicit types here 🙏

src/components/Catalyst/Step2.js
```diff
@@ -0,0 +1,96 @@
...
+          <Text style={styles.subTitle}>
+            {intl.formatMessage(messages.subTitle)}
+          </Text>
+          <Text style={styles.description}>
+            {intl.formatMessage(messages.description)}
+          </Text>
+          {pinCards}
+        </View>
+        <Button
+          onPress={() => navigation.navigate(CATALYST_ROUTES.STEP6)}
```
v-almonacid: not sure why we go to step 6 here

src/components/Catalyst/Step4.js
```diff
@@ -0,0 +1,217 @@
...
+    try {
+      const decryptedKey = await KeyStore.getData(
+        walletManager._id,
+        'MASTER_PASSWORD',
+        '',
+        password,
+        intl,
+      )
+
+      await generateTransaction(decryptedKey)
```
v-almonacid, line 133: shouldn't `setGeneratingTransaction` set to false after the tx is generated too?

src/components/Catalyst/Step4.js
```diff
@@ -0,0 +1,217 @@
...
+  navigation: Navigation,
+  route: Object, // TODO(navigation): type
+  intl: IntlShape,
+|}
+
+export default injectIntl(
+  connect(
+    (state) => ({
+      isEasyConfirmationEnabled: easyConfirmationSelector(state),
+      utxos: utxosSelector(state),
```
v-almonacid, line 210: I think you need to fetch the UTXOs from the server before. They are not automatically updated in the Store.

src/components/Catalyst/Step6.js
```diff
@@ -0,0 +1,128 @@
...
+  },
+  description: {
+    id: 'components.catalyst.step6.description',
+    defaultMessage:
+      '!!!Backup this QR code or catalyst secret code. This will be required to use the catalyst App.',
+  },
+  note: {
+    id: 'components.catalyst.step6.note',
+    defaultMessage:
+      '!!!Keep it — you won’t be able to access this code after clicking Complete.',
```
v-almonacid: maybe "tapping on complete" since we don't really click

src/components/Catalyst/Step5.js
ashisherc: its handed where `submitTx` is called. But I'm now adding trycatch with finally block

src/components/Catalyst/Step4.js
ashisherc: set to true at the `generateTransaction` func itself, though modified the code to use finally

v-almonacid (approved):

PR #1247, opened 2021-03-12
switch to rubik font

v-almonacid:
This PR sets the default `<Text>` font to Rubik, which is the same font used by the extension.

(I also fixed an issue with the epoch timer display but this is not relevant since the dashboard should be redesigned soon).

# Before

dashboard

send

tx_list

# After

dashboard

send

tx_list


Marta-Ryzhok:
@v-almonacid 
wohoo, cool that font is changed to Rubik! 


In some places, I think thinness is Rubik Bold, can you update it to Rubik Medium, please?
image
image

ashisherc (approved):

PR #1270, opened 2021-04-23
fix wallet creation for small screens

### LINKED ISSUE #1155: Yoroi on Android 10 go
The 15 words appear on the first screen but only 14 on the 2nd , not really sure why? No scroll options

Took some photo evidence and the 2nd screen dropped to 13 xD 
IMG_20210117_030337
IMG_20210117_030214
IMG_20210117_030312
IMG_20210117_030235

New here so dno if I did images right...

wolverineks:
remove unnecessary `SafeAreaView`s
add `ScrollView` to accommodate smaller screens
remove fixed height for mnemonic words

tested on ipod Touch (simulator) and Galaxy Nexus (emulator), both with text and display at max and at default

closes #1155

### DIFF
--- a/src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
+++ b/src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
@@ -143,55 +143,64 @@ const MnemonicCheckScreen = ({
   const last = _.last(partialPhrase)
 
   return (
-    <SafeAreaView style={styles.safeAreaView}>
-      <View style={styles.container}>
-        <StatusBar type="dark" />
+    <SafeAreaView
+      edges={['left', 'right', 'bottom']}
+      style={styles.safeAreaView}
+    >
+      <StatusBar type="dark" />
+
+      <View style={styles.content}>
         <ScrollView
+          contentContainerStyle={styles.scrollViewContentContainer}
           automaticallyAdjustContentInsets={shouldScreenScroll()}
           bounces={shouldScreenScroll()}
         >
-          <View style={styles.content}>
+          <View style={styles.instructions}>
             <Text>{intl.formatMessage(messages.instructions)}</Text>
-            <View
-              style={[
-                styles.recoveryPhrase,
-                !isPhraseValid &&
-                  isPhraseComplete &&
-                  styles.recoveryPhraseError,
-              ]}
-            >
-              {initial.map((id) => (
-                <Text style={styles.wordText} key={id}>
-                  {words[id]}
-                </Text>
-              ))}
-              {last != null && (
-                <WordBadge
-                  value={last}
-                  selected={false}
-                  word={words[last]}
-                  onPress={deselectWord}
-                />
-              )}
-            </View>
-            {!(isPhraseValid || !isPhraseComplete) && (
-              <Text style={styles.error}>
+          </View>
+
+          <View
+            style={[
+              styles.recoveryPhrase,
+              !isPhraseValid && isPhraseComplete && styles.recoveryPhraseError,
+            ]}
+          >
+            {initial.map((id) => (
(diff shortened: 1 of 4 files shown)


src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
```diff
@@ -143,75 +142,79 @@ const MnemonicCheckScreen = ({
   const last = _.last(partialPhrase)
 
   return (
-    <SafeAreaView style={styles.safeAreaView}>
```
v-almonacid, line 146: What's the rationale for removing `SafeAreView`?

wolverineks: my understanding is that `react-navigation` takes care of notches / status bars, etc...
so `<SafeAreaView>` would only be necessary in scenes that do not have a `react-navigation` header

wolverineks: from the `react-navigation` docs
> React Navigation handles safe area in the default header. However, if you're using a custom header, it's important to ensure your UI is within the safe area.

v-almonacid: Actually they only handle it for their components (eg. the navigation bar). Quoting from their docs:

> While React Navigation handles safe areas for the built-in UI elements by default, your own content also needs to handle it to ensure that content isn't hidden by these items.
> 
> (...)
> 
> While React Native exports a SafeAreaView component, it has some inherent issues, i.e. if a screen containing safe area is animating, it causes jumpy behavior. In addition, this component only supports iOS 10+ with no support for older iOS versions or Android. We recommend to use the react-native-safe-area-context library to handle safe areas in a more reliable way.
> 

I don't have many devices to check if this doesn't have unexpected effects, and probably for these screens there shouldn't be any issue, but as a general rule I'd prefer that we always wrap the content inside a `SafeAreaView` from `react-native-safe-area-content` in particular 🙏

wolverineks: my understanding is that `<SafeAreaView>` adds the top/bottom padding regardless if the component would overlap the notches.
for instance, this is what im seeing on the ios simulator
ive added red borders around the `<SafeAreaView>` and `<WalletForm>`
Screen Shot 2021-04-26 at 13 14 30

wolverineks: it doesnt do that on android
Screen Shot 2021-04-26 at 13 17 27

wolverineks: might be related
https://github.com/th3rdwave/react-native-safe-area-context/issues/107

wolverineks: oh i see what's happening.

i agree now, that each of the main scene components should be wrapped in a <SafeAreaView>`, but if the navigator uses a `react-navigation` header or footer, then the `<SafeAreaView>` also needs to specify the `edges` prop  to avoid double padding 
(but only in ios for some weird reason)

ill take a look at this again tomorrow

wolverineks:
alright, `<SafeAreaView>` is back in, as well as removing the duplicated safe area padding at the top of the screens
`<ScrollView>` added to accommodate smaller screens
wrapped each section in a `<View>` for easier styling

v-almonacid:
@wolverineks could you please increase the margin a little bit here 🙏 
image

The design specs show a 32px margin, but let me know if it looks like too much/too low/

Edit: wrong image

v-almonacid (approved):
I was able to reproduce and confirm the fix on a real device setting the text size to the maximum size.

Thank you @wolverineks !

PR #1276, opened 2021-04-29
use staking key for voting rewards

v-almonacid:
TODO:

- [x] disable voting or add prompt for users not delegating

### DIFF
--- a/src/components/Catalyst/Step1.js
+++ b/src/components/Catalyst/Step1.js
@@ -4,7 +4,7 @@
  * Step 1 for the Catalyst registration - landing
  */
 
-import React, {useEffect} from 'react'
+import React, {useEffect, useState} from 'react'
 import {
   View,
   ScrollView,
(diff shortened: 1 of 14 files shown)


src/crypto/shelley/catalystUtils.js
```diff
@@ -49,7 +50,9 @@ export async function generateRegistration(request: {|
       '2': `0x${Buffer.from(
         await (await request.stakePrivateKey.to_public()).as_bytes(),
       ).toString('hex')}`,
-      '3': `0x${Buffer.from(request.receiverAddress).toString('hex')}`,
+      '3': `0x${Buffer.from(await request.rewardAddress.to_bytes()).toString(
```
v-almonacid: I still don't know if this should be the raw keyhash or the reward address bytes

ashisherc: this is correct

ashisherc (approved):

src/components/Catalyst/Step1.js
```diff
@@ -39,6 +41,13 @@ const messages = defineMessages({
     defaultMessage:
       '!!!Before you begin, make sure to download the Catalyst Voting App.',
   },
+  stakingKeyNotRegistered: {
+    id: 'components.catalyst.step1.stakingKeyNotRegistered',
+    defaultMessage:
+      '!!!Catalyst voting rewards are sent to delegation accounts and your ' +
+      'wallet does not seem to have a registered delegation certificate. If ' +
+      'you want to receive voting rewards, you need to delegate your funds first.',
+  },
```
v-almonacid, line 50: @yushih @vantuz-subhuman this is the message I put in the confirmation dialog

PR #1330, opened 2021-05-08
add spinner to delegation dashboard

### LINKED ISSUE #1261: Unable to scroll dashboard screen while loading in progress
From the dashboard screen sometimes it can take a while for all of the details to load. In particular it can take around 10 seconds for my stake pool delegation info to load. While this loading is happening and the loading indicator is being shown I am unable to scroll down which is a slightly irritating user experience. I would suggest that the loading indicator should be displayed per item on this page. (i.e. one loading indicator each for "epoch progress", "your summary" and "stake pool delegated") and I should still be able to scroll on this page even if the data is not ready to be displayed yet. Regardless thanks for open sourcing this project, keep up the good work!

wolverineks:
closes #1261

### DIFF
--- a/src/components/Delegation/StakingDashboard.js
+++ b/src/components/Delegation/StakingDashboard.js
@@ -4,7 +4,13 @@ import React from 'react'
 import type {ComponentType} from 'react'
 import {connect} from 'react-redux'
 import {compose} from 'redux'
-import {View, ScrollView, RefreshControl, Platform} from 'react-native'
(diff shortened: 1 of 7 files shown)


wolverineks:
| before | after |
|--------|------|
| before | after |

v-almonacid (approved):

v-almonacid:
Thanks @wolverineks !

PR #1381, opened 2021-06-02
add flow types for react-navigation

wolverineks:
add react-navigation types
add navigator routes types

`<Navigator.Screen>` no longer needs route names pulled from `'RoutesList.js'`

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,6 @@
     "@react-native-community/masked-view": "0.1.10",
     "@react-native-community/netinfo": "5.9.9",
     "@react-navigation/bottom-tabs": "5.11.8",
-    "@react-navigation/drawer": "5.12.5",
     "@react-navigation/material-top-tabs": "5.3.10",
     "@react-navigation/native": "5.9.3",
     "@react-navigation/stack": "5.14.5",
--- a/src/components/Delegation/StakingDashboard.js
+++ b/src/components/Delegation/StakingDashboard.js
@@ -717,10 +717,11 @@ class StakingDashboard extends React.Component<Props, State> {
   }
 }
 
-type ExternalProps = {
+type ExternalProps = {|
   navigation: Navigation,
+  route: any,
   intl: IntlShape,
-}
+|}
 
 export default injectIntl(
   (compose(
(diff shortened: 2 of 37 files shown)


package.json
```diff
@@ -40,7 +40,6 @@
     "@react-native-community/masked-view": "0.1.10",
     "@react-native-community/netinfo": "5.9.9",
     "@react-navigation/bottom-tabs": "5.11.8",
-    "@react-navigation/drawer": "5.12.5",
```
v-almonacid, line 43: why removing this?

src/components/Delegation/StakingDashboard.js
```diff
@@ -717,10 +717,10 @@ class StakingDashboard extends React.Component<Props, State> {
   }
 }
 
-type ExternalProps = {|
+type ExternalProps = {
```
v-almonacid: why some props need to be unsealed?

wolverineks: react-navigation passes in `Route` and `Navigation`
it is a flow error to pass in a prop that is not specified in a sealed prop type

package.json
wolverineks: there are no imports for `@react-navigation/drawer`
why keep it around?

src/components/Delegation/StakingDashboard.js
v-almonacid: mhm then probably better to add `Route` in the types object and keep the object sealed

wolverineks: the only place in the flow docs i see sealed props for react components is defaultProps for class components
all the other instances in the flow docs use unsealed props

what are the benefits of sealed props?

wolverineks: and you would have to pass those props in for storybook, even though the component doesn't use them

v-almonacid: Using exact objects is recommended as a general rule. It may prevent some bugs. Also, I believe they wanted to use exact objects as the default behaviour in the future.

for storybook there's already a decorator that passes on mocks of `navigation` and `route` to every component

wolverineks: what bugs could it prevent?

v-almonacid: For instance if during development you instantiate a component with an extra prop, and you haven't added the code that handles that prop yet, you won't see any error if props are not sealed.

v-almonacid (approved):

PR #1393, opened 2021-06-07
remove react-native-permissions

wolverineks:


v-almonacid:
I think this one was added because in the installation  instructions of `react-native-qrcode-scanner` , they say it should be added. However, it's already included in their dependencies. 

Can you double check that there are no changes involved in Podfile/Podfile.lock?

PR #1398, opened 2021-06-09
update yarn lock

wolverineks:
i forgot to update yarn lock after pinning flow-coverage-report


v-almonacid (approved):

v-almonacid:
Weird, I already updated it. Not sure why the diff doesn't catch it 🤔

PR #1399, opened 2021-06-09
add @react-navigation/routers

wolverineks:
i forgot to add @react-navigations/routers when resetting the send screen
it still worked because @react-navigation/native  requires @react-navigation/core which requires @react-navigation/routers


v-almonacid (approved):

v-almonacid:
@wolverineks please fix the merging conflict when you have time

wolverineks:
@v-almonacid

PR #1402, opened 2021-06-13
add eslint-react-hooks

wolverineks:


wolverineks:
@v-almonacid

package.json
```diff
@@ -127,6 +127,7 @@
     "eslint-plugin-flowtype": "3.12.1",
     "eslint-plugin-import": "2.22.1",
     "eslint-plugin-react": "7.21.5",
+    "eslint-plugin-react-hooks": "^4.2.0",
```
v-almonacid: please remember to use exact versions

src/App.js
```diff
@@ -19,33 +19,35 @@ enableScreens()
...
-      RNBootSplash.show()
-    }
-  }
-
   // note: previously this was hanlded in the applicationDidEnterBackground
   // event in AppDelegate.m, but after moving to a .storyboard launch screen
   // that solution didn't seem feasible anymore.
-  if (Platform.OS === 'ios') {
-    useEffect(() => {
+  // eslint-disable-next-line consistent-return
```
v-almonacid: I would avoid removing rules (otherwise what's the point on adding them?). I think we can just return an empty clean up function in the other case.

src/components/Catalyst/Step4.js
```diff
@@ -183,6 +183,7 @@ const Step4 = ({
     if (isEasyConfirmationEnabled) {
       onContinue()
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v-almonacid: I think the rule is right here, `isEasyConfirmationEnabled` should be in the dep list, don't you think?

src/components/Catalyst/Step5.js
```diff
@@ -112,6 +112,7 @@ const Step5 = ({
     unSignedTx.fee().then((o) => {
       setFees(o.getDefault())
     })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v-almonacid: same for unsignedTx, it should probably be in the deps list

src/components/Common/PinRegistrationForm.js
```diff
@@ -86,6 +86,7 @@ const PinRegistrationForm = ({
       })
       return unsubscribe
     },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v-almonacid: again, `clearPin` being a prop, it should probably be in the deps list (even though it should never change).

src/components/Send/BiometricAuthScreen.js
```diff
@@ -150,40 +150,43 @@ const BiometricAuthScreen = (
...
+        // the app triggers a new biometric prompt; but may also be an issue for
+        // some specific Android versions
+        await KeyStore.cancelFingerprintScanning(KeyStore.REJECTIONS.CANCELED)
+      }
+    }
+
     AppState.addEventListener('change', handleAppStateChange)
 
     return () => AppState.removeEventListener('change', handleAppStateChange)
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v-almonacid: not sure why it complains here 🤔

src/components/TxHistory/TxHistory.js
```diff
@@ -163,6 +163,7 @@ const TxHistory = ({
 
   useEffect(() => {
     fetchAccountState()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v-almonacid: again, maybe just add `fetchAccountState`?

src/components/WalletInit/RestoreWallet/SaveReadOnlyWalletScreen.js
```diff
@@ -156,17 +156,18 @@ const SaveReadOnlyWalletScreen = (
...
+      const {addresses, accountPlate} = await generateShelleyPlateFromKey(
+        publicKeyHex,
+        1,
+        networkId,
+      )
+      setPlate({addresses, accountPlate})
+    }
+
     generatePlates()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v-almonacid: I guess it complains because `publicKeyHex` and `networkId` come from a prop? If that's the case, maybe we should just add them to the deps list.

src/components/WalletInit/RestoreWallet/WalletAddress.js
```diff
@@ -36,6 +36,7 @@ const FadeOutView = (props) => {
       delay: 3000,
       useNativeDriver: true,
     }).start()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
v-almonacid: again, consider adding `fadeAnim` in the deps list

v-almonacid (changes requested):
Really nice to have this. Minor details to review and then we are good to merge

src/components/Catalyst/Step4.js
wolverineks: this change has some cascading effects
ill follow the rules, but it's gonna need some manual verification

wolverineks:
@v-almonacid

v-almonacid (approved):

PR #1404, opened 2021-06-14
upgrade flow

wolverineks:
v0.122.0 is over a year old
flow@latest catches >700 errors
there were some significant changes to flow at v0.127.0
upgrading beyond v0.126.0 should be in a different pr

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "eslint-plugin-react": "7.21.5",
     "eslint-plugin-react-native": "3.10.0",
     "extract-react-intl-messages": "^0.14.0",
-    "flow-bin": "0.122.0",
+    "flow-bin": "0.126.0",
     "flow-coverage-report": "0.8.0",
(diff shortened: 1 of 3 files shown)


v-almonacid:
Historically we upgrade flow when we upgrade react native, because usually upgrading flow reveals errors in dependencies, including react native (as you may have noticed). But upgrading to this version looks okay (just remember to remove that file from the ignore list later on).

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "flow": "flow --show-all-errors .",
+    "flow:all": "flow --show-all-errors .",
```
v-almonacid: why changing this? this will break the CI I guess

wolverineks: fixed

wolverineks:
whats the ignore list?
nvm

v-almonacid (approved):

PR #1406, opened 2021-06-16
enable storybook browser ui, fix actions

wolverineks:
`yarn storybook`: automatically runs `prestorybook` generating all the stories, then launches browser to control simulator
`asyncStorage` enables returning to last used story when reloading simulator


wolverineks:
@v-almonacid

v-almonacid (approved):

PR #1415, opened 2021-06-22
Stackchain/Emurgo Q2 UI updates [Receive Screen]

stackchain:


stackchain:
What has changed so far:

https://user-images.githubusercontent.com/30806844/123264295-458bad80-d4d0-11eb-818b-13249afc9593.mp4

src/crypto/Address.dto.js
```diff
@@ -0,0 +1,98 @@
...
+ * @returns KeyHashesCardano|null You don't need to test for Byron it will return null
+ */
+export const getKeyHashesCardano = async (
+  wasmAddress: Address,
+): Promise<KeyHashesCardano | null> => {
+  let addr, pay, stake, paymentKeyHash, stakingKeyHash
+  for (const [i, addressType] of ADDRESS_TYPES.entries()) {
+    // First = Byron
+    if (!i) {
+      const addr = addressType.from_address(wasmAddress)
```
v-almonacid: I think there is an `await` missing here

src/crypto/Address.dto.js
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * @description The resolver for the Cardano key hashes
+ * @param {Address} wasmAddress A Cardano wallet Address
+ * @returns KeyHashesCardano|null You don't need to test for Byron it will return null
+ */
+export const getKeyHashesCardano = async (
+  wasmAddress: Address,
+): Promise<KeyHashesCardano | null> => {
+  let addr, pay, stake, paymentKeyHash, stakingKeyHash
+  for (const [i, addressType] of ADDRESS_TYPES.entries()) {
```
v-almonacid: Probably instead of relying on a for sequence that can be broken if the `ADDRESS_TYPES` object is modified, it might be better to use the `normalizeToAddress` function

src/crypto/Address.dto.js
```diff
@@ -0,0 +1,98 @@
+// @flow
+
+import {
+  Address,
+  ByronAddress,
+  BaseAddress,
+  PointerAddress,
```
v-almonacid: Heads up: `PointerAddress` and `EnterpriseAddress` are not implemented in our native bindings.

src/crypto/Address.dto.js
```diff
@@ -0,0 +1,98 @@
...
+  EnterpriseAddress,
+  RewardAddress,
+} from '@emurgo/react-native-haskell-shelley'
+import {Logger} from '../utils/logging'
+
+// NOTE: ** IMPORTANT ** The order matters
+const ADDRESS_TYPES = Object.freeze([
```
v-almonacid: Heads up: `react-native-haskell-shelley` is disabled in flow, so you can't rely on types (everything is just `any`). You can enable it but flow's report will be polluted by many errors unrelated to your code, which come from the fact that `react-native-haskell-shelley` does not cover the whole API given by `cardano-serialization-lib`.

src/i18n/locales/en-US.json
```diff
@@ -78,18 +78,26 @@
...
   "components.receive.addressmodal.copyLabel": "Copy address",
-  "components.receive.addressmodal.walletAddress": "Your wallet address",
+  "components.receive.addressmodal.spendingKeyHash": "Spending key hash",
+  "components.receive.addressmodal.stakingKeyHash": "Staking key hash",
+  "components.receive.addressmodal.title": "Verify address",
+  "components.receive.addressmodal.walletAddress": "Address",
   "components.receive.addressverifymodal.afterConfirm": "Once you tap on confirm, validate the address on your Ledger device, making sure both the path and the address match what is shown below:",
   "components.receive.addressverifymodal.title": "Verify Address on Ledger",
+  "components.receive.addressview.copiedAddressLabel": "Copied",
+  "components.receive.addressview.copyAddressLabel": "Copy address",
```
v-almonacid: It might be a good idea to reuse strings that are already there, as any new key must be translated and we rely on volunteers from the community. In this case for instance you could just use `components.receive.addressmodal.copyLabel`

src/crypto/Address.dto.js
stackchain: The object is frozen, normalize will return only the address.

src/crypto/Address.dto.js
stackchain: Roger that, commenting for now.

src/crypto/Address.dto.js
stackchain: Roger that, removing the typing, setting as `any`.

src/i18n/locales/en-US.json
stackchain: Roger that, gonna be moving the most frequent texts to global later.

stackchain:
Final version

https://user-images.githubusercontent.com/30806844/123886515-04f3c000-d926-11eb-9fef-fc25424fa8b9.mp4

src/components/Receive/AddressModal.js
```diff
@@ -14,11 +14,17 @@ import {
   walletMetaSelector,
 } from '../../selectors'
 import {formatPath} from '../../crypto/commonUtils'
-
 import {Text, Button, Modal} from '../UiKit'
 
 import styles from './styles/AddressModal.style'
-
+import copyIcon from '../../assets/img/icon/copy-ext.png'
+// TODO: replace to the right icon after getting it from UI department
```
v-almonacid: is this meant to be done in a follow up PR?

src/components/Receive/AddressVerifyModal.js
```diff
@@ -56,7 +57,7 @@ const AddressVerifyModal = ({
       </Text>
       <View style={styles.addressDetailsView}>
         <Text secondary style={styles.paragraph}>
-          {address}
+          {addressInfo.address}
```
v-almonacid, line 60: Ledger devices show the staking key path in the verification flow, so it's something you should probably consider adding.

src/components/WalletNavigator.js
```diff
@@ -110,7 +110,7 @@ const WalletTabNavigator = injectIntl(
       isReadOnly: any,
     }) => (
       <Tab.Navigator
-        initialRouteName={WALLET_ROUTES.TX_HISTORY}
+        initialRouteName={WALLET_ROUTES.DASHBOARD}
```
v-almonacid: What happens if it's a Byron wallet (for which the `DASHBOARD` tab won't be rendered)?

src/crypto/Address.dto.js
```diff
@@ -0,0 +1,97 @@
...
+  throw new Error('getSpendingKeyHash unknown address type')
+}
+
+export interface YoroiAddressInfoInterface<KeyHashes> {
+  +address: string;
+  +getKeyHashes: () => Promise<KeyHashes | null>;
+}
+
+// prettier-ignore
+export class AddressDTOCardano
```
v-almonacid: Note that there are plans to integrate Ergo so in general we try to organise the code keeping in mind that other wallet types and networks might be added in the future. In this case I'd have probably defined the interface inside `crypto` but all the ADA-specific logic inside `crypto/shelley/`. That folder is for logic specific to Shelley wallets from the "Haskell Shelley" era.

src/components/WalletNavigator.js
stackchain: Tab.Navigator will default to the next route, but I think is better to make it explicit, so I'll be fixing it.

src/crypto/Address.dto.js
stackchain: Totally agree.

src/components/Receive/AddressVerifyModal.js
stackchain: Added it to Asana.

zuzunker (approved):

PR #1421, opened 2021-06-30
Remove Byron-era rust bindings and related logic

v-almonacid:
This PR removes the Byron-era library `react-native-cardano` and related deprecated logic. Restoring Byron wallets is still possible since I've added support for deriving Byron addresses (Icarus) in `react-native-haskell-shelley`. Similarly, functions for encryption/decryption were added recently in the Haskell Shelley library and we consume those from now on.

Some cryptographic primitives, like generating a HD wallet root key, were still being sourced from the old Byron bindings, and have been replaced in this PR. I added some unit tests to make sure the new functions work as the previous ones.

There are no functional changes in this PR, but since functions required for creating and restoring wallets have been replaced,  careful review is needed. 

Some positive implications:
- installing packages should be faster
- Compilation time is reduced significantly
- We should now be able to upgrade the nodeJS version


stackchain (approved):
LGTM

package.json
```diff
@@ -113,7 +112,7 @@
     "vm-browserify": "1.1.2"
   },
   "devDependencies": {
-    "@emurgo/cardano-serialization-lib-nodejs": "6.0.0",
+    "@emurgo/cardano-serialization-lib-nodejs": "7.1.0",
```
wolverineks, line 115: should this be in dependencies instead of devDependencies?

src/crypto/walletManager.js
```diff
@@ -869,38 +868,28 @@ class WalletManager {
...
+      'Ae2tdPwUPEZKAx4zt8YLTGxrhX9L6R8QPWNeefZsPgwaigWab4mEw1ECUZ7',
+      'Ae2tdPwUPEZAghGCdQykbGxc991wdoA8bXmSn7eCGuUKXF4EsRhWj4PJitn',
+      'addr1qynqc23tpx4dqps6xgqy9s2l3xz5fxu734wwmzj9uddn0h2z6epfcu' +
+        'kqmswgwwfruxh7gaddv9x0d5awccwahnhwleqqc4zkh4',
+      'addr1q9tr0a0feutyhdj34gxnasv8vef699fcry5avyrt6hn4n540f7le3l' +
+        'aqc6cgpcds86z06psxczmnuk7txsajs4jdt4nqlhj8aa',
+    ]
+    const externalChain = this._wallet.externalChain
+    const match = addrs.indexOf(externalChain.addresses[0])
+    if (match !== -1) {
```
wolverineks: this is more readable
`if(addrs.includes(externalChange.addresses[0])) {`

src/crypto/walletManager.js
```diff
@@ -869,38 +868,28 @@ class WalletManager {
...
-      if (this._wallet == null) throw new WalletClosed()
-      const externalChain = this._wallet.externalChain
-      Logger.debug(
-        'WalletManager::checkForFlawedWallets wallet addresses [0]:',
-        externalChain.addresses[0],
-      )
+  checkForFlawedWallets(): boolean {
+    if (CONFIG.IS_TESTNET_BUILD) return false
+    if (this._wallet == null) throw new WalletClosed()
+    const addrs = [
```
wolverineks, line 874: where do these come from?

package.json
v-almonacid: nope

src/crypto/walletManager.js
v-almonacid: those are just dev wallets

package.json
zuzunker: @wolverineks I think this library is only used for Flow types, that's why it's in devs. It is never used for any actual functionality. This lib is the WASM version of the serialization-lib, which as we discussed we cannot use in the mobiles directly, that's why we have this react-native library in regular dependency, that is the native-wrapped version of that same serialization-lib:

https://github.com/Emurgo/yoroi-mobile/blob/ef47d293a96cb8c4c1139d1fce3b3ed1c3ba28c0/package.json#L37

src/crypto/walletManager.js
zuzunker: Good point!

zuzunker: Added a small change, plz check: https://github.com/Emurgo/yoroi-mobile/pull/1421/commits/b14d9ad613708012ab737adf222b11206d27a940

package.json
v-almonacid: Sorry, now I can answer more in detail. The nodejs libraries are used both in the extension and mobile for jest, not for flow, since jest uses a node runtime environment. 
As a side note, it's important you keep this in mind when adding tests that consume the rust libs: If your test uses functions that are not in `react-native-haskell-shelley`, or that have bugs, you won't notice it because you are actually consuming the nodeJS version (cc @stackchain).

src/crypto/walletManager.js
v-almonacid: better!

zuzunker (approved):

zuzunker (commented):
/check

zuzunker (commented):
/check

PR #1447, opened 2021-07-06
Stackchain/Update Alonzo auxiliary data

stackchain:


zuzunker (commented):
/check

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -570,7 +570,7 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
     tokens: SendTokenList,
     defaultToken: DefaultTokenEntry,
     serverTime: Date | void,
-    metadata: Array<JSONMetadata> | void,
+    auxiliaryData: Array<JSONMetadata> | void,
```
wolverineks, line 471: why call this auxiliaryData when there is an AuxiliaryData type? would it make sense to name this something else?

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -787,15 +789,15 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
         [], // no delegations
         [], // no withdrawals
         false,
-        metadata,
+        auxiliaryData,
       )
 
       // prettier-ignore
```
wolverineks: probably don't need this anymore

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -942,7 +944,7 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
         request.ledgerNanoCatalystRegistrationTxSignData,
       )
 
-      metadata = await catalystUtils.generateRegistration({
+      auxiliaryData = await catalystUtils.generateRegistration({
```
wolverineks: why does `generateRegistration` return an `AuxiliaryData`?

src/crypto/shelley/ShelleyWallet.js
```diff
@@ -912,7 +914,7 @@ export default class ShelleyWallet extends Wallet implements WalletInterface {
       useUSB,
     )
 
-    let metadata
+    let auxiliaryData
```
wolverineks: what type is this?

src/crypto/shelley/catalystUtils.js
```diff
@@ -90,15 +90,15 @@ export async function generateRegistration(request: {|
...
   )
-  const trxMetadata = await TransactionMetadata.from_bytes(
+  const auxiliary = await AuxiliaryData.from_bytes(
     await metadataList.to_bytes(),
   )
-  return trxMetadata
+  return auxiliary
 }
 
 // prettier-ignore
```
wolverineks: probably don't need this anymore

src/crypto/shelley/catalystUtils.test.js
```diff
@@ -50,7 +50,7 @@ test('Generate Catalyst registration tx', async () => {
   }
 
   const nonce = 1234
-  const txMetaData = await generateRegistration({
+  const metadata = await generateRegistration({
```
wolverineks: why does `generateRegistration` return metadata?

src/crypto/shelley/catalystUtils.test.js
```diff
@@ -59,9 +59,9 @@ test('Generate Catalyst registration tx', async () => {
   })
 
   const result = await GeneralTransactionMetadata.from_bytes(
-    await (await (await MetadataList.from_bytes(
-      await txMetaData.to_bytes(),
-    )).get(0)).to_bytes(),
+    await (await (await MetadataList.from_bytes(await metadata.to_bytes())).get(
```
wolverineks: would this be more readable if assigned to intermediate variables?

src/crypto/shelley/transactionUtils.js
```diff
@@ -103,7 +103,7 @@ export const createUnsignedTx = async (
         [], // no certificates
         [], // no withdrawals
         false, // do not allow no outputs
```
wolverineks, line 91: you don't have to fix it  but I just wanted to point out that this line is....ummm..... what the what?!?

wolverineks (changes requested):
* remove comments if unused
* improve readability when possible
* review variable names

src/crypto/shelley/ShelleyWallet.js
stackchain: Hey @wolverineks, when building cardano objects we always have 2 stages, 1st JS thing (type JSONMetadata), and 2nd we serialize all objects as late as possible, so the data that we ingest at first as JSON will end up on the other side in our bindings (Buffer) serialized as Ptr. The AuxiliaryData that you see is actually Ptr from the bindings, if we convert to early we will lose all JS thing, so far it makes sense to keep this way, what would you suggest?

src/crypto/shelley/ShelleyWallet.js
stackchain: ?

src/crypto/shelley/ShelleyWallet.js
stackchain: From now on we are also trying to respect the semi-formal implementation, you can read the specs on 
https://hydra.iohk.io/build/6752480/download/1/alonzo-changes.pdf (page 10, fig 3)

src/crypto/shelley/ShelleyWallet.js
stackchain: Because Registration is one of the possible metadata that is carried inside the auxiliary data, we have the same implementation on yoroi-frontend 
https://github.com/Emurgo/yoroi-frontend/blob/7021ffb4f37ca28b058fcd5ece99ce8c5741bfe9/packages/yoroi-extension/app/api/ada/lib/cardanoCrypto/catalyst.js
Would you suggest a better approach?

src/crypto/shelley/ShelleyWallet.js
stackchain: It is an AuxiliaryData type, but heads up ->  react-native-haskell-shelley is disabled in flow, so you can't rely on it, therefore everything is just `any`. If you enable it, tons of errors will raise, due to the fact that react-native-haskell-shelley does not cover the whole API given by cardano-serialization-lib. I could set it to AuxiliaryData, but will be only decorative and not restrictive.

src/crypto/shelley/catalystUtils.test.js
stackchain: Check above explanation.

src/crypto/shelley/catalystUtils.test.js
stackchain: I see no problems with data piping/fn composing, perhaps we should discuss with @vsubhuman if we want to go fully procedural and refactor everywhere.

src/crypto/shelley/transactionUtils.js
stackchain: What would you suggest?

src/crypto/shelley/ShelleyWallet.js
wolverineks: my issue is about variable names, 
`registration` vs `metadata` vs `auxiliaryData`
generateRegistration should return something called registration

src/crypto/shelley/catalystUtils.test.js
wolverineks: `generateRegistration` should return something called `registration`

src/crypto/shelley/ShelleyWallet.js
wolverineks: wait, why does `generateRegistrationMetadata` return `RustModule.WalletV4.AuxiliaryData`?

stackchain: yap, it doesn't look intuitive, perhaps a better approach would be to build a "factory" of auxiliary data which receives what it wants to create and return it wrapped `buildAuxiliaryDataWith<T>(metadata: T): AuxiliaryData<{| metadata: T |}> =>`

src/crypto/shelley/metadataUtils.js
```diff
@@ -8,23 +8,23 @@ import {
...
 
-  for (const meta of metadata) {
+  for (const meta of auxiliary) {
     const metadatum = await encode_json_str_to_metadatum(JSON.stringify(meta.data), MetadataJsonSchema.BasicConversions)
-    await transactionMetadata.insert(await BigNum.from_str(meta.label), metadatum)
+    await metadata.insert(await BigNum.from_str(meta.label), metadatum)
   }
 
-  return await TransactionMetadata.new(transactionMetadata)
+  return await AuxiliaryData.new(metadata)
```
yushih, line 27: Are you sure that the constructor of `AuxiliaryData` takes a parameter of `GeneralTransactionMetadata`?

Seems the constructor takes no parameter. And the `AuxiliaryData` object has a `set_metadata` method that takes a `GeneralTransactionMetadata`.

I'm not familiar with `react-native-haskell-shelley` to say for sure. But if this passes flow it's probably correct.

yushih:
Please check in `src/crypto/shelley/transactions.test.js` whether `witnesses.scripts()` needs to be changed to `witnesses..native_scripts()`.

yushih (approved):

src/crypto/shelley/metadataUtils.js
stackchain: Hey @yushih, you can't rely on flow for the bindings because is not 100% covered, and it is been ignored for now. You are correct, the implementations are different, nice catch. You can check it on https://github.com/Emurgo/react-native-haskell-shelley/blob/develop-alonzo/rust/src/android/auxiliary_data.rs#L56 it is not on main yet.

PR #1448, opened 2021-07-07
Catalyst registration password fix

zuzunker:
- Added password to callback dependencies in Catalyst registration step4
- Plus added using safe-navigation in tokens metadata

### DIFF
--- a/src/components/Catalyst/Step4.js
+++ b/src/components/Catalyst/Step4.js
@@ -174,7 +174,7 @@ const Step4 = ({
       }
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [],
+    [password],
   )
 
   useEffect(
--- a/src/utils/format.js
+++ b/src/utils/format.js
@@ -46,21 +46,21 @@ export const getAssetDenomination = (
 ): ?string => {
   switch (denomination) {
     case ASSET_DENOMINATION.TICKER:
-      return token.metadata.ticker
+      return token?.metadata?.ticker
     case ASSET_DENOMINATION.SYMBOL:
       // if we don't have a symbol for this asset, default to ticker, though
       // ticker can still be null
       // prettier-ignore
-      return token.metadata.ticker
+      return token?.metadata?.ticker
         ? utfSymbols.CURRENCIES[token.metadata.ticker]
           ? utfSymbols.CURRENCIES[token.metadata.ticker]
           : token.metadata.ticker
         : null
     case ASSET_DENOMINATION.NAME: {
-      if (token.metadata.longName !== null) {
+      if (token?.metadata?.longName !== null) {
         return token.metadata.longName
       }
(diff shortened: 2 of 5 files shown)


stackchain (approved):

src/utils/format.js
```diff
@@ -46,21 +46,21 @@ export const getAssetDenomination = (
 ): ?string => {
   switch (denomination) {
     case ASSET_DENOMINATION.TICKER:
-      return token.metadata.ticker
+      return token?.metadata?.ticker
```
wolverineks, line 49: token is guaranteed by flow, there is no need for `token?.`

src/components/Catalyst/Step4.js
```diff
@@ -174,7 +174,7 @@ const Step4 = ({
       }
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [],
+    [password],
```
wolverineks, line 177: what happens if you add all the dependencies?

src/utils/format.js
zuzunker: It was failing when I have been running it on simulator 🤔 It might be some kind of my local problem, but it looked like the metadata was non-existing when the send page was just opened for the first time, as no value have been selected in the drop-down yet. I will try to look into removing uneccessary checks later

src/components/Catalyst/Step4.js
zuzunker: @wolverineks , what other dependencies are critical inside that callback? 🤔 I couldn't figure anything except the password. I see it's using the setters from other declared states (like `setGeneratingTransaction` and `setErrorData`), but never actually refers to the values.

src/utils/format.js
zuzunker: I will do some investigation into whether no selected value in the drop-down will cause the `token` to be empty as well, if not - will remove the checks and see why it could have been failing in my simulator. But if yes - will review the flow types to make sure that's reflected there

stackchain: @vantuz-subhuman it was happening to me as well, but only at first run.

zuzunker: hmmmmmmm 🤔

src/components/Catalyst/Step4.js
wolverineks: including all the dependencies should be the default mindset, and excluding any deserves scrutiny
my preference is to figure out why something is excluded, and try to solve the problem before it gets to `useEffect`
that might mean defining functions in `useCallback` or defining them directly in `useEffect`

zuzunker: @wolverineks , good point! This PR was in response to a specific found issue, but I would have no objections to have another PR from you improving the code wherever necessary 👌

src/utils/format.js
wolverineks: @vantuz-subhuman did you find anything out yet?

PR #1458, opened 2021-07-15
remove storyLoader from git

wolverineks:
storyLoader is automatically generated when running `yarn storybook`


zuzunker (approved):

zuzunker (changes requested):
@wolverineks , flow is failing because the storybook-loader import cannot be resolved now:

image

Gotta either exclude storybook index from Flow, or mock the flow types for the loader.

PR #1472, opened 2021-07-20
move tos, redux -> component state

wolverineks:
the less redux, the better
we could also add `onPressIn={() => loadTOS(languageCode)}` to the settings button to reduce the loading time


wolverineks:
closes #685

zuzunker (approved):

PR #1478, opened 2021-07-21
Stackchain/Started refactoring to allow multiple backend providers

stackchain:
- added the provider property to the wallet
- every shelley implementation may have a provider (empty will fallback to default)
- in nightly or test build will show the add alonzo wallet
- in nightly or test build will show the provider in my wallet list


stackchain:
- update the ui with provider info
- update the tx detail to allow copy the tx.id and open the explorer

b2beb59ae263c8e1cf8d7f2311a96e0d
9f8a66becaba881411bad30933d8308b

src/components/UiKit/Copy.js
```diff
@@ -0,0 +1,26 @@
...
+import copyIcon from '../../assets/img/icon/copy-ext.png'
+import copiedIcon from '../../assets/img/icon/copied.png'
+
+export type CopyProps = {|
+  value: string,
+|}
+
+export const CopyView = ({value}: CopyProps) => {
+  const [isCopying, copy] = useCopy()
+  const _copyHandler = useCallback(() => {
```
wolverineks: why the leading underscore?

wolverineks: you dont need a `useCallback` here
use `() => copy(value)` in the button

src/components/UiKit/Copy.js
```diff
@@ -0,0 +1,26 @@
...
+
+import {useCopy} from '../../utils/useCopy'
+import copyIcon from '../../assets/img/icon/copy-ext.png'
+import copiedIcon from '../../assets/img/icon/copied.png'
+
+export type CopyProps = {|
+  value: string,
+|}
+
+export const CopyView = ({value}: CopyProps) => {
```
wolverineks: component name, file name, and prop type should match
my preference is `CopyButton.js` and `export const CopyButton` and `CopyButtonProps`

src/components/TxHistory/styles/TxDetails.style.js
```diff
@@ -58,4 +58,9 @@ export default StyleSheet.create({
     borderTopWidth: 1,
     borderColor: 'rgba(173, 174, 182, 0.3)',
   },
+  dataContainer: {
+    flexDirection: 'row',
+    paddingRight: 70,
+    marginBottom: 20,
```
wolverineks, line 64: avoid margin like the plague
https://mxstbr.com/thoughts/margin/

src/components/TxHistory/TxDetails.js
```diff
@@ -278,8 +283,9 @@ const TxDetails = ({intl, route}: Props & RouterProps) => {
         <View style={styles.content}>
           <Label>{intl.formatMessage(messages.fromAddresses)}</Label>
           {fromFiltered.map((item, i) => (
-            <>
-              <AddressEntry key={i} {...item} showModalForAddress={showModalForAddress} />
+            // TODO: replace the idx
```
wolverineks: yes

src/components/TxHistory/TxDetails.js
```diff
@@ -289,16 +295,17 @@ const TxDetails = ({intl, route}: Props & RouterProps) => {
...
           ))}
           {cntOmittedFrom > 0 && <Text>{intl.formatMessage(messages.omittedCount, {cnt: cntOmittedFrom})}</Text>}
 
           <View style={styles.borderTop}>
             <Label>{intl.formatMessage(messages.toAddresses)}</Label>
           </View>
           {toFiltered.map((item, i) => (
-            <>
-              <AddressEntry key={i} {...item} showModalForAddress={showModalForAddress} />
+            // TODO: replace the idx
```
wolverineks: yup

src/components/Receive/ReceiveScreenNavigator.js
```diff
@@ -38,7 +38,6 @@ const ReceiveScreenNavigator = injectIntl(({intl}: {intl: IntlShape}) => (
       name={RECEIVE_ROUTES.MAIN}
       component={ReceiveScreen}
       options={({navigation}) => ({
-        // $FlowFixMe it says optional chain is not required but it is
```
wolverineks, line 41: thank you

src/config/networks.js
```diff
@@ -27,145 +27,209 @@ const _DEFAULT_BACKEND_RULES = {
...
+
+type NetworkConfig =
+  | typeof NETWORKS.BYRON_MAINNET
+  | typeof NETWORKS.HASKELL_SHELLEY
+  | typeof NETWORKS.HASKELL_SHELLEY_TESTNET
+  | typeof NETWORKS.JORMUNGANDR
+
+// PROVIDER
+// NOTE: For now we are using the same obj, later we will decouple network/protocol/connection
+export const PROVIDERS = {
```
wolverineks, line 207: provider has a specific meaning in react
is there an alternative name you can use?

src/utils/useCopy.js
```diff
@@ -0,0 +1,23 @@
...
+
+const MESSAGE_TIMEOUT = 1500
+
+export const useCopy = () => {
+  const [copying, setCopying] = useState<string>('')
+
+  useEffect(() => {
+    const isCopying = !!copying
+    if (isCopying) {
+      const timeout = setTimeout(() => {
```
wolverineks: this waits to copy to the clipboard
should it copy immediately? and only use the delay for the message

src/utils/useCopy.js
```diff
@@ -0,0 +1,23 @@
...
+        clearTimeout(timeout)
+        Clipboard.setString(copying)
+        setCopying('')
+      }, MESSAGE_TIMEOUT)
+    }
```
wolverineks, line 19: this should return a clearTimeout callback

wolverineks (changes requested):

src/components/UiKit/Copy.js
stackchain: ctrl+c +v fixed

stackchain: https://medium.com/@abdullahdanyal/how-to-improve-performance-issues-caused-by-arrow-functions-in-react-11dd25527ddf

src/components/UiKit/Copy.js
stackchain: When suggesting a rename please add the code suggestion directly.

src/config/networks.js
stackchain: I suggest sticking to it, is a well-known term in the crypto all blockchain flavors that derive from eth use it, here it has not the same meaning yet, but as mentioned we will be decoupling the network config, protocol config, transport config of the same object asap.

src/utils/useCopy.js
stackchain: Nice catch in fact we don't need to wait to copy, just to re-render the ui.

src/components/UiKit/Copy.js
wolverineks: he is only partly correct.
when a parent renders, unless the child is `class Child extends React.PureComponent` or `React.memo(Child)`,
**all the children render, all the time**. 
`TouchableOpacity` is not a memoized component, so `useCallback` is actually _slower_

src/config/networks.js
```diff
@@ -27,145 +27,209 @@ const _DEFAULT_BACKEND_RULES = {
...
+
+export type YoroiProviderConfig =
+  | typeof BYRON_MAINNET
+  | typeof HASKELL_SHELLEY
+  | typeof HASKELL_SHELLEY_TESTNET
+  | typeof ALONZO_MAINNET
+  | typeof ALONZO_TESTNET
+  | typeof JORMUNGANDR
+
+export const getYoroiProvider = (networkConfig: NetworkConfig, provider: ?YoroiProvider): YoroiProviderConfig => {
```
wolverineks, line 224: why is this called YoroiProvider and not just Provider? or maybe CardanoProvider?

stackchain: In a long term, we will be building yoroi for many targets i'd like to start identifying the yoroi interfaces by prefixing them, therefore when we have many implementations for ex. to Address we know it will have to implement YoroiAddress.

PR #1490, opened 2021-07-25
update ChangePasswordScreen

wolverineks:
* class -> function
* remove hocs
* TextInput


src/components/Settings/ChangePasswordScreen.js
```diff
@@ -30,6 +27,10 @@ const messages = defineMessages({
     id: 'components.settings.changepasswordscreen.newPasswordInputLabel',
     defaultMessage: '!!!New password',
   },
+  passwordStrengthRequirement: {
+    id: 'components.walletinit.createwallet.createwalletscreen.passwordLengthRequirement',
+    defaultMessage: '!!!Minimum characters',
```
zuzunker: @wolverineks , I think the passed value `requiredPasswordLength` should be used somewhere in the default message?

Also, the PR does not contain any change in the en-US file to replace that default value.

zuzunker (changes requested):

src/components/Settings/ChangePasswordScreen.js
wolverineks: change the default message to something like this?
```ts
defaultMessage: '!!!Minimum {requirePasswordLength} characters',
```

the PR didnt add any string, the string already existed

zuzunker: Ahh, I see now, the message exists in a different page so you just copied the descriptor. 🤔 

In the future we might think about extracting commonly used descriptors into a separate module, so we avoid duplication like that. In other projects we have it something like: `globalMessages.requiredPasswordLength` so then it can be just imported and reused.

But ok for now, approving 👌

zuzunker (approved):

stackchain (approved):

PR #1491, opened 2021-07-25
update RestoreWallet

wolverineks:
* remove hocs
* TextInput

### DIFF
--- a/src/components/WalletInit/RestoreWallet/RestoreWalletScreen.js
+++ b/src/components/WalletInit/RestoreWallet/RestoreWalletScreen.js
@@ -1,15 +1,13 @@
 // @flow
 
 import React from 'react'
-import {View, ScrollView} from 'react-native'
-import {compose} from 'redux'
-import {connect} from 'react-redux'
-import {withHandlers, withStateHandlers} from 'recompose'
+import {View, ScrollView, Platform} from 'react-native'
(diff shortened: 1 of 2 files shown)


src/components/WalletInit/RestoreWallet/RestoreWalletScreen.js
```diff
@@ -86,93 +83,67 @@ const errorsVisibleWhileWriting = (errors) => {
...
-            style={styles.phrase}
-            value={phrase}
-            onChangeText={setPhrase}
-            placeholder={intl.formatMessage(messages.mnemonicInputLabel)}
-            blurOnSubmit
-            error={errorText}
-            autoCapitalize="none"
-            keyboardType="visible-password"
-            // hopefully this prevents keyboard from learning the mnemonic
-            autoCorrect={false}
```
stackchain, line 131: I'd suggest to don't remove the autoCorrect.

wolverineks: `autoCorrect={false}` is the default for `TextInput`

stackchain: That's why I like to prefix our local stuff with some identifier. TextInput is the same name from rn components.

stackchain (approved):

PR #1501, opened 2021-07-29
remove-wallet

wolverineks:
* move description into scroll view


src/components/Settings/styles/RemoveWalletScreen.style.js
```diff
@@ -12,17 +12,14 @@ const styles = StyleSheet.create({
...
   description: {
     fontSize: 16,
     lineHeight: 24,
     textAlign: 'center',
   },
 
-  walletInfo: {
-    flex: 1,
-  },
+  walletInfo: {},
```
stackchain: Perhaps we should drop it?

stackchain (changes requested):

stackchain (approved):

PR #1504, opened 2021-07-30
[chore] Circle ci update

stackchain:
- Move from old legacy circleci repo to cimg 
- Bump node img version to 16.5.0

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,6 @@ jobs:
 
       # Because npm link will write in this path
       - run: sudo chown -R "$(whoami):$(whoami)" /usr/local/lib/node_modules
-      - run: npm link rust-cardano-crypto
       - run: yarn lint
 
   android:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,11 +23,6 @@ jobs:
       - run: rustup install 1.41.0
       - run: rustup target add aarch64-linux-android armv7-linux-androideabi i686-linux-android x86_64-linux-android
       - run: cargo install cargo-lipo
-
-      - run:
-          name: Match bitrise's yarn version
-          command: sudo npm install -g yarn@${BITRISE_YARN_VER}
-
       - run:
           name: Install Dependencies
           command: yarn install --network-concurrency 1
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -93,7 +93,7 @@
     "react-native": "0.63.4",
     "react-native-background-timer": "2.4.1",
     "react-native-ble-plx": "2.0.2",
-    "react-native-blockies-svg": "Emurgo/react-native-blockies-svg",
+    "react-native-blockies-svg": "git+https://github.com/Emurgo/react-native-blockies-svg",
```
wolverineks, line 96: what does this do?

.circleci/config.yml
```diff
@@ -37,43 +32,33 @@ jobs:
...
   test:
     <<: *rn_defaults
     steps:
       - attach_workspace:
           at: ~/project
-      - run:
-          name: Match bitrise's yarn version
-          command: sudo npm install -g yarn@${BITRISE_YARN_VER}
-      - run: yarn test
+      - run: yarn test --runInBand
```
wolverineks, line 52: why run serially, instead of parallel?

wolverineks:
does circleci add anything beyond GitHub actions? could we ditch this altogether? 

if Android job isn't running, I'd like to get rid of the job

if Android job isn't running, do we still need to install rust?

package.json
stackchain: When adding modules that use git protocol over yarn/npm adding this way makes it compatible with both.

.circleci/config.yml
stackchain: We are using a free tier in circle-ci, it was the only way to make it run, we were getting too often out-of-memory when running in parallel.

package.json
wolverineks: why is this necessary?

stackchain: The project must be agnostic about the package tool, before it was tight to yarn.

stackchain:
> does circleci add anything beyond GitHub actions? could we ditch this altogether?
> 
> if Android job isn't running, I'd like to get rid of the job
> 
> if Android job isn't running, do we still need to install rust?

@vsubhuman has a plan to remove the ci pipeline from dev we will keep it only for releases probably, but that is not covered in this PR.
There is no longer the android job in ci.

PR #1569, opened 2021-08-25
[fix] Loosen promise after hoc removal

stackchain:


wolverineks (commented):
what does this change do?

stackchain:
> what does this change do?

initApp is async.

wolverineks:
@stackchain right, I might be misunderstanding how useEffect/promises work, but it doesn't look like this changes anything. wrapping a promise in another promise? did you notice a change in behavior?

stackchain:
> @stackchain right, I might be misunderstanding how useEffect/promises work, but it doesn't look like this changes anything. wrapping a promise in another promise? did you notice a change in behavior?

React at build time knows that useEffect is sync and keeps throwing warnings, this PR is to remove react's warnings.

PR #1575, opened 2021-08-27
move modals to AddressesList

wolverineks:


src/components/Receive/AddressView.js
```diff
@@ -1,338 +1,98 @@
...
-}: Props) => {
+const Row = (props: ViewProps) => <View {...props} style={styles.container} />
+const Address = (props) => <View {...props} style={styles.addressContainer} />
+const Actions = (props) => <View {...props} style={styles.actionContainer} />
+const CopyButton = ({text}: {text: string}) => {
+  const MESSAGE_TIMEOUT = 1000
   const [isCopying, setIsCopying] = useState<boolean>(false)
 
-  const address = addressInfo.address
   useEffect(() => {
```
stackchain: Perhaps replace it with [this](https://github.com/Emurgo/yoroi-mobile/blob/develop/src/utils/useCopy.js)

src/components/Receive/AddressView.js
```diff
@@ -1,338 +1,98 @@
...
-import type {HWDeviceInfo, DeviceId, DeviceObj} from '../../crypto/shelley/ledgerUtils'
-import type {WalletMeta} from '../../state'
-
-const messages = defineMessages({
-  copyLabel: {
-    id: 'components.receive.addressmodal.copyLabel',
-    defaultMessage: '!!!Copy',
+import {type ViewProps} from 'react-native/Libraries/Components/View/ViewPropTypes'
+
+const styles = StyleSheet.create({
```
stackchain: In most places we have the styles apart from the components, do you think we should move everything inside?

stackchain (changes requested):

PR #1618, opened 2021-09-15
Revert "upgrade react intl"

stackchain:
Reverts Emurgo/yoroi-mobile#1617


stackchain:
It looks like some polyfill is missing/not working properly

PR #1632, opened 2021-09-28
Added receiver address validation x network

stackchain:
- This PR adds the validation for the address field on SendScreen (the receiver)
- If the open wallet is Shelley, it will check the chain network id of the wallet inputted versus the current wallet, and will not allow sending if doesn't match.


src/utils/validators.js
```diff
@@ -120,13 +121,32 @@ export const getWalletNameError = (
...
+    try {
+      const networkConfig = getNetworkConfigById(walletNetworkId)
+      const configNetworkId = networkConfig.CHAIN_NETWORK_ID && Number(networkConfig.CHAIN_NETWORK_ID)
+      const addressNetworkId = await address.network_id()
+      if (addressNetworkId !== configNetworkId && !isNaN(configNetworkId)) {
+        return {invalidAddress: true}
+      }
+      // eslint-disable-next-line no-empty
+    } catch (e) {
+      // NOTE: old unsupported addresses will be deprecated soon
```
wolverineks: log unexpected errors

wolverineks: if invalid is indicated by thrown error, return `{invalidAddress: false}`

wolverineks (changes requested):

PR #1675, opened 2021-10-11
Fixed duplicate words handling in wallet confirmaiton

zuzunker:
Mnemonic phrases may contain any number of duplicate words, it's valid. Right now duplicates are not handled correctly - all duplicates are removed in case a single similar word is selected. Fixed that problem and also tried to implement the optimal duplicates handling so that we will correctly hide the exact word the user tapped, no matter its position among the duplicates.

```
/*
   * The mnemonic are handled in "word entries" instead of plain text word
   * Where each entry is [word, wordIndex] with index in sorted array
   * This is done so that each word including any duplicates
   * is uniquely identified by its index in the sorted array
   * to improve the UX of the word selecting.
   *
   * Example: original words might be [air, sand, air, desk], which is valid.
   * Sorted entries will be: [[air, 0], [air, 1], [desk, 2], [sand, 3]]
   *
   * We don't care which of the "air" words the user will want to use first.
   * If user clicks on [air, 1] we will be able to detect that this specific
   * was selected and hide it from the options while keeping the first "air"
   * visible, which should be the most intuitive and expected behaviour for the users.
   *
   * When comparing with the original mnemonic we ignore the indexes.
   */
   ```

### DIFF
--- a/src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
+++ b/src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
@@ -42,13 +42,38 @@ const MnemonicCheckScreen = () => {
   const navigation = useNavigation()
   const route = (useRoute(): any)
   const mnemonic: string = route.params.mnemonic
-  const sortedWords = mnemonic.split(' ').sort()
-  const [partialPhrase, setPartialPhrase] = React.useState<Array<string>>([])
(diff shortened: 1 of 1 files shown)


zuzunker (commented):
/check

zuzunker (commented):
/check

zuzunker:
Closing as a duplicate of better https://github.com/Emurgo/yoroi-mobile/pull/1677

PR #1677, opened 2021-10-12
Mnemonic-duplicate-words

### LINKED ISSUE #1566: Duplicate word in recovery phrase prevents confrimation.
When creating a new account, the user is required to confirm the recovery phrase by putting the words in the correct order. 

If the recovery phrase has a repeated word, then this becomes impossible since the GUI removes all instances of the repeated word when placing the first instance of that repeated word.

Afterwards, the user cannot place other instances of that repeated word and will not be able to confirm the recovery phrase.

Using android app, version 4.3.0

### LINKED ISSUE #1671: Recovery phrase is impossible to verify if you get duplicate words
Was helping my dad setup a Yoroi wallet and he had "oval" twice in his recovery phrase.

When he went to verify the recovery phrase, clicking "oval" removed both instances from the word list making it impossible to complete the verification step.

wolverineks:
Mnemonic phrases may contain any number of duplicate words, it's valid. Right now duplicates are not handled correctly - all duplicates are removed in case a single similar word is selected. Fixed that problem and also tried to implement the optimal duplicates handling so that we will correctly hide the exact word the user tapped, no matter its position among the duplicates.


src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
```diff
@@ -114,24 +102,21 @@ const MnemonicCheckScreen = () => {
...
       <View style={[styles.recoveryPhraseOutline, error && styles.recoveryPhraseError]}>
-        {partialPhrase.map((word, index, array) => {
+        {userEntries.map((entry, index, array) => {
           const isLast = index === array.length - 1
 
           return (
-            <View key={word} style={[styles.wordBadgeContainer, !isLast && styles.selected]}>
-              <WordBadge word={word} disabled={!isLast} onPress={isLast ? () => onPress(word) : undefined} />
+            <View key={entry.id} style={[styles.wordBadgeContainer, !isLast && styles.selected]}>
+              <WordBadge word={`${entry.word} x`} disabled={!isLast} onPress={() => onPress(entry)} />
```
zuzunker: It's a good call that you replaced removing arbitrary entry with `removeLastEntry` that takes no argument, but then I am confused why here you added `x` to each displayed word? 🤔 

I think `onPress` here also can be changed in type to take no arguments, since it's private local component so it doesn't have to be generic.

zuzunker (commented):
/check

src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
zuzunker: Ah also I just noticed that previous function was only processing tap on the last word and ignoring taps on any other words, so functionally it also only removed last word, but with your change user can tap anywhere in the area and it will remove the last word, which I like 👍 

But still I think we don't need the `x` added there, or at least maybe only to the last word

src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
```diff
@@ -1,64 +1,52 @@
...
 const MnemonicCheckScreen = () => {
   const intl = useIntl()
   const navigation = useNavigation()
-  const route = (useRoute(): any)
-  const mnemonic: string = route.params.mnemonic
-  const sortedWords = mnemonic.split(' ').sort()
-  const [partialPhrase, setPartialPhrase] = React.useState<Array<string>>([])
-  const deselectWord = (removeWord: string) => setPartialPhrase(partialPhrase.filter((word) => word !== removeWord))
-  const selectWord = (addWord: string) => setPartialPhrase([...partialPhrase, addWord])
+  const {mnemonic, password, name, networkId, walletImplementationId, provider} = useParams<Params>()
```
zuzunker, line 33: `useParams` will take same parameters from `route.params`, right? So no need to change how params are passed into the component, only internal change?

wolverineks: yeah, just a wrapper that gets you types
longer term we can abandon storing state in navigation, and migrate to a context

src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
wolverineks: I moved the x from word badge, so only the mnemonic input displays the x  and not the word badge list

wolverineks: it might make more sense to have a single "backspace" button, instead of an "x" on each badge, then it's easier to remove multiple entries

zuzunker: > it might make more sense to have a single "backspace" button, instead of an "x" on each badge, then it's easier to remove multiple entries

@wolverineks , agree

zuzunker: I've noticed the removed `x` in the word badge now. Works for now then, but can be improved with a backspace button in the future, yes

zuzunker (approved):

src/components/WalletInit/CreateWallet/MnemonicCheckScreen.js
```diff
@@ -42,13 +42,38 @@ const MnemonicCheckScreen = () => {
...
+    .split(' ')
+    .sort()
+    .map((s, i) => [s, i]);
+
+  const [partialPhraseEntries, setPartialPhraseEntries] =
+    React.useState<Array<[string, number]>>([])
+  const selectWord = (addWordEntry: [string, number]) =>
+    setPartialPhraseEntries([...partialPhraseEntries, addWordEntry])
+  const deselectWord = ([, removeWordIdx]: [string, number]) =>
+    setPartialPhraseEntries(partialPhraseEntries.filter(([, idx]) => idx !== removeWordIdx))
```
stackchain: 👍

stackchain:
https://user-images.githubusercontent.com/30806844/137045923-25bd6781-cecb-4e00-b9b4-f2bd0569247f.mp4

Added one script that creates mnemonics with duplicated words, it is going to be a reference for future testing.

LGTM

stackchain (approved):

PR #1700, opened 2021-10-25
Hotfix/mnemonic 4.4.2

### LINKED ISSUE #1705: UI Issue: 3 letter mnemonics (especially on iOS)
3 iOS users have reported an issue while restoring wallets using mnemonics on v4.4.1, for 3-letter words like `ski` - where one of the two behaviour is seen:
1. Typing `ski` shows the word correctly in pulldown for hint, but clicking on `ski` makes the cell "empty" without a way to manually type ski and skip selection. Users experienced this on iOS, but I was randomly able to replicate it *only* once on Android, but retrying since resulted in second UI behaviour below.
2. Typing 3 word letter does not show a pull down , but I am able to click on next word cell and proceed through. User reporting was able to do this, but at the end - was not able to "activate" "Restore Wallet" until hitting return key on iOS. On android you're still able to click through on next word or not have trouble clicking on Restore Wallet.

Overall the UX with 3 letter words for MnemonicsInput is a bit inconsistent

stackchain:


zuzunker:
Release: https://github.com/Emurgo/yoroi-mobile/releases/tag/4.4.2

PR #1702, opened 2021-10-27
Mock-wallet-data

wolverineks:


legacy/crypto/walletManager.js
```diff
@@ -154,6 +154,14 @@ class WalletManager {
     return this._wallets
   }
 
+  getWallet() {
```
stackchain, line 157:
```suggestion
  get wallet() {
```

legacy/crypto/walletManager.js
```diff
@@ -222,7 +230,7 @@ class WalletManager {
    */
 
   get isInitialized() {
-    return this._wallet
+    return this.getWallet().isInitialized
```
stackchain, line 233:
```suggestion
    return this.wallet.isInitialized
```

legacy/crypto/walletManager.js
stackchain: my concern is only about the `throw` here, throwing inside getters would not be a good practice

stackchain:
I think we should consider implementing the state pattern for the wallet.

legacy/crypto/walletManager.js
wolverineks: why make it a getter?

wolverineks:
> I think we should consider implementing the state pattern for the wallet.

I'm not sure what you mean. could you elaborate?

stackchain (approved):

PR #1766, opened 2021-11-10
fix/Tx details error

### LINKED ISSUE #1633: Clicking on transaction id doesn't give  the right result.
Under 'transactions' when you are in a transaction.

The assumption is when you click on the transaction ID button (green) you would go to the Cardano explorer and view it

It opens up to:
https://explorer.cardano.org/de/transaction.html?id=

...Without the id

And in German language setting...

(Yes the transaction was created by a German individual, but I am not🤭)

stackchain:
- Fix the error when opening the tx-details modal
- Fix the error when tapping the open in browser button (closes #1633)

https://app.asana.com/0/1201135371863834/1201328081609151/f

### DIFF
--- a/legacy/components/TxHistory/TxDetails.js
+++ b/legacy/components/TxHistory/TxDetails.js
@@ -313,8 +313,14 @@ const TxDetails = () => {
           </View>
         </View>
       </Screen>
-      {/* $FlowFixMe TODO: index does not exist in AddressModal props */}
-      <AddressModal visible={!!addressDetail} onRequestClose={hideAddressModal} address={addressDetail} index={null} />
+      {addressDetail && (
+        <AddressModal
+          visible
+          onRequestClose={hideAddressModal}
+          address={addressDetail}
+          onAddressVerify={hideAddressModal}
+        />
+      )}
     </View>
   )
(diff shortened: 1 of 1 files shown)


legacy/components/TxHistory/TxDetails.js
```diff
@@ -313,8 +313,14 @@ const TxDetails = () => {
...
         </View>
       </Screen>
-      {/* $FlowFixMe TODO: index does not exist in AddressModal props */}
-      <AddressModal visible={!!addressDetail} onRequestClose={hideAddressModal} address={addressDetail} index={null} />
+      {addressDetail && (
+        <AddressModal
+          visible
+          onRequestClose={hideAddressModal}
+          address={addressDetail}
+          onAddressVerify={hideAddressModal}
```
stackchain, line 321: note: it is reserved for HW wallets, but is not implemented, we will be changing it on the next refactor, keeping it like this for now.

Nebyt (approved):
Checked locally on an emulator. Works fine, no errors. Redirecting to the Cardano explorer works fine.

wolverineks (approved):

PR #1769, opened 2021-11-13
Change-password

wolverineks:


legacy/state.js
```diff
@@ -110,7 +110,21 @@ export type State = {
 }
 
 export const getInitialState = (): State => ({
-  wallets: {},
+  wallets: {
```
stackchain: I think it's possible this is part of mock stuff.

src/Settings/SettingsScreenNavigator.tsx
```diff
@@ -141,22 +50,22 @@ const SettingsScreenNavigator = () => {
...
+        options={{title: strings.changeWalletNameTitle}}
       />
       <Stack.Screen
         name={SETTINGS_ROUTES.TERMS_OF_USE}
         component={TermsOfServiceScreen}
-        options={{title: intl.formatMessage(messages.termsOfServiceTitle)}}
+        options={{title: strings.termsOfServiceTitle}}
       />
-      <Stack.Screen
+      <Stack.Screen //
```
stackchain: I think this is a comment that didn't survive.

stackchain (changes requested):

src/Settings/SettingsScreenNavigator.tsx
wolverineks: it's only there to prevent formatting to a single line. I'll change it to use `// prettier-ignore` to be clearer

stackchain (approved):

PR #1775, opened 2021-11-16
Tx item improvements

stackchain:
- Added the `fee` for `sent` and `intrawallet` transactions
- Renamed the transaction status from `high` -> `success`  to match the new layout


wolverineks (approved):

stackchain (commented):
/check

PR #1776, opened 2021-11-16
Fix withdraw/open url button

stackchain:
- Withdraw is always enable from now on (is the only way to deregister from a pool)
- Removed the open url button for testnet (since we don't have pool site for it)


legacy/components/Delegation/dashboard/DelegatedStakepoolInfo.js
```diff
@@ -124,14 +124,16 @@ const DelegatedStakepoolInfo = ({intl, poolTicker, poolName, poolHash, poolURL}:
...
 
-        <View style={styles.bottomBlock}>
-          <Button
-            outlineOnLight
-            shelleyTheme
-            onPress={openExternalURL}
-            title={intl.formatMessage(messages.fullDescriptionButtonLabel)}
-          />
-        </View>
+        {poolURL && (
```
wolverineks, line 127: the types say that `poolURL` is a string
why check if it exists?

wolverineks: oh, `StakingDashboard` is a mess, thats why

wolverineks (approved):

stackchain (commented):
/check

PR #1782, opened 2021-11-17
Bugfix multi taps on catalyst app download

stackchain:
- Scenario: while tapping many times on any of the catalyst app download buttons, the app crashes, wrapped them into an async function, and swallowing errors.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201363618983586

### DIFF
--- a/legacy/components/Catalyst/Step1.js
+++ b/legacy/components/Catalyst/Step1.js
@@ -15,6 +15,7 @@ import AppDownload from '../../assets/img/pic-catalyst-step1.png'
 import globalMessages, {confirmationMessages} from '../../i18n/global-messages'
 import {CATALYST_ROUTES} from '../../RoutesList'
 import {isDelegatingSelector} from '../../selectors'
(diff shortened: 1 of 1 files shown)


wolverineks:
what's the error? and hope does this fix it?

stackchain:
> 
> 
> what's the error? and hope does this fix it?

Following the video, when the user is on the store and heads back to the app this error is popping up, I think the way the app handles errors right now (the general catch-all) is having some troubles with the loosen promise.

https://user-images.githubusercontent.com/30806844/142302101-60dd81e8-2695-42f4-b14b-fc538f4ba4f7.MOV

Nebyt (approved):
Checked on the Android 11. Samsung Galaxy A52. Works fine. There is no delay between pressing the store button and opening the app page in the store now. 🚀

wolverineks (approved):

PR #1804, opened 2021-11-28
TxHistoryListItem migrating to tx

stackchain:


src/components/Icon/Icon.stories.tsx
```diff
@@ -2,68 +2,85 @@ import {storiesOf} from '@storybook/react-native'
...
+import {TransactionAssurance, TransactionDirection, TransactionStatus} from '../../types/cardano'
 import {Icon, Spacer} from '..'
 
 storiesOf('Icon', module).add('Gallery', () => (
   <ScrollView contentContainerStyle={{padding: 16}}>
     <Row>
-      <Item icon={<Icon.Export size={40} />} title={'Export'} />
-      <Item icon={<Icon.Magnify size={40} />} title={'Magnify'} />
-      <Item icon={<Icon.Received />} title={'Received'} />
+      <Item icon={<Icon.Export size={40} />} title="Export" />
```
wolverineks: if you are concerned about curly braces around strings
[react/jsx-curly-brace-presence](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-brace-presence.md)

src/components/Icon/Icon.stories.tsx
```diff
@@ -2,68 +2,85 @@ import {storiesOf} from '@storybook/react-native'
...
+      <Item icon={<Icon.ChevronDown />} title="ChevronDown" />
+      <Item icon={<Icon.ChevronUp />} title="ChevronUp" />
+      <Item icon={<Icon.Verify />} title="Verify" />
     </Row>
 
     <Spacer height={16} />
 
     <Row>
-      <Item icon={<Icon.NoAssetImage />} title={'No Asset Image'} />
+      <Item icon={<Icon.Direction transaction={transactionFor('SENT', 'SUCCESSFUL')} />} title="Success-Sent" />
```
wolverineks: we should set up storybook mocks
```ts
// storybook/mocks.ts
export const wallet = () => ({ ... })
export const transaction = () => ({ ... })
export const token = () => ({ ... })
```

wolverineks (changes requested):

src/components/Icon/Direction.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
+import {View} from 'react-native'
+
+import {TransactionDirection, TransactionInfo} from '../../types/cardano'
+import {Received} from '../Icon/Received'
+import {Sent} from '../Icon/Sent'
+import {Transaction} from '../Icon/Transaction'
+
+type Props = {
+  transaction: Required<Pick<TransactionInfo, 'assurance' | 'direction' | 'status'>>
```
wolverineks: after creating mocks, this can be changed to `TransactionInfo`

wolverineks (changes requested):

src/types/cardano.ts
```diff
@@ -0,0 +1,169 @@
+import BigNumber from 'bignumber.js'
+
+// Wallet Candidates
+
+export type TransactionDirection = 'SENT' | 'RECEIVED' | 'SELF' | 'MULTI'
+export type TransactionStatus = 'SUCCESSFUL' | 'PENDING' | 'FAILED'
+export type TransactionAssurance = 'PENDING' | 'FAILED' | 'LOW' | 'MEDIUM' | 'HIGH'
+
+export type Block = {
```
wolverineks, line 9: my guess is, all blockchains will have blocks with heights and hashes, but not all will have epochs or slots
so, maybe something like:
```ts
namespace Yoroi {
  type Block = {
    height: number, 
    hash: string
  }
}
```
and 

```ts
namespace Cardano {
  type Block = {
    height: number, 
    hash: string, 
    epoch: number, 
    slot: number
  }
}
```

src/types/cardano.ts
```diff
@@ -0,0 +1,169 @@
...
+  height: number
+  epoch: number
+  slot: number
+  hash: string
+}
+
+export type TransactionInfo = {
```
wolverineks, line 16: similar here
```ts
type TransactionInfo = {...} // cardano specific, combine with RawTransaction?
```

https://emurgo.atlassian.net/wiki/spaces/CARDANO/pages/618430503/Platform+Agnostic#Transaction
```ts
type Transaction = {
  ...all the same stuff except inputs, outputs...
}
```

src/components/Icon/Icon.stories.tsx
```diff
@@ -116,8 +123,9 @@ const transactionFor = (
   direction: TransactionDirection,
```
wolverineks: i think we can put this directly inline with the component

wolverineks (changes requested):

src/components/Icon/Icon.stories.tsx
stackchain: I migrated to the mock and put it inline

legacy/selectors.d.ts
```diff
@@ -27,10 +27,12 @@ export var easyConfirmationSelector: (state: State) => boolean
 export var isHWSelector: (state: State) => boolean
 export var walletNameSelector: (state: State) => string
 export var walletNamesSelector: (state: State) => Array<string>
+export var defaultNetworkAssetSelector: (state: State) => Token
```
wolverineks: this returns `DefaultAsset`

stackchain: Yap, it returns from the `availableAssetsSelector` selector which return type is not precise, since it can return Token | DefaultAsset.

src/types/cardano.ts
stackchain: This will be the next step, for now just keeping the same types that we were using before.

wolverineks (approved):

PR #1815, opened 2021-12-04
fix send screen loop

wolverineks:


wolverineks:
createUnsignedTx is now only called when changing address, amount, token

this was caused by an unstable identity for the selected token object
a new token was repeatedly created with the same values
in the long term, moving all the wallet selectors into a wallet object would prevent this

stackchain (approved):

PR #1826, opened 2021-12-10
Denis/appium tests beginning

Nebyt:
The beginning of automation with Appium.


test/config/wdio.andriod.app.conf.js
```diff
@@ -0,0 +1,28 @@
+const { join } = require('path');
```
wolverineks, line 1: file name typo
andiod -> android

wolverineks (approved):
if its not already setup, can you set up prettier / eslint for these files?
would it be possible/beneficial to write these in typescript?
just fyi, the screen objects can be written as plain javascript objects

-  fix the filename typo

wolverineks:
> if its not already setup, can you set up prettier / eslint for these files?
> would it be possible/beneficial to write these in typescript?
> just fyi, the screen objects can be written as plain javascript objects
> 
> -  fix the filename typo

@Nebyt

PR #1833, opened 2021-12-13
Denis/move js2ts

Nebyt:
Changing JS to TS in appium tests.


wolverineks (commented):
what is driver? its not defined anywhere

Nebyt:
> what is driver? its not defined anywhere

The answer from the [webdriver.io](https://webdriver.io/docs/browserobject/)
> The browser object is the session instance you use to control the browser or mobile device with. If you use the WDIO test runner, you can access the WebDriver instance through the global browser or driver object.
> 
> The session is initialized by the test runner. The same goes for ending the session. This is also done by the test runner process.

image

tsconfig.json
```diff
@@ -11,7 +11,16 @@
...
     "noImplicitAny": false,
     "target": "esnext",
-    "types": ["react", "react-native", "react-redux", "jest"],
+    "types": [
+      "react",
+      "react-native",
+      "react-redux",
+      "jest",
+      "node",
+      "webdriverio/sync",
```
wolverineks, line 20: the docs mention `webdriverio/async`
is there a difference?

wolverineks:
> > what is driver? its not defined anywhere
> 
> The answer from the [webdriver.io](https://webdriver.io/docs/browserobject/)
> > The browser object is the session instance you use to control the browser or mobile device with. If you use the WDIO test runner, you can access the WebDriver instance through the global browser or driver object.
> > 
> > The session is initialized by the test runner. The same goes for ending the session. This is also done by the test runner process.
> 
> image
> 

I'm seeing a ts error with `driver`, are the types installed correctly?

tsconfig.json
Nebyt: Yeah, you are right. `webdriverio/async` should be here.

wolverineks (approved):

stackchain (approved):

PR #1835, opened 2021-12-15
Denis/wallet with existing name

Nebyt:
Fixed the wdio config. Added the test "Already existing name".


wolverineks (changes requested):
the tests should not depend on strings
check that an error exists and that the user is prevented from continuing

this could also be wrapped up in a wallet creation test
test that you can't use a duplicate name, change the name, then finish the wallet creation test

Nebyt:
> the tests should not depend on strings check that an error exists and that the user is prevented from continuing

Hmm... Where do tests depend on strings?

PR #1840, opened 2021-12-17
Fixing tokens displayed when sending an ALL_TOKEN transaction

### LINKED ISSUE #1838: "Send all [asset]" wants to send unrelated assets too
Hi everyone,

Trying to send some assets between wallets, there might be a bug when using the the 'Send all" checkbox.

When checking the "Send all [asset]" for a given transaction 
IMG_DDF1AAF20D09-1
 
The wallet shows a warning, however displays that it wants to send all the assets in the wallet
sending all hosk

The feature works properly with the browser extension.

zuzunker:
Added check for send-all AND default-asset-selected

### DIFF
--- a/legacy/components/Send/SendScreen.js
+++ b/legacy/components/Send/SendScreen.js
@@ -259,7 +259,7 @@ class SendScreenLegacy extends Component<LegacyProps, State> {
 
       const fee = (await transactionData.fee()).getDefault()
 
-      const isDefaultAssetSelected = selectedTokenMeta.isDefault;
+      const isDefaultAssetSelected = selectedTokenMeta.isDefault
       const defaultAssetAmount = isDefaultAssetSelected
         ? parseAmountDecimal(amount, selectedTokenMeta)
         : // note: inside this if balanceAfter shouldn't be null
--- a/legacy/components/Send/SendScreen.js
+++ b/legacy/components/Send/SendScreen.js
@@ -259,16 +259,17 @@ class SendScreenLegacy extends Component<LegacyProps, State> {
 
       const fee = (await transactionData.fee()).getDefault()
 
-      const defaultAssetAmount = selectedTokenMeta.isDefault
+      let isDefaultAssetSelected = selectedTokenMeta.isDefault;
+      const defaultAssetAmount = isDefaultAssetSelected
         ? parseAmountDecimal(amount, selectedTokenMeta)
         : // note: inside this if balanceAfter shouldn't be null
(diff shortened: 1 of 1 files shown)


Nebyt (approved):
Checked. Fixed. :rocket:

wolverineks:
what is this change for?

legacy/components/Send/SendScreen.js
```diff
@@ -259,25 +259,23 @@ class SendScreenLegacy extends Component<LegacyProps, State> {
 
       const fee = (await transactionData.fee()).getDefault()
 
-      const defaultAssetAmount = selectedTokenMeta.isDefault
+      let isDefaultAssetSelected = selectedTokenMeta.isDefault;
```
wolverineks: why is this `let`?

wolverineks:
> what is this change for?

it looks like this is making a list of tokens to display on the confirm screen
we should have a single source of truth
is there any way we can derive this from the transaction data itself?
what else can we derive from the transactionData?

```ts
const tokens: Array<{ tokenId: string, amount: string, address: string }> = getTokens(transactionData)
const fee: string = await getFee(transactionData)
```

Nebyt:
> what is this change for?

It is for this https://github.com/Emurgo/yoroi-mobile/issues/1838 issue

wolverineks (changes requested):
change `let` -> `const`
run the changes through prettier

legacy/components/Send/SendScreen.js
zuzunker: ah, automatically generated, haven't noticed. Good catch!

zuzunker:
> change `let` -> `const` run the changes through prettier

@wolverineks, Updated. Runned it through `yarn prettify`

wolverineks (approved):

wolverineks:
> > change `let` -> `const` run the changes through prettier
> 
> @wolverineks, Updated. Runned it through `yarn prettify`

just realized that `yarn prettify` doesnt format the old code, so you have to manually run prettier on that file

zuzunker:
Weird, I have runned it through the prettify but now it fails with some prettier errors 🤔

zuzunker:
> > > change `let` -> `const` run the changes through prettier
> > 
> > 
> > @wolverineks, Updated. Runned it through `yarn prettify`
> 
> just realized that `yarn prettify` doesnt format the old code, so you have to manually run prettier on that file

ahhhhhh, that explains it

zuzunker (commented):
/check

PR #1842, opened 2021-12-18
Using fingerprint in asset-list and asset-selector

zuzunker:
In the asset-list (in transactions page) and in the asset-selector (in send page) replacing undecoded asset name on the second line with the fingerprint which is more useful for users to see.


zuzunker (commented):
/check

PR #1863, opened 2022-01-12
Denis/restore wallet 24 test

Nebyt:
Reworked the previous test and added a test for restoring a 24-word wallet.


tests/constants.ts
```diff
@@ -9,21 +22,58 @@ export const SPENDING_PASSWORD = '1234567890'
...
+      'artwork',
+      'material',
+      'three',
+      'thumb',
+      'melody',
+      'zoo',
+      'fence',
+      'east',
+    ],
+    type: 1,
```
wolverineks: just a nitpick here, but I try to avoid vague values like this. 
something like this reads better
`type: 15-word-mnemonic` 
`if (wallet.type === '15-word-mnemonic') ...`
you could use an `enum` to keep the readability while also keeping the number
```
enum Wallet type {
    SomeType;
    SomeOtherType
}

if (wallet.type === WalletType.SomeType) ...
```
or maybe something else like
`if (wallet.mnemonicLength === 15) ...`
but like I said, just a nitpick

wolverineks (approved):

tests/constants.ts
Nebyt: Good point, I will change it.

stackchain (approved):

PR #1864, opened 2022-01-12
Changed the biometrics behaviour at startup

### LINKED ISSUE #627: (Android/iOS) Touch ID feature doesn't work when the user tries to enable "Sign in with your biometric" option
1. Disable "Sign in with your biometric" and enter the PIN
2. Enable "Sign in with your biometric" and put your finger on the home button
Please, see the video
https://emurgo.slack.com/archives/GC9AM0E8K/p1549371247026900

### LINKED ISSUE #643: (iOS) Sometimes it is possible to disable "Sign in with your biometrics" without turning off the "Easy Transaction Confirmation" feature.
Which leads to the crash of the app if the user tries to send a transaction:
img_0125

Steps to reproduce this are currently unknown. I will add them later (if I can).

### LINKED ISSUE #1467: “Sign in with biometrics” asks about access to fingerprints even though my iPhone uses Face ID and has no fingerprint sensor.

### LINKED ISSUE #1841: Authorize with your biometrics
Hey 
I can't log in on my Yoroi wallet on iphone 6s..

/the screen includes Authorize with your biometrics and use other login method which i can't click them too 

also i don't find my recovery words and that's really important to me log in with my phone .. Please any way to solve that ?

stackchain:
- Right now if the app is launched when the biometrics was set in the app but disabled or no longer supported by the device, the user is locked out of the app.
- To prevent this situation which no fallback is possible, when launching the app with biometrics disabled on phone but set in the app, at first launch it will require to set the PIN that will be used in place of biometrics.


Nebyt (approved):
Checked. Works fine. :+1:

zuzunker (approved):

PR #1879, opened 2022-01-18
Resync return to main nav

stackchain:
- In react-navigation 6.x we have `popToTop` function but since we can't bump it yet, resetting the navigation after wallet resync.


wolverineks (approved):

Nebyt (approved):
:+1:

PR #1895, opened 2022-01-26
Retrying if bio was disabled wallet selection

stackchain:
- This behavior will change in version 5.0.0


Nebyt (approved):
Fixed :superhero:

PR #1914, opened 2022-02-22
Fix-restore-wallet-crash

wolverineks:


Nebyt (approved):
Fixed restoring a read-only wallet and removing a wallet. :rocket:

PR #1919, opened 2022-02-28
invalidate queries from mutations

wolverineks:


Nebyt (approved):
Checked creating, restoring a wallet and restoring a read-only wallet.

stackchain (approved):

PR #1932, opened 2022-03-18
Pending tx store

stackchain:


src/store/pendingTx.ts
```diff
@@ -0,0 +1,56 @@
+import {PendingTransaction, WalletInterface} from '../types'
+import {PendingTxStore, StoreKey} from './store'
+
+export class PendingTxService {
```
wolverineks, line 4: this should receive a generic store instead of wallet

wolverineks, line 4: avoid classes, convert to function with closure

src/store/pendingTx.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (!store) return
+
+    store.data = store.data.filter((r) => r.id !== id)
+    return this._wallet.store.write(PendingTxService.storeKey, store)
+  }
+
+  async save(record: Readonly<PendingTransaction>) {
+    const store = await this._readStore()
+    if (store) {
+      const index = store.data.findIndex((r) => r.id === record.id)
```
wolverineks, line 42: is pending tx store typed?

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
...
+  key: StoreKey
+} & (
+  | {
+      value: StoreData<StoreKey>
+      method: StoreMethod.Write
+    }
+  | {method: StoreMethod.Remove}
+)
+
+export class StoreService {
```
wolverineks, line 34: should pending txs be stored separate from confirmed txs?
could there be a generic tx store and pending is a status?

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
+import AsyncStorage, {AsyncStorageStatic} from '@react-native-async-storage/async-storage'
+import {Subject} from 'rxjs'
```
wolverineks, line 2: what does this get us that we can do ourselves?

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import {PendingTransactions} from '../types/transactions'
+
+export enum StoreKey {
```
wolverineks, line 6: why use an enum if there is only 1?

src/store/store.ts
wolverineks, line 34: just call this Store

src/store/store.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      method: StoreMethod.Write,
+      value: mockedVersionablePendingTx,
+    }
+    const remove2ndEvent: StoreEvent = {
+      key: StoreKey.PendingTx,
+      method: StoreMethod.Remove,
+    }
+
+    // act
+    const subscriber1 = storage.publisher$.subscribe(subscriber1fn)
```
wolverineks, line 52: leaky abstraction, put subscribe directly on the store

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
...
+  PendingTx = 'pendingTx',
+}
+
+export type VersionableData<T> = {
```
wolverineks, line 10: why does this need to be versionable?

src/store/store.ts
wolverineks: why does the store need to know about any keys?

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
...
+  Write,
+  Remove,
+}
+
+export type StoreEvent = {
+  key: StoreKey
+} & (
+  | {
+      value: StoreData<StoreKey>
+      method: StoreMethod.Write
```
wolverineks, line 29: why do we need to emit the type of event?

src/store/store.ts
wolverineks, line 34: avoid classes, convert to function with closure

storybook/mocks/wallet.ts
```diff
@@ -26,6 +26,7 @@ export const mockWallet: WalletInterface = {
   isUsedAddressIndex: {},
   numReceiveAddresses: 0,
   isInitialized: true,
+  store: new StoreService('/mocked_wallet_v2/wallet-id/'),
```
wolverineks, line 29: use the mocked wallets id

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
...
+  private _storageKey(key: string) {
+    return `${this._prefix}/${key}`
+  }
+
+  public async write(key: StoreKey, value: StoreData<typeof key>): Promise<void> {
+    const storageKey = this._storageKey(key)
+
+    await this._storage.setItem(storageKey, JSON.stringify(value))
+
+    this._emitter$.next({
```
wolverineks, line 58: how is this better than ?
```
subscribers.forEach(subscriber => subscriber(data))
```

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    await this._storage.removeItem(storageKey)
+
+    this._emitter$.next({
+      key,
+      method: StoreMethod.Remove,
+    })
+  }
+
+  public destroy() {
```
wolverineks, line 86: throwing errors due to incorrect code is better than ignoring the incorrect code

src/store/store.ts
```diff
@@ -0,0 +1,91 @@
...
+export class StoreService {
+  private readonly _storage: AsyncStorageStatic
+  private readonly _prefix: string
+  private readonly _emitter$: Subject<StoreEvent>
+
+  get publisher$() {
+    return this._emitter$
+  }
+
+  constructor(prefix: string) {
```
wolverineks, line 43: instead of passing the prefix in, this should be initialized with an implementation with the prefix already included
```
const wallet = await createWallet()
new Storage({
  read: (path: string) => RNAS.getItem(`${wallet.id}/${path}`),
  write: (path: string, data: unknown) => RNAS.setItem(`${wallet.id}/${path}`, data)
  ...
})
```

src/store/pendingTx.ts
wolverineks: might not need to be a class at all

wolverineks, line 4: doesnt look like this is used anywhere?

legacy/crypto/walletManager.js
```diff
@@ -509,6 +511,7 @@ class WalletManager {
 
     await wallet.restore(data, walletMeta)
     wallet.id = walletMeta.id
+    wallet.store = new StoreService(`/wallet_v2/${walletMeta.id}`)
```
wolverineks, line 514: move this to something like...
`new ShelleyWallet({ storage: Storage })`

stackchain:
Redesigning without classes

PR #1935, opened 2022-03-18
fix restore hw wallet

wolverineks:


Nebyt (approved):
Checked localy.
Checked:
* Restoring a usual wallet
* Restoring a HW wallet
* Creating a wallet
* Removing a wallet
* Renaming
:+1:

PR #1948, opened 2022-04-06
Fix catalyst step6 navigation

stackchain:


wolverineks (changes requested):
wrap the use effect on a custom hook and give it a good name

src/Catalyst/Step6.tsx
```diff
@@ -151,6 +127,35 @@ const CopyButton = ({text}: {text: string}) => (
...
+        }
+      >,
+    ) => {
+      if (e.data.action.type !== 'RESET') {
+        e.preventDefault()
+      }
+    }
+    navigation.addListener('beforeRemove', handleBeforeRemove)
+    return () => {
+      navigation.removeListener('beforeRemove', handleBeforeRemove)
```
wolverineks: i think removeListener is deprecated
`const unsubscribe = navigation.addListener(...)`

wolverineks: the only thing this does is prevent going back, except for "reset"?

wolverineks: also, there is a `listeners` prop for each screen, can we use that instead?
https://reactnavigation.org/docs/navigation-events/#listeners-prop-on-screen

wolverineks (changes requested):

src/Catalyst/Step6.tsx
stackchain: Hmm weird this example was added 8 days ago https://github.com/react-navigation/react-navigation/blob/4e4935ac2584bc1a00209609cc026fa73e12c10a/example/types.check.tsx#L91
`addListener` doesn't return anything (on types) https://github.com/react-navigation/react-navigation/blob/main/packages/core/src/types.tsx#L127

stackchain: > i think removeListener is deprecated `const unsubscribe = navigation.addListener(...)`

it looks like is not marked as deprecated, anyway updating to the unsub/sub made the job easier

stackchain: > also, there is a `listeners` prop for each screen, can we use that instead? https://reactnavigation.org/docs/navigation-events/#listeners-prop-on-screen

Tried, but the event object is not coming with the same type.

wolverineks (approved):

PR #1953, opened 2022-04-12
Update pr-checks.yml

stackchain:


.github/workflows/pr-checks.yml
```diff
@@ -3,16 +3,6 @@ name: PR Checks
 on: [push]
 
 jobs:
-  flow:
```
wolverineks, line 6: does anything check typescript? `yarn tsc`?

stackchain: right now no, feel free to add to the scripts+workflow, later I set it to the required status check

PR #1955, opened 2022-04-14
Fix open wallet after resync

stackchain:
- After the wallet resync, the wallet needs to be opened again.


ios/emurgo.xcodeproj/project.pbxproj
```diff
@@ -1345,7 +1345,7 @@
 				BUNDLE_ID_SUFFIX = "";
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_IDENTITY = "iPhone Distribution";
-				CURRENT_PROJECT_VERSION = 372;
```
wolverineks, line 1348: why does this change need native changes?

src/SelectedWallet/WalletSelection/WalletSelectionScreen.tsx
```diff
@@ -73,10 +71,11 @@ export const WalletSelectionScreen = () => {
       await showErrorDialog(errorMessages.itnNotSupported, intl)
       return
     }
-    if (wallet?.id === walletMeta.id) {
-      return navigateToTxHistory()
+    if (params?.reopen || wallet?.id !== walletMeta.id) {
```
wolverineks, line 74: whats going on here?

wolverineks: we should avoid using params

ios/emurgo.xcodeproj/project.pbxproj
stackchain: When publishing the nightly for our testers I need to bump the build version, otherwise, I can't publish it.

src/SelectedWallet/WalletSelection/WalletSelectionScreen.tsx
stackchain: The optimization to just navigate to the wallet history when the wallet selected is the current one needs to be avoided when the user resyncs the wallet.

stackchain: About the params, my idea is to tackle all params as feature context in a ticket for it.

wolverineks (approved):

PR #1956, opened 2022-04-14
Fix loop reading qr

stackchain:
- Fix the loop when reading invalid qr for ro wallet.


wolverineks (commented):
what's the change?

stackchain:
> what's the change?

Reactivating the scanner without waiting for the dialog was causing a loop.

wolverineks (approved):

PR #1970, opened 2022-05-01
Hotfix/metadata migration

stackchain:


src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,56 @@
+import React, {Component, ErrorInfo, ReactNode} from 'react'
```
wolverineks, line 1: use react-error-boundary package instead

src/components/ErrorBoundary/ErrorBoundary.stories.tsx
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
wolverineks: not needed

src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  render() {
+    if (this.state.hasError) {
+      return (
+        <View
+          style={{
+            flex: 1,
+            justifyContent: 'center',
+            alignItems: 'center',
+            marginHorizontal: 12,
```
wolverineks: avoid margin when possible

src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,56 @@
...
+            paddingTop: 80,
+          }}
+        >
+          <Text>Oops!!! Something went wrong..</Text>
```
wolverineks: is there a figma for this? if not, there should be

wolverineks (changes requested):

legacy/crypto/walletManager.js
```diff
@@ -128,6 +128,27 @@ class WalletManager {
           checksum = w.checksum
         }
         const isHW = data != null && data.isHW != null ? data.isHW : false
+
+        // migration - missing checksum
+        const metadataUpdate: WalletMeta = {...w}
```
wolverineks: can this be pulled out and given a name?
would it make sense to add a test for these 2 migrations?

src/components/ErrorBoundary/ErrorBoundary.tsx
wolverineks: https://github.com/bvaughn/react-error-boundary

stackchain: what are the advantages of using this package?

src/components/ErrorBoundary/ErrorBoundary.tsx
stackchain: atm there is not :/

legacy/crypto/walletManager.js
stackchain: 🤔

src/components/ErrorBoundary/ErrorBoundary.tsx
stackchain: I was checking it, is pretty stuffed with things that I don't see we using for now, like a hook to set the error, a throw will do the job, what I liked was the "reset" impl, I will sync with Dima to cover on Figma, perhaps we can add it later.

wolverineks: it's excellent, very popular, well maintained by a Facebook relay developer, fully tested

stackchain: Ticket created to add it, don't want to add any packages in prod, then we can move the current errorboundary as the fallback screen of it.

src/appStorage.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+function migrateAttribute(attr: keyof WalletMeta, walletMetaUpdated: Partial<WalletMeta>): Promise<void> {
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const id = walletMetaUpdated!.id
+  Logger.warn(`migrateAttribute::${attr} of wallet ${id}`)
+  return storage.write(`/wallet/${id}`, walletMetaUpdated)
+}
+
+function isWalletMeta(walletMeta?: WalletMeta | object | undefined): walletMeta is WalletMeta {
```
wolverineks: `?` implies undefined

wolverineks: when migrating to ts, i think `unknown` might be the best type

src/appStorage.ts
```diff
@@ -0,0 +1,129 @@
...
+    'isEasyConfirmationEnabled' in walletMeta 
+      && typeof walletMeta.isEasyConfirmationEnabled === 'boolean' &&
+    'checksum' in walletMeta 
+      && typeof walletMeta.checksum === 'object' &&
+    ('provider' in walletMeta 
+      && typeof walletMeta.provider === 'string'
+      || !('provider' in walletMeta)) &&
+    'walletImplementationId' in walletMeta 
+      && typeof walletMeta.walletImplementationId === 'string' 
+      && Object.values(WALLET_IMPLEMENTATION_REGISTRY).indexOf(walletMeta.walletImplementationId) > -1 &&
```
wolverineks: .includes

wolverineks (approved):

src/appStorage.ts
stackchain: 👍  copy and paste from the old validator

stackchain:
Release: https://github.com/Emurgo/yoroi-mobile/releases/tag/4.7.12

PR #1986, opened 2022-05-20
Update faq menu option

banklesss:


legacy/i18n/locales/en-US.json
```diff
@@ -3,7 +3,8 @@
   "menu.allWallets": "All Wallets",
   "menu.catalystVoting": "Catalyst Voting",
   "menu.settings": "Settings",
-  "menu.faq": "FAQ",
+  "menu.faqTitle": "Any Questions?",
+  "menu.faqLink": "ASK OUR SUPPORT TEAM",
```
stackchain: I'd suggest capitalizing on the view only.
```suggestion
  "menu.faqLink": "Ask our support team",
```

legacy/i18n/locales/en-US.json
```diff
@@ -3,7 +3,8 @@
   "menu.allWallets": "All Wallets",
   "menu.catalystVoting": "Catalyst Voting",
   "menu.settings": "Settings",
-  "menu.faq": "FAQ",
+  "menu.faqTitle": "Any Questions?",
```
stackchain: I think we need to update the design.
```suggestion
  "menu.faqTitle": "Any question?",
```

src/Menu/Menu.tsx
```diff
@@ -2,19 +2,22 @@ import {defineMessage} from '@formatjs/intl'
...
 import {useIntl} from 'react-intl'
-import {Linking, ScrollView, TouchableOpacity, View} from 'react-native'
+import {Image, Linking, ScrollView, StyleSheet, TouchableOpacity, View} from 'react-native'
 import {SafeAreaView} from 'react-native-safe-area-context'
 import {useSelector} from 'react-redux'
 
 import {Text} from '../../legacy/components/UiKit'
 import {CONFIG} from '../../legacy/config/config'
 import {tokenBalanceSelector} from '../../legacy/selectors'
+import FaqImage from '../assets/img/icon/shape.png'
```
stackchain, line 12: I'd suggest a name more specific, also please sync with design, to get `@2x` `@3x`, the high-resolution images.

yarn.lock
```diff
@@ -1480,9 +1480,9 @@
   integrity sha512-6U71C2Wp7r5XtFtQzYrW5iKFT67OixrSxjI4MptCHzdSVlgabczzqLe0ZSgnub/5Kp4hSbpDB1tMytZY9pwxxA==
 
 "@emurgo/cardano-serialization-lib-nodejs@^9.1.0":
```
stackchain: I see this update in the `serialization-lib` used for the tests, but I don't see the update in the `package.json`, you can add the `package.json` or drop this file.

stackchain (changes requested):

legacy/i18n/locales/en-US.json
stackchain: I also believe that the text for the CTA is `Find answers on Yoroi Helpdesk`

src/Menu/Menu.tsx
wolverineks: or svg

src/Menu/Menu.tsx
```diff
@@ -37,62 +40,77 @@ export const Menu = () => {
...
+          <HR />
+
+          <Settings //
+            label={strings.settings}
+            onPress={navigateTo.settings}
+            left={<Icon.Gear size={24} color="#6B7384" />}
+          />
+          <HR />
+        </View>
+
```
wolverineks, line 69: the spacing should be accomplished by using a `<Spacer fill />`

src/Menu/Menu.tsx
wolverineks: actually, see if you can get this to work
`<Icon.QuestionMark />`

wolverineks (changes requested):

src/Menu/Menu.tsx
wolverineks: and add it to the icon gallery story while youre at it :thumbsup:

PR #1993, opened 2022-05-23
Fixing overlap in the wallet details header and wallets list item name

banklesss:


src/navigation.ts
```diff
@@ -28,8 +28,8 @@ export const defaultStackNavigationOptionsV2: StackNavigationOptions = {
     fontSize: 16,
     fontFamily: 'Rubik-Medium',
   },
-  headerRightContainerStyle: {
-    paddingRight: 12,
+  headerTitleContainerStyle: {
+    marginHorizontal: 45,
```
stackchain: We avoid using margin, margin is evil.

stackchain (changes requested):

src/navigation.ts
banklesss: That header title is absolute positioned by default, thats the reason why was overlapped. I'm thinking of adding a width of 70% to the full width of the screen. what do you think?

wolverineks: percentages are hacks
margins are terrible hacks
react navigation is hacky
hack if you must

wolverineks (approved):

src/navigation.ts
banklesss: hahaha. Ok I will add a 70% of width if everyone agrees

PR #2006, opened 2022-06-04
[YOMO-163] New icon set

banklesss:


src/TxHistory/ModalInfo/ModalInfo.tsx
```diff
@@ -39,10 +38,3 @@ const useStrings = () => {
     infoTitle: intl.formatMessage(messages.infoTitle),
   }
 }
-
-const styles = StyleSheet.create({
-  infoIcon: {
-    alignSelf: 'center',
```
stackchain, line 45: I think this will change the alignment from `center` to `left`.

src/components/Icon/NftAsset.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import Svg, {Path} from 'react-native-svg'
+
+type Props = {
+  size?: number
+  color?: string
+}
+
+export const NftAsset = ({size = 36, color = 'black'}: Props) => (
```
stackchain, line 9: Would you be so kind to check with the design team, it looks a little bit different from the TabBar that is on Figma.

stackchain (changes requested):

src/TxHistory/ModalInfo/ModalInfo.tsx
banklesss: yeah you are totally right

stackchain (approved):

PR #2008, opened 2022-06-05
fix languages

wolverineks:


src/i18n/LanguageProvider.tsx
```diff
@@ -74,3 +74,10 @@ type LanguageContext = {
   selectLanguageCode: SaveLanguageCode
   supportedLanguages: SupportedLanguages
 }
+
+const systemLanguageCode = Platform.select({
+  ios: () =>
+    NativeModules.SettingsManager.settings.AppleLocale || NativeModules.SettingsManager.settings.AppleLanguages[0],
```
stackchain, line 86: I think we'll need to check if our translations support the user device's language, otherwise, it will list all text with the default text `!!!`, when the application state is empty (new installation).

stackchain (approved):

PR #2014, opened 2022-06-12
chore: Tx sync minor improvement

stackchain:
- Minor tx improvement, step 2, keeping it minimal due to the high chance to cherry-pick into prod


banklesss (approved):

src/yoroi-wallets/cardano/shelley/transactionCache.ts
```diff
@@ -23,7 +21,9 @@ type SyncMetadata = {
 }
 type TransactionCacheState = {
   transactions: Record<string, Transaction>
+  // @deprecated
```
wolverineks, line 24: why are these deprecated?
why not get rid of them altogether?

PR #2029, opened 2022-06-22
4.7.16 rc.3

stackchain:
- [x] #2021


Nebyt (approved):
I have reproduced the problem on mainnet on version 4.7.15.
Checked the fix on version 4.7.16. Fixed. :rocket:

PR #2052, opened 2022-07-06
add balance banner stories

wolverineks:


wolverineks:
check the error story

src/hooks/index.ts
```diff
@@ -648,9 +648,8 @@ export const useExchangeRate = ({
   const query = useQuery<number, Error>({
     suspense: true,
     staleTime: 60000,
-    retry: 3,
     retryDelay: 1000,
-    queryKey: [to],
+    queryKey: [wallet.id, 'exchangeRate', {to}],
```
stackchain, line 652: I think we should leverage the query cache across the wallets.

stackchain: our backend sometimes is very unstable due to the load, and often we receive some errors, perhaps a few retries could save the user from a bad experience on the first load

wolverineks: what about ergo wallets?

wolverineks: cache across network?

wolverineks: either way, each key should be loggable / readable

stackchain: the symbol will be different for the default asset ERG, the network won't affect the price

stackchain: > either way, each key should be loggable / readable

easily convinced 👍

stackchain (approved):

PR #2083, opened 2022-07-20
fix: Display keyhashes

stackchain:
## Important
There are some implementations required in `yoroi-lib` to improve what we display to the users. Right now the lib for mobile doesn't support `EnterpriseAddress` and `PointerAddress`. We are also not resolving the script key hashes.


stackchain:
This implementation should be replaced by `yoroi-lib`.

src/Receive/AddressModal.tsx
```diff
@@ -156,11 +156,15 @@ const useStrings = () => {
 }
 
 const useKeyHashes = (address) => {
-  const [keyHashes, setKeyHashes] = React.useState<null | {staking: string; spending: string}>(null)
+  const [keyHashes, setKeyHashes] = React.useState<{staking: string | null; spending: string | null}>({
```
wolverineks: useState<CardanoKeyHashes>

src/yoroi-wallets/cardano/addressInfo.ts
```diff
@@ -0,0 +1,87 @@
...
+  return keyHash.toBytes().then((bytes) => Buffer.from(bytes ?? '').toString('hex'))
+}
+
+/**
+ * @description Try to resolve the spending and staking key hashes for a bech32 address
+ *
+ * @param {string} address expects to be a bech32
+ * @returns {Promise<CardanoKeyHashes>}
+ */
+export async function getKeyHashes(address: string): Promise<CardanoKeyHashes> {
```
wolverineks: should these be split up and loaded separately?

src/Receive/AddressModal.tsx
```diff
@@ -156,11 +156,15 @@ const useStrings = () => {
...
+    staking: null,
+    spending: null,
+  })
 
   React.useEffect(() => {
-    const addressInfo = new AddressDTOCardano(address)
-    addressInfo.getKeyHashes().then(setKeyHashes)
+    getKeyHashes(address)
+      .then(setKeyHashes)
+      .catch((e) => console.error(e))
```
banklesss:
```suggestion
      .catch(console.error)
```

src/yoroi-wallets/cardano/addressInfo.ts
stackchain: it does

banklesss (approved):

src/yoroi-wallets/cardano/addressInfo.ts
wolverineks: i meant, 2 different hooks, then you can move them into separate components with separate boundaries

stackchain: I think we will always display them together, we are using it only in 1 place, but now I get your point, will do one more iteration, then when lib comes out with the hooks will be ready.

PR #2090, opened 2022-08-01
Send feature isolate data

banklesss:


src/Send/Context/TokenBalanceContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+  )
+}
+
+export const useTokenBalanceContext = () => React.useContext(TokenBalanceContext) || missingProvider()
+
+const missingProvider = () => {
+  throw new Error('TokenBalanceProvider is missing')
+}
+
+const useTokenBalance = (wallet: YoroiWallet): MultiToken => {
```
wolverineks, line 44: can consume directly without context

src/Send/Context/TokenBalanceContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const tokenBalance = (utxos || []).reduce(
+    (acc: MultiToken, remoteValue: RemoteValue) =>
+      acc.joinAddMutable(multiTokenFromRemote(remoteValue, wallet.networkId)),
+    defaultMultiToken,
+  )
+
+  return tokenBalance
+}
+
+const useIsStale = (wallet: YoroiWallet, tokenBalanceRef: MutableRefObject<TokenBalanceRef>) => {
```
wolverineks, line 57: can consume directly without context

src/Send/Context/TokenBalanceContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+const missingProvider = () => {
+  throw new Error('TokenBalanceProvider is missing')
+}
+
+const useTokenBalance = (wallet: YoroiWallet): MultiToken => {
+  const utxos = useUTXOs(wallet)
+  const defaultMultiToken = new MultiToken([], getDefaultNetworkTokenEntry(wallet.networkId))
+
+  const tokenBalance = (utxos || []).reduce(
```
wolverineks, line 48: no more MultiToken

src/Send/Context/TokenBalanceContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+  return tokenBalance
+}
+
+const useIsStale = (wallet: YoroiWallet, tokenBalanceRef: MutableRefObject<TokenBalanceRef>) => {
+  const query = useQuery({
+    queryKey: [wallet.id],
+    queryFn: async () => {
+      let lastTx: LastYoroiTransaction | null = null
+
+      try {
```
wolverineks, line 63: remove try/catch
dont use try/catch in queryFn

src/Send/Context/TokenBalanceContext.tsx
wolverineks, line 57: eliminate `useIsState()`

src/Send/Context/TokenBalanceContext.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+export const useTokenBalanceContext = () => React.useContext(TokenBalanceContext) || missingProvider()
+
+const missingProvider = () => {
+  throw new Error('TokenBalanceProvider is missing')
+}
+
+const useTokenBalance = (wallet: YoroiWallet): MultiToken => {
+  const utxos = useUTXOs(wallet)
+  const defaultMultiToken = new MultiToken([], getDefaultNetworkTokenEntry(wallet.networkId))
```
wolverineks, line 46: this will create a new Multitoken on each render
this could be a performance issue

src/Send/Context/TokenBalanceContext.tsx
wolverineks: this also creates a new multitoken on each render
this could be a performance issues
it could also be a problem due to the unstable identity
**if a object contains a value that doesnt change, then the identity of the object shouldnt change**

PR #2096, opened 2022-08-06
chore: Lint improvements

stackchain:


src/App.tsx
```diff
@@ -17,6 +17,7 @@ import {StorageProvider} from './Storage'
 enableScreens()
 
 if (Platform.OS === 'android') {
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
   if (UIManager.setLayoutAnimationEnabledExperimental) {
```
wolverineks: ?
`if (!!UIManager.setLayoutAnimationEnabledExperimental) {`

src/legacy/utils.ts
```diff
@@ -398,3 +398,23 @@ export const ignoreConcurrentAsyncHandler = <Props, T, R>(
...
+ *
+ * @param value The string to inspect
+ * @return {boolean} Returns true if the string is empty, else false.
+ *
+ * @example isEmptyString('') returns true
+ * @example isEmptyString(' ') returns false
+ * @example isEmptyString(null) returns true
+ * @example isEmptyString(undefined) returns true
+ */
+export function isEmptyString(value: string | null | undefined): value is '' | null | undefined {
```
wolverineks, line 418: given the name `isEmptyString`, this function should validate that the value is of type `string` and has a length of 0
`null` and `undefined` are not strings
what is this function supposed to do?

src/legacy/utils.ts
```diff
@@ -398,3 +398,23 @@ export const ignoreConcurrentAsyncHandler = <Props, T, R>(
...
+  return isEmpty(value)
```
wolverineks, line 419: have you tested this?

src/App.tsx
stackchain: it will conflict with the no-extra-bool check, but I can check against null

src/legacy/utils.ts
stackchain: It's just a temporary wrapper for lodash `isEmpty` until the safeguard PR is merged. Accepting naming suggestions.

src/BiometricAuth/BiometricAuthScreen.tsx
```diff
@@ -73,7 +73,7 @@ export const BiometricAuthScreen = () => {
     <FingerprintScreenBase
       onGoBack={cancelScanning}
       headings={[strings.headings1, strings.headings2]}
-      subHeadings={route.params?.instructions || undefined}
+      subHeadings={route.params?.instructions ?? undefined}
```
wolverineks: what type is `instructions`?
can this be changed to...
`route.params?.instructions ?? undefined` -> `route.params?.instructions`

stackchain: @wolverineks, whenever is possible, please make it as git suggestions so I can commit it directly.

banklesss (approved):

PR #2106, opened 2022-08-17
Fix: Loader icon is stuck

banklesss:


src/hooks/index.ts
```diff
@@ -702,3 +704,18 @@ export const useLogout = () => {
     dispatch(signout())
   }
 }
+
+export const useRefreshOnFocusHack = () => {
```
wolverineks: move this out of here
this is for cross platform hooks only

wolverineks (changes requested):
see comments

src/hooks/index.ts
banklesss: done

stackchain (approved):
WR PR.

src/TxHistory/TxHistoryList/TxHistoryList.tsx
```diff
@@ -58,6 +58,27 @@ export const TxHistoryList = ({onScrollUp, onScrollDown, ...props}: Props) => {
...
 }
 
+// workaround for https://emurgo.atlassian.net/browse/YOMO-199
+// related to https://github.com/facebook/react-native/issues/15694
+export const useRefreshOnFocusHack = () => {
+  const [key, setKey] = React.useState(0)
+  const navigation = useNavigation()
+
+  React.useEffect(() => {
+    if (Platform.OS !== 'ios') {
```
wolverineks, line 68: why add this?

banklesss: Discussed on call

PR #2114, opened 2022-08-24
fix: Clipped views

stackchain:
- Updated the use of `removeClippedSubviews` [reference here](https://reactnative.dev/docs/view#removeclippedsubviews)
- It should be used together with `overflow: hidden`


wolverineks:
what is this supposed to do?

stackchain:
> what is this supposed to do?

"is useful for scrolling content when there are many subviews, most of which are offscreen". It literally clips off views, in our `TxHistory` since we are not paginating the records yet, it does a massive performance improvement, and also helps keep the app at 60fps, even when you have more than 4k txs.

banklesss (approved):

PR #2117, opened 2022-08-25
chore: Reanimating tests

stackchain:


banklesss (approved):

src/legacy/commonUtils.ts
```diff
@@ -59,7 +59,7 @@ export const decryptData = async (ciphertext: string, secretKey: string): Promis
   try {
     return await decryptWithPassword(secretKeyHex, ciphertext)
   } catch (error) {
-    if ((error as Error).message === 'Decryption error') {
```
wolverineks: we should wait for pedro to fix this

wolverineks (changes requested):

src/legacy/commonUtils.ts
stackchain: Changed the testing strategy, for now, checking for `CardanoError`, until it gets fixed)

wolverineks (approved):

PR #2123, opened 2022-08-30
upgrade to cross-csl

wolverineks:


jestSetup.ts
```diff
@@ -31,4 +32,4 @@ jest.mock('react-native-reanimated', () => {
   return Reanimated
 })
 jest.mock('react-native/Libraries/Animated/NativeAnimatedHelper')
-jest.mock('@emurgo/yoroi-lib-mobile', () => mockYoroiLib)
+jest.mock('@emurgo/cross-csl-mobile', () => mockCSL)
```
stackchain: it's mocking `rnhs` already, -> csl -> rnhs (node version).

stackchain:
```suggestion
```

jestSetup.ts
```diff
@@ -2,9 +2,10 @@
 /* This module sets up Jest */
 import 'react-native-gesture-handler/jestSetup'
 
+import * as mockCSL from '@emurgo/cross-csl-nodejs'
```
stackchain:
```suggestion
```

jestSetup.ts
wolverineks: where is that happening?

stackchain: > where is that happening?

inside `@__mocks__` (perhaps we should move all inside setup dunno)

src/yoroi-wallets/Wallet.ts
```diff
@@ -14,8 +13,8 @@ import {Logger} from '../legacy/logging'
 import {getCardanoNetworkConfigById, isJormungandr} from '../legacy/networks'
 import {IsLockedError, nonblockingSynchronize, synchronize} from '../legacy/promise'
 import type {BackendConfig, EncryptionMethod} from '../legacy/types'
-import {NetworkId, WalletImplementationId, YoroiProvider} from './cardano'
-import {AddressChain, AddressChainJSON, Addresses} from './cardano/chain'
+import {CardanoTypes, NetworkId, WalletImplementationId, YoroiProvider} from './cardano'
+import {AddressChain, AddressChain, AddressChainJSON, AddressChainJSON, Addresses} from './cardano/chain'
```
stackchain:
```suggestion
import {AddressChain, AddressChainJSON, Addresses} from './cardano/chain'
```

stackchain (changes requested):
can we test mocking only rnhs

stackchain (approved):

PR #2143, opened 2022-09-11
Remove-auth-redux

wolverineks:


src/auth/AuthProvider.tsx
```diff
@@ -0,0 +1,49 @@
...
+        ...actions,
+      }}
+    >
+      {children}
+    </AuthContext.Provider>
+  )
+}
+
+export const useAuth = () => useContext(AuthContext) || missingProvider()
+
```
stackchain, line 25: perhaps?
```suggestion
export const useAuth = () => {
  const auth = useContext(AuthContext)
  if (!auth) missingProvider()
  return {
    isLoggedIn: auth.status === 'logged-in',
    isLoggedOut: auth.status === 'logged-out',
    ...auth,
  }
}
```

wolverineks: this moves the derivation from the context to the hook, spreading the state into 2 places
there are now 2 types. one for the context, and one for the hook
and it expands the useAuth type to use `boolean` instead of `true` or `false`

wolverineks: we would have to rewrite `useAuth` to something like:
```export const useAuth = () => {
  const auth = useContext(AuthContext) || missingProvider()

  return auth.status === 'logged-in'
    ? {
        status: auth.status,
        login: auth.login,
        logout: auth.logout,
        isLoggedIn: true,
        isLoggedOut: false,
      } as const
    : {
        status: auth.status,
        login: auth.login,
        logout: auth.logout,
        isLoggedIn: false,
        isLoggedOut: true,
      } as const
}
```

banklesss (approved):

stackchain (approved):

PR #2148, opened 2022-09-19
Refactor auth with OS

stackchain:


src/auth/MasterKey.ts
```diff
@@ -0,0 +1,74 @@
+import ExtendableError from 'es6-error'
+import {useMutation, UseMutationOptions} from 'react-query'
+
+import {decryptData, encryptData} from '../legacy/commonUtils'
+import storage from '../legacy/storage'
+import {YoroiWallet} from '../yoroi-wallets'
+
+export class CredentialsNotFound extends ExtendableError {}
+
+export const useRevealMasterKey = (
```
wolverineks: id like to avoid the use of the term "master"
"main" or "root"?

wolverineks: if a hook returns a value, prefer useQuery()

stackchain: > id like to avoid the use of the term "master" "main" or "root"?

Moving to `root`

stackchain: > if a hook returns a value, prefer useQuery()

🤦

banklesss (approved):

storybook/mocks/wallet.ts
```diff
@@ -234,18 +234,18 @@ export const mockYoroiSignedTx: YoroiSignedTx & {mock: true} = {
   mock: true,
 }
 
-export const mockKeyStore = (overrides?: {
-  getData?: typeof KeyStore.getData
-  storeData?: typeof KeyStore.storeData
-  deleteData?: typeof KeyStore.deleteData
+export const mockMasterKey = (overrides?: {
```
wolverineks: `Partial<MasterKey>`

storybook/mocks/wallet.ts
```diff
@@ -234,18 +234,18 @@ export const mockYoroiSignedTx: YoroiSignedTx & {mock: true} = {
...
   ({
-    getData: async (_keyId, _encrpytionMethod, _message, password, _intl) => {
+    reveal: async (password) => {
       if (password !== 'password') throw new Error('Invalid Password')
 
       return 'masterkey'
     },
-    storeData: async () => undefined,
-    deleteData: async () => undefined,
+    keep: async () => undefined,
```
wolverineks: `throw new Error("Not implemented")`?

src/auth/authOS.ts
```diff
@@ -0,0 +1,246 @@
...
+    defaultMessage: '!!!Too many attempts',
+  },
+  unknownError: {
+    id: 'components.send.biometricauthscreen.UNKNOWN_ERROR',
+    defaultMessage: '!!!Unknown error!',
+  },
+})
+
+const authMethodKey = '/appSettings/authMethod'
+const iosSupportedBioTypes = [Keychain.BIOMETRY_TYPE.TOUCH_ID, Keychain.BIOMETRY_TYPE.FACE_ID]
```
wolverineks: why does yoroi-mobile not support all biometry types?

src/auth/authOS.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  return decoder
+}
+
+export async function isOSAuthSupported() {
+  const checker = Platform.select({
+    android: () =>
+      Keychain.getSupportedBiometryType().then(
+        (supportedBioType) => supportedBioType === Keychain.BIOMETRY_TYPE.FINGERPRINT,
```
wolverineks: why only support fingerprint? can yoroi-mobile support all biometry types?

src/auth/authOS.ts
```diff
@@ -0,0 +1,246 @@
...
+      ),
+    ios: () =>
+      Promise.all([
+        Keychain.canImplyAuthentication({
```
wolverineks: im not sure what this is
if the user has a biometry type, is that good enough?

storybook/mocks/wallet.ts
stackchain: hmm, I think if someone wants another behavior than the default"ed" mock, it should pass it thought `overrides`, when using this mock I'd expect to have these methods mocked doing nothing.

src/auth/authOS.ts
stackchain: it would be restricted to `keychain` impl but yes that is a no-brainer.

src/auth/authOS.ts
stackchain: on `iOS` you can disable bio auth for that specific app

src/auth/authOS.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+export const useChangeAuthMethod = (storage: Storage, options?: UseMutationOptions<void, Error, AuthMethod>) => {
+  const mutation = useMutationWithInvalidations({
+    mutationFn: (authMethod) => storage.setItem(authMethodKey, JSON.stringify(authMethod)),
+    invalidateQueries: [['authMethod']],
+    ...options,
+  })
+
+  return mutation.mutate
```
wolverineks: why not return the whole mutation? what about errors and loading?

src/auth/authOS.ts
```diff
@@ -0,0 +1,246 @@
...
+}
+
+export const useAuthOsErrorDecoder = () => {
+  const strings = useStrings()
```
wolverineks: strings doesnt seem auth related
can this be done in the component file?

src/auth/authOS.ts
```diff
@@ -0,0 +1,246 @@
...
+  const decoder = (error: Error | null | undefined) => {
+    if (!error || /code: 13/.test(error?.message)) return // 13 = Cancelled by user
+    if (/code: 7/.test(error?.message)) return strings.tooManyAttempts
+    return strings.unknownError
+  }
+
+  return decoder
+}
+
+export async function isOSAuthSupported() {
```
wolverineks: ~~async~~

wolverineks: isOSAuthSupported -> canEnableOsAuth?

src/auth/authOS.ts
```diff
@@ -0,0 +1,246 @@
...
+        ([canAuth, supportedBioType]) =>
+          supportedBioType != null && iosSupportedBioTypes.includes(supportedBioType) && canAuth,
+      ),
+    default: () => Promise.reject('OS Authentication is not supported'),
+  })
+
+  return checker()
+}
+
+const useStrings = () => {
```
wolverineks: yeah, this is coupling react-intl to these hooks...

wolverineks:
what ever happened to `wallet.storage` and `wallet.secureStorage`?

src/auth/authOS.ts
stackchain: I copied to `useSaveAuthMethod` and forgot to delete it 🤦

src/auth/authOS.ts
stackchain: I didn't want to create a helpers fns file, `canEnableOsAuth` is the hook's name.

stackchain: anyway, supported doesn't look correct here

stackchain:
> what ever happened to `wallet.storage` and `wallet.secureStorage`?

next iteration maybe, still not sure about it, probably it would be a facade/proxy cuz we need it as well at the app level, so it'd be more like a key sanitization, perhaps blend both in one storage like in mmkv package, still dunno, lets see

src/auth/authOS.ts
stackchain: hmm blowing the tests 🤔

wolverineks:
what do you think about `useKeychain`, or `useAuthenticate `?

src/Settings/LinkAuthWithOs/LinkAuthWithOsScreen.tsx
```diff
@@ -0,0 +1,150 @@
...
+import {StyleSheet} from 'react-native'
+
+import {useAuthOsAppKey, useAuthOsErrorDecoder, useLoadSecret, useSaveAuthMethod, useSaveSecret} from '../../auth'
+import {Button} from '../../components'
+import globalMessages from '../../i18n/global-messages'
+import {isEmptyString} from '../../legacy/utils'
+import {OsAuthBaseScreen} from '../../OsAuth'
+import {useStorage} from '../../Storage'
+
+export const LinkAuthWithOsScreen = () => {
```
wolverineks: `EnableLoginWithOS`?

wolverineks:
```
const {enable, disable, login} = useLoginWithOS()
```
?

src/Settings/LinkAuthWithOs/LinkAuthWithOsScreen.tsx
```diff
@@ -0,0 +1,150 @@
...
+              authenticationPrompt: {
+                title: strings.authorize,
+                cancel: strings.cancel,
+              },
+            },
+            {
+              onSuccess: () => {
+                saveAuthMethod('os', {
+                  onSuccess: () => navigation.goBack(),
+                  onError,
```
wolverineks: pyramid of death

wolverineks:
i find this api to be more readable. can we make something like this work?

```
const {authenticate, isAuthenticated, isLoading, error} = useAuthenticate({ onSuccess: login })
const loginType = useLoginType()
const {enable, disable, login} = useLoginWithOS()
const {data, loadData, isLoading, error} = useSecureStorage({ storage: wallet.secureStorage, key })
const {data, isLoading, error} = useStorage({ storage: wallet.storage, key })
```

src/AppNavigator.tsx
```diff
@@ -3,119 +3,123 @@ import {NavigationContainer, useNavigationContainerRef} from '@react-navigation/
...
     <Stack.Navigator screenOptions={{headerShown: false}}>
       {/* Initial Route */}
       <Stack.Group>
         {isMaintenance && <Stack.Screen name="maintenance" component={MaintenanceScreen} />}
-        {!isAppSetupComplete && !hasAnyWallet && <Stack.Screen name="first-run" component={FirstRunNavigator} />}
+        {!isTosAccepted && <Stack.Screen name="first-run" component={FirstRunNavigator} />}
       </Stack.Group>
 
       {/* Not Authenticated */}
       {isLoggedOut && hasAnyWallet && (
```
wolverineks: why does having wallets affect the logged out ui?

wolverineks:
can we limit this pr to just migrating android to react-native-keychain, and refactor is a separate pr?

src/Settings/LinkAuthWithOs/LinkAuthWithOsScreen.tsx
stackchain: `use-Login-` that enable + disable, and login? I'd expect `login` login only.

src/Settings/LinkAuthWithOs/LinkAuthWithOsScreen.tsx
stackchain: it's more like a callback hell, yap, we should stop composing mutations, mutations should be atomic,

src/AppNavigator.tsx
stackchain: it was designed for UX, so the user is not prompted for authentication when there is no wallet, I'd like to change it as well.

stackchain:
> i find this api to be more readable. can we make something like this work?
> 
> ```
> const {authenticate, isAuthenticated, isLoading, error} = useAuthenticate({ onSuccess: login })
> const loginType = useLoginType()
> const {enable, disable, login} = useLoginWithOS()
> const {data, loadData, isLoading, error} = useSecureStorage({ storage: wallet.secureStorage, key })
> const {data, isLoading, error} = useStorage({ storage: wallet.storage, key })
> ```

tiny naming issue here, we are using keychain for authentication + authorization for os, and PIN for authentication + root password for authorization. sending you the seq diag in slack. There is also a broken boundary since keychain interacts directly with the UI.

stackchain:
> can we limit this pr to just migrating android to react-native-keychain, and refactor is a separate pr?

🤔 I don't think it'd worth.

stackchain:
> > i find this api to be more readable. can we make something like this work?
> > ```
> > const {authenticate, isAuthenticated, isLoading, error} = useAuthenticate({ onSuccess: login })
> > const loginType = useLoginType()
> > const {enable, disable, login} = useLoginWithOS()
> > const {data, loadData, isLoading, error} = useSecureStorage({ storage: wallet.secureStorage, key })
> > const {data, isLoading, error} = useStorage({ storage: wallet.storage, key })
> > ```
> 
> tiny naming issue here, we are using keychain for authentication + authorization for os, and PIN for authentication + root password for authorization. sending you the seq diag in slack. There is also a broken boundary since keychain interacts directly with the UI.

About the `useSecureStorage` the decryption will happen where? on `loadData`? What will hold `data`?

wolverineks:
> > can we limit this pr to just migrating android to react-native-keychain, and refactor is a separate pr?
> 
> 🤔 I don't think it'd worth.

why not?

stackchain:
> > > can we limit this pr to just migrating android to react-native-keychain, and refactor is a separate pr?
> > 
> > 
> > 🤔 I don't think it'd worth.
> 
> why not?

some bio/init code will need to be updated twice.

src/Settings/LinkAuthWithOs/LinkAuthWithOsScreen.tsx
wolverineks: if we consider "login with x" as an entity, then this hook can wrap/provide everything

wolverineks: maybe `useOsAuth()`?

wolverineks: and `OsAuth` and `PinAuth` are the entities

src/auth/Keychain.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export function saveSecret({key, value}: {key: string; value: string}) {
+  return Keychain.setGenericPassword(key, value, {
+    service: key,
+    accessControl: Keychain.ACCESS_CONTROL.BIOMETRY_ANY_OR_DEVICE_PASSCODE,
+    accessible: Keychain.ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
+    securityLevel: Keychain.SECURITY_LEVEL.SECURE_HARDWARE,
+  }).then((result) => {
+    if (result === false) throw new Error('Unable to store secret')
+    return result as Keychain.Result
```
wolverineks: why return anything from a mutation?

src/auth/authOS.ts
```diff
@@ -0,0 +1,212 @@
...
+import {Storage} from '../Storage'
+import {WalletJSON, walletManager, YoroiWallet} from '../yoroi-wallets'
+import {AuthMethod} from '../yoroi-wallets/types'
+import * as Keychain from './Keychain'
+
+export const useCanEnableAuthOs = (options?: UseQueryOptions<boolean, Error>) => {
+  const query = useQuery({
+    queryKey: ['canEnableAuthOs'],
+    queryFn: Keychain.canEnableAuthOs,
+    ...options,
```
wolverineks, line 27: does this need to be refetched when the app is blurred/focused?

src/auth/authOS.ts
```diff
@@ -0,0 +1,212 @@
...
+  })
+
+  return {
+    ...query,
+    canEnableOsAuth: query.data,
+  }
+}
+
+export const AUTH_METHOD_OS: AuthMethod = 'os'
+export const useEnableAuthWithOs = (
```
wolverineks, line 45: is there a corresponding useDisableAuthWithOS?

src/auth/authOS.ts
```diff
@@ -0,0 +1,212 @@
...
+) => {
+  const mutation = useMutationWithInvalidations({
+    mutationFn: () =>
+      storage
+        .getItem(INSTALLATION_ID_KEY)
+        .then((data) => {
+          if (data != null) return data
+          throw new Error('Invalid state')
+        })
+        .then(JSON.parse)
```
wolverineks: `JSON.parse` returns `any`

src/auth/Keychain.ts
```diff
@@ -0,0 +1,66 @@
...
+        Keychain.getSupportedBiometryType(),
+      ]).then(
+        ([canAuth, supportedBioType]) =>
+          supportedBioType != null && iosSupportedBioTypes.includes(supportedBioType) && canAuth,
+      ),
+    default: () => Promise.reject('OS Authentication is not supported'),
+  })()
+}
+
+const iosSupportedBioTypes = [Keychain.BIOMETRY_TYPE.TOUCH_ID, Keychain.BIOMETRY_TYPE.FACE_ID]
```
wolverineks: why limit biometry types?

src/AppNavigator.tsx
```diff
@@ -201,3 +192,53 @@ const useAutoLogout = () => {
...
+      if (authMethod?.isOS && appState === 'active' && !hasTurnedOffOSAuth) refetch()
+    })
+    return () => appStateSubscription?.remove()
+  }, [authMethod, refetch, hasTurnedOffOSAuth])
+
+  React.useEffect(() => {
+    if (hasTurnedOffOSAuth) {
+      Alert.alert(strings.biometricsChangeTitle, strings.biometricsChangeMessage)
+
+      logout()
```
wolverineks, line 169: hooks that return a value should not have side effects

wolverineks: maybe split this into multiple hooks

src/AppMigrations.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+import AppNavigator from './AppNavigator'
+import {useMigrateAuthMethod} from './auth'
+import {useStorage} from './Storage'
+
+export const AppMigrations = () => {
+  const storage = useStorage()
+  const {isSuccess, migrate} = useMigrateAuthMethod(storage)
+
+  React.useEffect(() => {
```
wolverineks: is this fundamentally different than initApp? can it be part of the same process? same hook?

src/AppNavigator.tsx
```diff
@@ -2,120 +2,115 @@ import {useReduxDevToolsExtension} from '@react-navigation/devtools'
...
+      authenticationPrompt: {
+        cancel: strings.cancel,
+        title: strings.authorize,
+      },
+      storage,
+    },
+    {
+      onSuccess: login,
+      onSettled: () =>
+        RNBootSplash.hide({
```
wolverineks: what does the splash screen have to do with auth?

wolverineks: oh I see, this should be in a conditional component

src/AppNavigator.tsx
```diff
@@ -144,18 +138,9 @@ const NavigatorSwitch = () => {
 }
 
 const CreatePinScreenWrapper = () => {
-  const dispatch = useDispatch()
```
wolverineks, line 152: do we need a wrapper anymore?

src/AppNavigator.tsx
```diff
@@ -201,3 +192,53 @@ const useAutoLogout = () => {
     duration: 120 * 1000,
   })
 }
+
+export const useAuthAction = (storage: Storage) => {
```
wolverineks: what is auth action?

src/AppNavigator.tsx
wolverineks: something like
```
authAction === "os" <LoginWithOS /> : <LoginWithPin />
```

src/AppNavigator.tsx
```diff
@@ -2,120 +2,115 @@ import {useReduxDevToolsExtension} from '@react-navigation/devtools'
...
-    }
-  }, [hasAnyWallet, isLoggedOut, isSystemAuthEnabled, canEnableBiometrics, isMaintenance, strings])
-
   return (
     <Stack.Navigator screenOptions={{headerShown: false}}>
       {/* Initial Route */}
       <Stack.Group>
         {isMaintenance && <Stack.Screen name="maintenance" component={MaintenanceScreen} />}
-        {!isAppSetupComplete && !hasAnyWallet && <Stack.Screen name="first-run" component={FirstRunNavigator} />}
+        {!isTosAccepted && <Stack.Screen name="first-run" component={FirstRunNavigator} />}
```
wolverineks: does this need to be in a group?

src/auth/Keychain.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+describe('canEnableAuthOs()', () => {
+  describe('on Android', () => {
+    it.each`
+      expected | bioType
+      ${true}  | ${Keychain.BIOMETRY_TYPE.FINGERPRINT}
+      ${true}  | ${Keychain.BIOMETRY_TYPE.IRIS}
+      ${true}  | ${Keychain.BIOMETRY_TYPE.FACE}
+    `('should return $expected when type is $bioType', async ({expected, bioType}) => {
+      Platform.select = jest.fn((objs) => objs['android'])
```
wolverineks: what is objs?

src/AppNavigator.tsx
```diff
@@ -201,3 +192,53 @@ const useAutoLogout = () => {
     duration: 120 * 1000,
   })
 }
+
+export const useAuthAction = (storage: Storage) => {
+  const strings = useStrings()
+  const {logout} = useAuth()
+  const {authMethod, isLoading: loadingAuthMethod} = useAuthMethod(storage)
+  const {canEnableOsAuth, isLoading: loadingCanEnableOsAuth, refetch} = useCanEnableAuthOs()
```
wolverineks: can this be loaded with suspense?

src/AppNavigator.tsx
```diff
@@ -201,3 +192,53 @@ const useAutoLogout = () => {
...
+
+  const isLoading = loadingAuthMethod || loadingCanEnableOsAuth
+  const hasTurnedOffOSAuth = !isLoading && !canEnableOsAuth && authMethod?.isOS
+
+  React.useEffect(() => {
```
wolverineks: this should be in the query hooks

src/AppNavigator.tsx
```diff
@@ -201,3 +192,53 @@ const useAutoLogout = () => {
...
+    if (hasTurnedOffOSAuth) {
+      Alert.alert(strings.biometricsChangeTitle, strings.biometricsChangeMessage)
+
+      logout()
+    }
+  }, [hasTurnedOffOSAuth, strings.biometricsChangeTitle, strings.biometricsChangeMessage, logout])
+
+  if (isLoading) return
+  if (authMethod?.isNone || hasTurnedOffOSAuth) return 'create-link-pin'
```
wolverineks: what is `create-link-pin`?

src/components/LoadingOverlay/LoadingOverlay.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import {ActivityIndicator, StyleSheet, View} from 'react-native'
+import LinearGradient from 'react-native-linear-gradient'
+
+export function LoadingOverlay({loading}: {loading: boolean}) {
```
wolverineks: why did you change from const to function?

src/hooks/index.ts
```diff
@@ -450,7 +421,7 @@ export const useCreatePin = (storage: Storage, options: UseMutationOptions<void,
       const saltHex = cryptoRandomString({length: 2 * 32})
       const nonceHex = cryptoRandomString({length: 2 * 12})
       const encryptedPinHash = await Cardano.encryptWithPassword(pinHex, saltHex, nonceHex, installationIdHex)
-
+      await storage.setItem(AUTH_METHOD_KEY, JSON.stringify(AUTH_METHOD_PIN))
```
wolverineks: is the existence of a pinHash on disk not enough to imply authWithPin?
seems like duplication of state

src/auth/authOS.ts
```diff
@@ -0,0 +1,212 @@
...
+        .then((installationId) => Keychain.saveSecret({key: installationId, value: installationId}))
+        .then(({service}) => Keychain.loadSecret({key: service, authenticationPrompt}))
+        .then(() => storage.setItem(AUTH_METHOD_KEY, JSON.stringify(AUTH_METHOD_OS)))
```
wolverineks: what is going on with all these set's / get's

src/auth/authOS.ts
```diff
@@ -0,0 +1,212 @@
...
+          throw new Error('Invalid state')
+        })
+        .then(JSON.parse)
+        .then((key) => Keychain.loadSecret({key, authenticationPrompt})),
+    ...options,
+  })
+
+  return {
+    authWithOs: mutation.mutate,
+    ...mutation,
```
wolverineks, line 76:
```
return {
  ...mutation,
  authWithOs: mutation.mutate
}
```

src/auth/authOS.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+}
+
+export const useAuthOsWithEasyConfirmation = (
```
wolverineks, line 81: can this be merged with useAuthWithOs?

src/AppNavigator.tsx
wolverineks: is logout even necessary? is the alert not enough?

src/OsAuth/OsLoginScreen/OsLoginScreen.tsx
```diff
@@ -0,0 +1,69 @@
...
+    },
+    {
+      onSuccess: login,
+    },
+  )
+  const decodeAuthOsError = useAuthOsErrorDecoder()
+  const errorMessage = decodeAuthOsError(error)
+
+  return (
+    <OsAuthBaseScreen
```
wolverineks: what is base?

src/OsAuth/OsAuthBaseScreen/OsAuthBaseScreen.tsx
```diff
@@ -71,7 +81,7 @@ export const FingerprintScreenBase = ({headings, subHeadings, buttons, onGoBack,
 
         {error != null && error !== false ? <Text style={styles.error}>{error}</Text> : null}
 
-        <View style={styles.controls}>{buttons}</View>
+        <View style={[buttons.length > 1 ? styles.manyControls : styles.oneControl]}>{buttons}</View>
```
wolverineks: make 2 different named components

src/auth/authOS.ts
stackchain: args typed

src/auth/Keychain.ts
stackchain: that makes sense, expanded to `!=` null

src/auth/authOS.ts
stackchain: yap, it does, if the user disables the auth with OS we need to logout, and ask for a pin creation

src/auth/authOS.ts
stackchain: you only disable it, if you successfully create a PIN, so you disable it by `useEnableAuthWithPin` -> and vice versa.

src/AppNavigator.tsx
stackchain: it logout the user, to force them to create a PIN, it was designed this way, it won't switch to PIN until the user sets one.

src/AppMigrations.tsx
stackchain: it's the same analogy to the DB migrations, it must happen before any code that reads data

src/AppNavigator.tsx
stackchain: I didn't want to navigate the user to another screen on the very first attempt, so whether it succeeds or not, it hides the splash screen.

src/auth/authOS.ts
stackchain: make it a code suggestion next time

src/AppNavigator.tsx
stackchain: the group is there for clarity, do you think is creating noise?

src/auth/authOS.ts
stackchain: ? EasyConfirmation authorizes at the wallet level, and useAuthWithOs authenticate at the application level.

src/components/LoadingOverlay/LoadingOverlay.tsx
stackchain: this file was added, there is no change

src/auth/authOS.ts
stackchain: what do you mean by that?

src/hooks/index.ts
stackchain: hmm, in this case, the authentication method will be derived from something else

src/AppNavigator.tsx
wolverineks: what happens if the server goes into maintenance mode after the user is already logged in?

src/AppNavigator.tsx
wolverineks: why not navigate to create pin?

src/AppNavigator.tsx
wolverineks: refetch in the useCanEnableAuth everytime

src/AppNavigator.tsx
stackchain: yap, this logic happens in the navigator after this 1st attempt

src/AppNavigator.tsx
stackchain: maintenance screen will be displayed, whether the user is logged in or out, yet, is something that was never used, we should consider dropping it

src/AppNavigator.tsx
stackchain: also for clarity, so we can use the navigator with `component={Name}`

src/AppNavigator.tsx
stackchain: if we navigate to create pin and the user goes back, the user will end up in a state where it will fail when trying to submit a tx, then we will have to do some checks in all submit components. -> Also it makes sense logically, if you disable your authentication, the least you can expect is to be logged out.

stackchain (commented):
⬆️

src/AppNavigator.tsx
```diff
@@ -215,17 +214,9 @@ const useOsAuthDisabled = () => {
   return !canEnableOsAuth && authMethod?.OS === true
 }
 
-const useCheckOsAuthDisabled = (osAuthDisabled: boolean, authMethod?: AuthMethodState) => {
+const useCheckOsAuthDisabled = (osAuthDisabled: boolean) => {
   const strings = useStrings()
   const {logout} = useAuth()
-  const {refetch} = useCanEnableAuthOs()
```
wolverineks: are you getting rid of this or moving it?

stackchain: deleted

banklesss (approved):

src/auth/authOS.ts
```diff
@@ -0,0 +1,224 @@
...
+      await storage.setItem(AUTH_METHOD_KEY, JSON.stringify(AUTH_METHOD_OS))
+      return disableAllEasyConfirmation()
+    }
+    if (pin != null) {
+      return storage.setItem(AUTH_METHOD_KEY, JSON.stringify(AUTH_METHOD_PIN))
+    }
+  }
+}
+
+const disableAllEasyConfirmation = () =>
```
wolverineks, line 147: colocate helpers

src/auth/authOS.ts
```diff
@@ -0,0 +1,224 @@
...
+  const [[, authMethod], [, pin], [, isOldSystemAuth], [, installationId]] = await storage.multiGet([
+    AUTH_METHOD_KEY,
+    ENCRYPTED_PIN_HASH_KEY,
+    OLD_OS_AUTH_KEY,
+    INSTALLATION_ID_KEY,
+  ])
+  if (authMethod == null && installationId != null) {
+    const id = JSON.parse(installationId)
+    if (isOldSystemAuth != null && JSON.parse(isOldSystemAuth) === true) {
+      await keychainStorage.write(id, id) // the data here is irrelevant
```
wolverineks: `keychain.initialize()`

src/AppMigrations.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+
+import AppNavigator from './AppNavigator'
+import {useMigrateAuthMethod} from './auth'
+import {useStorage} from './Storage'
+
+export const AppMigrations = () => {
```
wolverineks: `<AppMigrations>` -> `useMigrations(): boolean`
move to `./index.ts`

src/AppNavigator.tsx
```diff
@@ -2,82 +2,99 @@ import {useReduxDevToolsExtension} from '@react-navigation/devtools'
...
 
 const IS_STORYBOOK = env.getBoolean('IS_STORYBOOK', false)
 
 export const AppNavigator = () => {
+  const strings = useStrings()
+  const storage = useStorage()
+  const [isReady, setIsReady] = React.useState(false)
   useAutoLogout()
+  useHideScreenInAppSwitcher()
+  const {authMethod} = useAuthMethod(storage)
```
wolverineks: already suspense, remove curly braces here and from hook

src/AppNavigator.tsx
```diff
@@ -2,82 +2,99 @@ import {useReduxDevToolsExtension} from '@react-navigation/devtools'
...
+  const osAuthDisabled = useOsAuthDisabled()
```
wolverineks: move into `useAuthMethod()`

src/auth/authOS.ts
```diff
@@ -0,0 +1,224 @@
...
+import {AuthMethod} from './types'
+
+export const useCanEnableAuthOs = (options?: UseQueryOptions<boolean, Error>) => {
+  const query = useQuery({
+    queryKey: ['canEnableAuthOs'],
+    queryFn: canEnableAuthOs,
+    suspense: true,
+    ...options,
+  })
+
```
wolverineks, line 29:
```
React.useEffect(() => {
  const unsub = AppState.subscribe(refetch)
}, [])
```

src/AppNavigator.tsx
```diff
@@ -210,3 +205,44 @@ const useAutoLogout = () => {
     duration: 120 * 1000,
   })
 }
+
+const useOsAuthDisabled = () => {
+  const storage = useStorage()
+  const {authMethod} = useAuthMethod(storage)
+  const {canEnableOsAuth} = useCanEnableAuthOs()
```
wolverineks: suspence, remove curly braces

src/AppNavigator.tsx
```diff
@@ -210,3 +205,44 @@ const useAutoLogout = () => {
...
+
+  return !canEnableOsAuth && authMethod?.OS === true
+}
+
+const useCheckOsAuthDisabled = (osAuthDisabled: boolean) => {
+  const strings = useStrings()
+  const {logout} = useAuth()
+
```
wolverineks, line 164: consume `useOsAuthDisabled()` rather than passing in as arg

src/AppNavigator.tsx
stackchain: need it to refresh on focus in the application settings

src/AppNavigator.tsx
stackchain: same thing need to force refetch in the application settings when focusing back

src/hooks/index.ts
```diff
@@ -1005,3 +974,79 @@ export const useBalances = (wallet: YoroiWallet): YoroiAmounts => {
 
   return Utxos.toAmounts(utxos, primaryTokenId)
 }
+
+export const useRefetchOnFocus = (refetch: () => unknown, canRefetch = true) => {
```
wolverineks: ~can this be replaced with `useFocusEffect`?~
what is this for?

wolverineks: what is `canRefetch` for? its never used anywhere

wolverineks: `useFocusEffect` has to take a memoized callback
this implementation will call the callback on ever render

src/Settings/ApplicationSettings/ApplicationSettingsScreen.tsx
```diff
@@ -1,85 +1,70 @@
...
-  const installationId = useSelector(installationIdSelector)
-  const dispatch = useDispatch()
+  const {navigation} = useWalletNavigation()
+
   const {currency} = useCurrencyContext()
+  const storage = useStorage()
+  const {authMethod, refetch: refetchAuthMethod} = useAuthMethod(storage)
+
+  const {authOsEnabledOnDevice, refetch: refetchCanEnableOSAuth} = useAuthOsEnabledOnDevice({
+    // on emulator
```
wolverineks: what is this for?

stackchain: testing without closing and opening the app on the emulator.

src/auth/KeychainStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+  read(KEYCHAIN_APP_AUTH_KEY, authenticationPrompt)
+
+export const KeychainStorage = {
+  read,
+  write,
+  remove,
+  initializeAppAuth,
+  appAuth,
+} as const
+
```
wolverineks, line 37: authOsEnabledOnDevice -> authenticateWithOsEnabled

src/auth/KeychainStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+export async function authOsEnabledOnDevice() {
+  return Platform.select({
```
wolverineks, line 39: if changing this requires minimizing the app, then subscribe to AppState and refetch right here

src/auth/KeychainStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+import * as Keychain from 'react-native-keychain'
+
+async function write(key: string, value: string) {
+  return Keychain.setGenericPassword(key, value, {
+    service: key,
+    accessControl: Keychain.ACCESS_CONTROL.BIOMETRY_ANY_OR_DEVICE_PASSCODE,
+    accessible: Keychain.ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
+    securityLevel: Keychain.SECURITY_LEVEL.SECURE_HARDWARE,
+  }).then((result) => {
+    if (result === false) Promise.reject('Unable to store secret')
```
wolverineks: reject with error, not a string?

src/auth/KeychainStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+}
+
+async function remove(key: string) {
+  return Keychain.resetGenericPassword({
+    service: key,
+  })
+}
+
+const KEYCHAIN_APP_AUTH_KEY = 'os-authentication'
+const initializeAppAuth = () => write(KEYCHAIN_APP_AUTH_KEY, KEYCHAIN_APP_AUTH_KEY) // value is irrelevant
```
wolverineks: store "", or null, or undefined, anything unimportant

src/auth/KeychainStorage.ts
stackchain: Enabled where? on the device or inside the app?

src/hooks/index.ts
```diff
@@ -971,3 +940,65 @@ export const useBalances = (wallet: YoroiWallet): YoroiAmounts => {
 
   return Utxos.toAmounts(utxos, primaryTokenId)
 }
+
+export const AUTH_METHOD_KEY = '/appSettings/authMethod'
+export const useAuthMethod = (storage: Storage, options?: UseQueryOptions<AuthMethodState, Error>) => {
```
wolverineks: i prefer ~`AuthMethodState`~ -> `'os' | 'pin' | undefined`

src/auth/KeychainStorage.ts
stackchain: like the idea to subscribe to the AppState, what do you mean by "right here"

src/hooks/index.ts
```diff
@@ -971,3 +940,65 @@ export const useBalances = (wallet: YoroiWallet): YoroiAmounts => {
...
+      return {
+        PIN: false,
+        OS: false,
+        None: true,
+      } as const
+    },
+    ...options,
+  })
+
+  // refetch on settings
```
wolverineks: what is this comment for?

src/hooks/index.ts
```diff
@@ -971,3 +940,65 @@ export const useBalances = (wallet: YoroiWallet): YoroiAmounts => {
...
+  return {
```
wolverineks: suspense query, return the value directly

src/auth/KeychainStorage.ts
stackchain: is the keyword `os-authentication` important ?

src/hooks/index.ts
wolverineks: if we know what will invalidate the query, and can subscribe to those events, then refetch in this hook

stackchain: the reason why.is not returning the data directly

src/hooks/index.ts
stackchain: ⬆️

src/hooks/index.ts
```diff
@@ -971,3 +940,65 @@ export const useBalances = (wallet: YoroiWallet): YoroiAmounts => {
...
+export const AUTH_METHOD_KEY = '/appSettings/authMethod'
+export const useAuthMethod = (storage: Storage, options?: UseQueryOptions<AuthMethodState, Error>) => {
+  const query = useQuery({
+    suspense: true,
+    queryKey: ['authMethod'],
+    queryFn: async () => {
+      const authMethod: AuthMethod | '' = await Promise.resolve(AUTH_METHOD_KEY)
+        .then(storage.getItem)
+        .then((storedAuthMethod) => (!isEmptyString(storedAuthMethod) ? storedAuthMethod : '""'))
+        .then(JSON.parse)
```
wolverineks: define a custom parse function that guarantees valid data

```
const parseAuthMethod = (data: any): data is 'os' | 'pin' | undefined => {
  if (!data) return undefined
  try {
    const parsed = JSON.parse(data)
    // return valid data
  } catch (error) {
    // return valid data
  }
  
}
```

src/auth/KeychainStorage.ts
```diff
@@ -0,0 +1,59 @@
...
+const appAuth = (authenticationPrompt: Keychain.Options['authenticationPrompt']) =>
```
wolverineks: is the return data used for anything? if not, return void

wolverineks: ~appAuth~ -> `authenticate`

src/Settings/ApplicationSettings/ApplicationSettingsScreen.tsx
wolverineks: thats more of an e2e test, isnt it?

wolverineks: we should leave that to denis

src/auth/authOS.ts
```diff
@@ -0,0 +1,193 @@
...
+export const useEnableAuthWithOs = (
+  {authenticationPrompt, storage}: {authenticationPrompt: AuthenticationPrompt; storage: Storage},
+  options?: UseMutationOptions<void, Error>,
+) => {
+  const mutation = useMutationWithInvalidations({
+    mutationFn: () =>
+      KeychainStorage.initializeAppAuth()
+        .then(() => KeychainStorage.appAuth(authenticationPrompt))
+        .then(() => storage.setItem(AUTH_METHOD_KEY, JSON.stringify(AUTH_METHOD_OS)))
+        // current behavior - on auth with os it deletes the pin stored
```
wolverineks: "current" is a really problematic word
does this mean the code im looking at right now? or the code on develop before this pr is merged?
time and order are dependencies that we should strive to eliminate

wolverineks: stick to defining rules and relationships between data
no time or order required

src/AppNavigator.tsx
```diff
@@ -201,12 +189,60 @@ const messages = defineMessages({
...
+      // when using OS auth and app is active again needs to check if still enabled
+      if (appState === 'active' && authMethod.OS) {
+        queryClient.invalidateQueries(['useAuthAction'])
+        refetch()
+      }
+    })
+    return () => appStateSubscription?.remove()
+  }, [authMethod.OS, queryClient, refetch])
+
+  React.useEffect(() => {
```
wolverineks: instead of logging the user out, can we unmount the authenticated root component and display CreatePin, or whatever, then remount the authenticated root component?

src/hooks/index.ts
stackchain: if (!data) return true?

src/auth/KeychainStorage.ts
stackchain: changed to invalidate the queries.

src/auth/KeychainStorage.ts
stackchain: whatever is thrown here will be translated into an unknown error.

src/auth/KeychainStorage.ts
stackchain: `authenticate` can mislead to easy conf. authenticate. that's why is appAuth, any better suggestion?

src/AppNavigator.tsx
stackchain: this is what logout does, it changed the accessible routes.

wolverineks: its async, im talking sync

src/auth/KeychainStorage.ts
wolverineks: it should be one and the same thing
if the user disables the permissions for the app, then the setting inside the app should update too

src/hooks/index.ts
```diff
@@ -971,3 +940,46 @@ export const useBalances = (wallet: YoroiWallet): YoroiAmounts => {
...
+      if (isAuthMethod(authMethod)) return authMethod
+      return Promise.reject(new Error('useAuthMethod invalid data'))
+    },
+    ...options,
+  })
+
+  return query.data
+}
+
+const parseAuthMethod = (data: string | null) => {
```
wolverineks: should return a value, and not throw

is there a fallback value? if not, undefined

wolverineks: data: unknown

stackchain: so 
`JSON.parse(data as string)` ??

stackchain: undefined will work here, but if we fallback to something it means that we are recovering from some state that should have be migrated, isn't ?

src/Settings/EasyConfirmation/EnableEasyConfirmationScreen.tsx
```diff
@@ -4,36 +4,52 @@ import {defineMessages, useIntl} from 'react-intl'
...
+      },
+      onError: (error) => {
+        throw error
+      },
     },
-    onError: (error) => {
-      if (!(error instanceof WrongPassword)) throw error
-      showErrorDialog(errorMessages.incorrectPassword, intl)
+  )
+  const {readRootKey, isLoading: loadingRootKey} = useReadRootKey(
```
wolverineks: move into useEnableEasyConfirmation

src/AppNavigator.tsx
```diff
@@ -88,39 +102,21 @@ const NavigatorSwitch = () => {
       {/* Initial Route */}
       <Stack.Group>
         {isMaintenance && <Stack.Screen name="maintenance" component={MaintenanceScreen} />}
-        {!isAppSetupComplete && !hasAnyWallet && <Stack.Screen name="first-run" component={FirstRunNavigator} />}
+        {!isTosAccepted && <Stack.Screen name="first-run" component={FirstRunNavigator} />}
```
wolverineks: auth-action === 'first-run'

src/Settings/ApplicationSettings/ApplicationSettingsScreen.tsx
```diff
@@ -94,18 +72,11 @@ export const ApplicationSettingsScreen = () => {
...
-        <SettingsItem
-          label={strings.biometricsSignIn}
-          disabled={!isBiometricHardwareSupported || shouldNotEnableBiometricAuth}
-        >
-          <Switch
-            value={isSystemAuthEnabled}
-            onValueChange={onToggleBiometricsAuthIn}
-            disabled={!isBiometricHardwareSupported || shouldNotEnableBiometricAuth}
-          />
+        <SettingsItem label={strings.biometricsSignIn} disabled={!authOsEnabled}>
```
wolverineks: rename biometrics

src/auth/EncryptedStorage.test.ts
```diff
@@ -0,0 +1,35 @@
+import {CredentialsNotFound, EncryptedStorage as ES, StorageKeys} from './EncryptedStorage'
+
+describe('EncryptedStorage', () => {
+  const password = '1234567890'
+  const data = '65656565494949'
+  const key = StorageKeys.rootKey('walletId')
```
wolverineks: beforeEach (reset)

src/auth/EncryptedStorage.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    await ES.write(key, data, password)
+
+    const decrypted = await ES.read(key, password)
+    expect(decrypted).toBe(data)
+    await ES.write(key, newData, password)
+    const decryptedNewData = await ES.read(key, password)
+    expect(decryptedNewData).toBe(newData)
+
+    await expect(ES.remove(key)).resolves
+    await expect(ES.read(key, password)).rejects.toBeInstanceOf(CredentialsNotFound)
```
wolverineks: resolve(undefined) or rejects(invalid key)?

src/auth/authOS.test.ts
```diff
@@ -0,0 +1,71 @@
...
+  const installationId = 'uuidv4'
+  const os = JSON.stringify('os')
+  const pin = JSON.stringify('pin')
+
+  beforeEach(async () => {
+    await storage.clear()
+  })
+
+  it('method = null and no pin/os means new setup, it should remain null', async () => {
+    await storage.setItem(INSTALLATION_ID_KEY, JSON.stringify(installationId))
```
wolverineks: can this be removed?

src/auth/authOS.ts
```diff
@@ -0,0 +1,204 @@
...
+const OLD_OS_AUTH_KEY = '/appSettings/isSystemAuthEnabled'
+const INSTALLATION_ID_KEY = '/appSettings/installationId'
+export const migrateAuthMethod = async (storage: Storage, keychain: typeof Keychain = Keychain) => {
+  const [[, authMethod], [, pin], [, isOldSystemAuth], [, installationId]] = await storage.multiGet([
+    AUTH_METHOD_KEY,
+    ENCRYPTED_PIN_HASH_KEY,
+    OLD_OS_AUTH_KEY,
+    INSTALLATION_ID_KEY,
+  ])
+  if (authMethod == null && installationId != null) {
```
wolverineks: name this state

src/auth/authOS.test.ts
wolverineks: beforeAll(setInstallationId)

src/auth/authOS.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    await storage.setItem(ENCRYPTED_PIN_HASH_KEY, JSON.stringify('encrypted-hash'))
+
+    await migrateAuthMethod(storage, mockKeychain)
+
+    await expect(await storage.getItem(AUTH_METHOD_KEY)).toBe(pin)
+  })
+})
+
+const OLD_OS_AUTH_KEY = '/appSettings/isSystemAuthEnabled'
```
wolverineks: import these keys from authOS.tx

src/auth/Keychain.ts
```diff
@@ -0,0 +1,28 @@
+import {YoroiWallet} from '../yoroi-wallets'
+import {AuthenticationPrompt, KeychainStorage} from './KeychainStorage'
+
+const authenticate = async (authenticationPrompt: AuthenticationPrompt) => {
+  await KeychainStorage.read(KEYCHAIN_APP_AUTH_KEY, authenticationPrompt)
+}
+const initialize = async () => {
```
wolverineks: move into authenticate

src/auth/authOS.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+export const AUTH_METHOD_OS: AuthMethod = 'os'
+export const useEnableAuthWithOs = (
+  {authenticationPrompt, storage}: {authenticationPrompt: AuthenticationPrompt; storage: Storage},
+  options?: UseMutationOptions<void, Error>,
+) => {
+  const mutation = useMutationWithInvalidations({
+    mutationFn: () =>
+      Keychain.initialize()
```
wolverineks: move into Keychain.authenticate

src/auth/authOS.ts
```diff
@@ -0,0 +1,204 @@
...
+  return {
+    ...mutation,
+    authWithOs: mutation.mutate,
+  }
+}
+
+export const useDisableEasyConfirmation = ({id}: {id: YoroiWallet['id']}, options?: UseMutationOptions) => {
+  const mutation = useMutationWithInvalidations({
+    ...options,
+    mutationFn: () => Keychain.removeWalletKey(id).then(() => walletManager.disableEasyConfirmation()),
```
wolverineks: move Keychain.removeWalletKey into walletManager.disableEasyConfirmation

src/auth/authOS.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+export const useEnableEasyConfirmation = (
+  {id}: {id: YoroiWallet['id']},
+  options?: UseMutationOptions<void, Error, string>,
+) => {
+  const mutation = useMutationWithInvalidations({
+    ...options,
+    mutationFn: (rootKey: string) =>
+      Keychain.setWalletKey(id, rootKey).then(() => walletManager.enableEasyConfirmation()),
```
wolverineks: move Ketchain.setWalletkey into walletManager.enableEasyConfirmation

src/hooks/index.ts
```diff
@@ -971,3 +939,36 @@ export const useBalances = (wallet: YoroiWallet): YoroiAmounts => {
 
   return Utxos.toAmounts(utxos, primaryTokenId)
 }
+
+export const AUTH_METHOD_KEY = '/appSettings/authMethod'
+export const useAuthMethod = (storage: Storage, options?: UseQueryOptions<AuthMethod, Error>) => {
```
wolverineks: useAuthSetting

src/legacy/customPin.ts
```diff
@@ -7,7 +7,7 @@ export const encryptCustomPin = async (installationId: string, pin: string) => {
   return pinHash
 }
 
-export const authenticateByCustomPin = async (customPinHash: string, pinCandidate: string): Promise<boolean> => {
+export const authWithOsByCustomPin = async (customPinHash: string, pinCandidate: string): Promise<boolean> => {
```
wolverineks: fix name

wolverineks: remove "custom"

src/yoroi-wallets/walletManager.ts
```diff
@@ -388,16 +322,11 @@ export class WalletManager {
...
-      await this.deleteEncryptedKey('BIOMETRICS')
-      await this.deleteEncryptedKey('SYSTEM_PIN')
-    }
-    await this.deleteEncryptedKey('MASTER_PASSWORD')
-
     await this._wallet.clear()
     await this.closeWallet()
     await storage.remove(`/wallet/${id}/data`)
     await storage.remove(`/wallet/${id}`)
+    await EncryptedStorage.remove(StorageKeys.rootKey(id))
```
wolverineks: `Keychain.removeWalletKey(id)` as well?

wolverineks: or push both of these into `wallet.remove()` or `wallet.delete()`

src/auth/EncryptedStorage.ts
```diff
@@ -0,0 +1,70 @@
...
+  })
+
+  return {
+    writeRootKey: mutation.mutate,
+    ...mutation,
+  }
+}
+
+type StorageKey = `/keystore/${string}-MASTER_PASSWORD`
+export const StorageKeys = {
```
wolverineks: `StorageKeys` -> `EncryptedStorageKeys`?

wolverineks: if the only key ever used with`EncryptedStorage.read` is `rootKey(wallet.id)`, should it accept `wallet` as the first argument and calculate the key inside the function?

wolverineks: or initialize each wallet with an instance of `EncryptedStorage` that is already prefixed with the wallet id?

wolverineks: `wallet.encryptedStorage.read("rootKey", password)`
`wallet.encryptedStorage.write("rootKey", rootKey, password)`

src/auth/authOS.test.ts
stackchain: moved to Settings

src/legacy/customPin.ts
stackchain: deleted (hooks took over)

src/App.tsx
```diff
@@ -33,7 +35,7 @@ const App = () => {
           <StorageProvider>
             <SelectedWalletMetaProvider>
               <SelectedWalletProvider>
-                <AppNavigator />
+                {IS_STORYBOOK ? <StoryBookNavigator /> : <AppNavigator />}
```
wolverineks: lets get rid of this storybook entry point, and go with just the dev menu

src/AppNavigator.tsx
```diff
@@ -10,34 +10,39 @@ import {useSelector} from 'react-redux'
...
 import {OsLoginScreen} from './OsAuth'
+import {AuthSettings} from './Settings/types'
 import {useStorage} from './Storage'
 import StorybookScreen from './StorybookScreen'
 import {WalletInitNavigator} from './WalletInit/WalletInitNavigator'
 import {WalletNavigator} from './WalletNavigator'
 
-const IS_STORYBOOK = env.getBoolean('IS_STORYBOOK', false)
+type AuthAction = 'auth-with-pin' | 'auth-with-os' | 'request-new-pin' | 'first-run'
+const useAuthAction = (authOsEnabled: boolean, authSettings: AuthSettings): AuthAction => {
```
wolverineks: only hooks should start with "use"

src/AppNavigator.tsx
```diff
@@ -149,14 +147,16 @@ const CreatePinScreenWrapper = () => {
...
 
-const StoryBook = () => (
-  <Stack.Navigator
-    screenOptions={{
-      detachPreviousScreen: false /* https://github.com/react-navigation/react-navigation/issues/9883 */,
-    }}
-  >
-    <Stack.Screen name="storybook" component={StorybookScreen} />
-  </Stack.Navigator>
+export const StoryBookNavigator = () => (
```
wolverineks: get rid of this, and just use StorybookScreen in the stack navigator

wolverineks: `<Stack.Screen name="storybook" component={StorybookScreen} options={{headerShown: true}}/>`

banklesss (approved):

wolverineks (approved):

PR #2151, opened 2022-09-19
Remove last redux stuff from hooks

banklesss:


src/Settings/RemoveWallet/RemoveWalletScreen.tsx
```diff
@@ -23,6 +25,8 @@ export const RemoveWalletScreen = () => {
     onMutate: () => resetToWalletSelection(),
     onSuccess: () => {
       dispatch(updateWallets())
+      dispatch(clearUTXOs())
```
wolverineks, line 28: clear utxos and account state should happen even if remove wallet fails

wolverineks: either onMutate or onSettled

wolverineks (changes requested):
see comments

wolverineks (approved):

PR #2154, opened 2022-09-20
move wallet manager to context

wolverineks:


src/yoroi-wallets/walletManager.ts
```diff
@@ -697,3 +697,5 @@ class WalletManager {
 export const walletManager = new WalletManager()
 
 export default walletManager
+
+export const mockWalletManager = {} as WalletManager
```
stackchain, line 701: does it worth mocking in here?

stackchain (approved):
lgtm

src/yoroi-wallets/walletManager.ts
wolverineks: not sure, I want a mock walletManager somewhere

banklesss (approved):

PR #2170, opened 2022-10-05
remove account state from redux

wolverineks:


src/Catalyst/Catalyst.stories.tsx
```diff
@@ -1,8 +1,10 @@
 import {action} from '@storybook/addon-actions'
 import {storiesOf} from '@storybook/react-native'
 import React from 'react'
+import {QueryClient, QueryClientProvider} from 'react-query'
```
banklesss, line 4: In `4.8.0-rc.4` branch is the decorator we added related to this. I suppose it will be merged soon. Perhaps it would be interesting to start using it from here?

wolverineks: we can clean up later

stackchain (approved):

banklesss (approved):

PR #2184, opened 2022-10-20
fix transition bug

wolverineks:
react navigation has a buggy transition animation
disables `detatchPreviousScreen` memory optimization
https://github.com/react-navigation/react-navigation/issues/9883


banklesss (approved):

stackchain (approved):
It's haunting us for a while, ty @wolverineks !

PR #2230, opened 2022-11-23
cleanup mocks

wolverineks:


stackchain (approved):
merging into `develop`, (just pushed the nightlies with the commit number), if we add anything later to `rc.2` I sync it back.

PR #2244, opened 2022-11-29
fix: ledger connection

stackchain:


android/app/build.gradle
```diff
@@ -254,7 +254,8 @@ dependencies {
     implementation "com.facebook.react:react-native:+"  // From node_modules
 
     implementation 'androidx.core:core-ktx:1.3.2'
-    implementation "org.jetbrains.kotlin:kotlin-stdlib:1.4.10"
+    implementation 'androidx.work:work-runtime:2.7.1'
```
stackchain: fix transitive dependencies (maybe it will be needed to enforce)

android/build.gradle
```diff
@@ -8,19 +8,20 @@ buildscript {
         targetSdkVersion = 31
         supportLibVersion = "28.0.0"
         androidXAnnotation = "1.1.0"
-        kotlinVersion = "1.4.10"
+        kotlinVersion = "1.6.10"
```
stackchain: adjust compatibility

package.json
```diff
@@ -141,7 +141,7 @@
     "react-native-crypto": "2.2.0",
     "react-native-device-info": "6.0.1",
     "react-native-easy-markdown": "1.4.1",
-    "react-native-gesture-handler": "^2.3.2",
```
stackchain, line 144: needed to bump kotlin was failing to build

package.json
```diff
@@ -158,7 +158,7 @@
     "react-native-svg": "^7.2.0",
     "react-native-tab-view": "^3.1.1",
     "react-native-vector-icons": "^8.1.0",
-    "react-native-webview": "11.14.4",
```
stackchain, line 161: ibid idem

banklesss (approved):

src/legacy/ledgerUtils.ts
```diff
@@ -492,3 +488,38 @@ export const signTxWithLedger = async (
...
+const BLUETOOTH_SCAN = 'android.permission.BLUETOOTH_SCAN'
+const BLUETOOTH_CONNECT = 'android.permission.BLUETOOTH_CONNECT'
+export const getLedgerPermissions = () => {
+  const permissions: Array<string> = [PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION]
+  if (Platform.Version >= 31) {
+    permissions.push(BLUETOOTH_CONNECT)
+    permissions.push(BLUETOOTH_SCAN)
+  }
+  // hack
+  return permissions as unknown as Permission[]
```
michalszorad: What I thought was to have it less hacky and use the `as` only when needed

```
const permissions: Array<Permission> = [PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION]
  if (Platform.Version >= 31) {
    permissions.push(BLUETOOTH_CONNECT as Permission)
    permissions.push(BLUETOOTH_SCAN as Permission)
  }
  
  return permissions
```

src/legacy/ledgerUtils.ts
```diff
@@ -492,3 +488,38 @@ export const signTxWithLedger = async (
...
+}
+
+export const requestLedgerPermissions = async () => {
+  if (Platform.OS !== 'android') return Promise.resolve()
+
+  const permissions = getLedgerPermissions()
+  const statuses = await PermissionsAndroid.requestMultiple(permissions)
+  const denied = Object.values(statuses).some((value) => value === 'denied')
+
+  return denied ? Promise.reject() : Promise.resolve()
```
michalszorad, line 512: An error message in `Promise.reject` could be helpful to have in the future

michalszorad (approved):

src/legacy/ledgerUtils.ts
stackchain: the message is handled by the `onError` (it's attached to the react-query hook)

src/legacy/ledgerUtils.ts
stackchain: that's a good idea can you add it as suggestion so we can commit it?

michalszorad:
```suggestion
  const permissions: Array<Permission> = [PermissionsAndroid.PERMISSIONS.ACCESS_FINE_LOCATION]
  if (Platform.Version >= 31) {
    permissions.push(BLUETOOTH_CONNECT as Permission)
    permissions.push(BLUETOOTH_SCAN as Permission)
  }
  
  return permissions
```

package.json
wolverineks: what does this mean?

src/legacy/ledgerUtils.ts
wolverineks: id also say its more idiomatic to throw errors in async functions and implicit return
```
...
  if (denied) throw ...
}
```

PR #2260, opened 2022-12-07
Update README.md

dynamicstrategies:
Added a section to Troubleshooting

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,6 +558,39 @@ The output will be something like that
    ssh-keygen -Hf ~/.ssh/known_hosts
    ```
 
+## Linking with 'cc' error
+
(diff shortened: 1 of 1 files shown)


stackchain:
That's correct I think we should change the `react-native-haskell-shelly` script to consider `sw_vers -productVersion`  output, instead of checking for `DEVELOPER_SDK_DIR` env var. @lisicky 

ty @dynamicstrategies

Nebyt:
Thank you @dynamicstrategies ! :raised_hands:

Nebyt (approved):

stackchain:
updating for now, when we update the `rnhs` I'll be reverting from the docs.

PR #2287, opened 2022-12-22
Storage2

wolverineks:
virtual filesystem

### DIFF
--- a/src/yoroi-wallets/storage.ts
+++ b/src/yoroi-wallets/storage.ts
@@ -69,7 +69,7 @@ export const debug = async () => {
 
 export const createPrefixedStorage = (options?: {storage?: PrefixedStorage; path?: string}) => {
   const {storage = makeRootStorage(), path = '/'} = options || {}
-  if (!validatePath(path)) throw new Error("invalid path: paths may not contain '/'")
+  if (!isValidPath(path)) throw new Error('invalid path')
   const isRootStorage = storage.path === '/'
   const isRootPath = path === '/'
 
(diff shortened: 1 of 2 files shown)


src/yoroi-wallets/storage.ts
```diff
@@ -44,3 +44,111 @@ export const makeStorageWithPrefix = (prefix: string): Storage => {
...
+    async multiGet(fileNames: string[]) {
+      const absolutePaths = fileNames.map((fileName) => toAbsolutePath(storage.path, path, fileName))
+      const items = await AsyncStorage.multiGet(absolutePaths)
+      return items.map(([absolutePath, file]) => [toFileName(absolutePath), file])
+    },
+
+    async setItem(fileName: string, file: string) {
+      const absolutePath = toAbsolutePath(storage.path, path, fileName)
+      return AsyncStorage.setItem(absolutePath, file)
+    },
```
michalszorad: It looks like when `fileName` is something like `example/folder/test.jpg`, then `toFileName` will return just `test.jpg`. Should we add a check to `fileName` to ensure it does not contain `/`?

src/yoroi-wallets/storage.ts
```diff
@@ -44,3 +44,111 @@ export const makeStorageWithPrefix = (prefix: string): Storage => {
...
+      return AsyncStorage.multiRemove(absolutePaths)
+    },
+    async clear() {
+      const allKeys = await this.getAllKeys()
+      return this.multiRemove(allKeys)
+    },
+
+    async getAllKeys() {
+      const allKeys = await AsyncStorage.getAllKeys()
+      const regex = isRootStorage && isRootPath ? new RegExp(`^/[^/]*$`) : new RegExp(`^${storagePath}/[^/]*$`)
```
michalszorad, line 127: Do we need to check the `storagePath` does not include any characters that would break / escape this regexp? Eg when `storagePath` is `test|`

michalszorad (approved):

src/yoroi-wallets/storage.ts
```diff
@@ -44,3 +44,111 @@ export const makeStorageWithPrefix = (prefix: string): Storage => {
...
+  const items = await AsyncStorage.multiGet(keys)
+
+  const data = Object.fromEntries(items)
+
+  return data
+}
+
+export const createPrefixedStorage = (options?: {storage?: PrefixedStorage; path?: string}) => {
+  const {storage = makeRootStorage(), path = '/'} = options || {}
+  if (!validatePath(path)) throw new Error("invalid path: paths may not contain '/'")
```
banklesss: Should a test for this branch be added?

src/yoroi-wallets/storage.ts
```diff
@@ -44,3 +44,124 @@ export const makeStorageWithPrefix = (prefix: string): Storage => {
...
+
+      return allKeys.filter((absolutePath) => absolutePath.match(regex)).map(toFilename)
+    },
+  }
+
+  return prefixedStorage
+}
+
+export const isValidPath = (path: string) => {
+  if (path.match(/[\\^$*+?.()|[\]{}]/g)) return false // contains escaping characters
```
stackchain, line 137: imho, chaging from disallowed to allowed will make it easier to read I.e `!x.match(^[a-zA-Z0-9_\-]+$)`

PR #2289, opened 2022-12-27
Lint-cleanups

wolverineks:


src/HW/LedgerConnect/LedgerConnect.tsx
```diff
@@ -174,7 +174,7 @@ class _LedgerConnect extends React.Component<Props, State> {
         refreshing: false,
         waiting: true,
       })
-      await onConnectBLE(device.id.toString())
+      onConnectBLE(device.id.toString())
```
michalszorad: `onConnectBLE` is an async function in `ConfirmTxWithHW.tsx`, will this not cause any issues? Eg the error handling below or animation (loading state). I remember we had a UI bug where we did not await an async function and this looks very similar. Maybe we should change the type of `onConnectBLE` to `onConnectBLE: (deviceId: DeviceId) => Promise<void> | void` and then `await onConnectBLE(device.id.toString()) `?

The same applies to `onConnectUSB`

src/yoroi-wallets/cardano/ShelleyWallet.ts
```diff
@@ -432,14 +432,14 @@ export class ShelleyWallet implements WalletInterface {
...
+  private getAddressedChangeAddress(): Promise<{address: string; addressing: CardanoTypes.Addressing}> {
     const changeAddr = this.getChangeAddress()
     const addressing = this.getAddressing(changeAddr)
 
-    return {
+    return Promise.resolve({
       address: changeAddr,
       addressing,
-    }
+    })
```
michalszorad, line 442: This function probably does even need to be async / return Promise

wolverineks: im not worried about this right now. im focusing on not changing behavior and we can fix this stuff later

michalszorad (approved):

src/yoroi-wallets/cardano/ShelleyWallet.ts
```diff
@@ -432,14 +432,14 @@ export class ShelleyWallet implements WalletInterface {
     return changeAddress
   }
 
-  private async getAddressedChangeAddress(): Promise<{address: string; addressing: CardanoTypes.Addressing}> {
+  private getAddressedChangeAddress(): Promise<{address: string; addressing: CardanoTypes.Addressing}> {
     const changeAddr = this.getChangeAddress()
     const addressing = this.getAddressing(changeAddr)
```
stackchain, line 437: `getChangeAddress` and `getAddressing` must not throw, it will need to be wrapped within a promise, otherwise `.catch()` won't catch.

src/yoroi-wallets/cardano/ShelleyWallet.ts
```diff
@@ -535,7 +535,7 @@ export class ShelleyWallet implements WalletInterface {
     throw new Error(`Missing address info for: ${address} `)
   }
 
-  async getAddressedUtxos() {
+  getAddressedUtxos() {
     const addressedUtxos = this.utxos.map((utxo: RawUtxo): CardanoTypes.CardanoAddressedUtxo => {
```
stackchain, line 539: same :arrow_up:

src/yoroi-wallets/cardano/ShelleyWallet.ts
wolverineks: if there were any `.catch`'s, typescript would have errored?

src/yoroi-wallets/cardano/ShelleyWallet.ts
wolverineks: if there were any .catch's, typescript would have errored?

banklesss (approved):

PR #2306, opened 2023-01-18
Memo storage

banklesss:
Related to [YOMO-309](https://emurgo.atlassian.net/browse/YOMO-309)

[YOMO-309]: https://emurgo.atlassian.net/browse/YOMO-309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/yoroi-wallets/cardano/shelley/transactionCache.ts
```diff
@@ -505,6 +518,22 @@ export const makeTxCacheStorage = (storage: Storage): TxCacheStorage => ({
   },
 })
 
+const addMemos = async (transactions: Record<string, Transaction>, memosStorage: Storage) => {
+  const memosTxIds = await memosStorage.getAllKeys()
+
+  memosTxIds.forEach(async (memoTxId) => {
+    const memo = (await memosStorage.getItem(memoTxId)) as string
```
michalszorad: `array.forEach` won't work with `async` and the `addMemos` may resolve before all the promises inside the function resolve. `const promises = memosTxIds.map( async (memotxId)) => ...` and then `await Promise.all(promises)` should work or a `for` loop should work too

src/yoroi-wallets/cardano/ShelleyWallet.ts
```diff
@@ -387,10 +389,14 @@ export class ShelleyWallet implements WalletInterface {
   }
 
   async clear() {
-    await this.transactionCache?.clear()
+    await this.transactionManager.clear()
```
michalszorad, line 414: Is the `?.` needed here?

src/yoroi-wallets/cardano/shelley/transactionCache.ts
```diff
@@ -505,6 +518,22 @@ export const makeTxCacheStorage = (storage: Storage): TxCacheStorage => ({
...
+    transactions = Object.assign(transactions, {
+      [memoTxId]: {
+        ...transactions[memoTxId],
+        memo,
+      },
+    })
+  })
```
michalszorad: `Object.assign` modifies the object in the first param, so I would suggest using `{...transactions, [memoTxId]: ...}` or `Object.assign({}, transactions, {...})`

src/yoroi-wallets/cardano/ShelleyWallet.ts
banklesss: The "optional chaining" operator was added previously because `transactionCache` was initialized in the method `initialize`.

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,34 @@
...
+import {TransactionCache} from './shelley'
+
+export const makeTransactionManager = async (storage: Storage) => {
+  const memosStorage = storage.join('memos/')
+  const transactionCache = await TransactionCache.create(storage.join('txs/'), memosStorage)
+
+  return {
+    // transactionCache api
+    get transactions() {
+      return transactionCache.transactions
```
stackchain: I'd remove all changes in the tx cache, the memo is wallet info, and is just linked to tx, why don't modify only the getter to combine the data, when present, here?

banklesss: Getting the memos is an async operation, that's the reason I'm adding in advance the memos.

Do you have any suggestion?

banklesss: I can convert the getter to an asynchronous function but then I would have to make modifications in the parts where it is being consumed. If you have a different idea please let me know

wolverineks: you could load all the memos into memory and merge on the fly

wolverineks: or make a new wallet method `await wallet.getTransaction(txid)` that wraps `wallet.transactionManager.getTransaction(txid)`

wolverineks: it looks like `getTransactionInfos` and `wallet.getTransactions` do basically the same things, could use some cleanup

banklesss: > you could load all the memos into memory and merge on the fly

@wolverineks At what point in the wallet cycle would you load it?

> or make a new wallet method await wallet.getTransaction(txid) that wraps wallet.transactionManager.getTransaction(txid)

when would that feature be useful? could you extend this idea? Thanks

banklesss: Maybe one idea is to create a method to load the memos into memory and subscribe this method to `transactionCache` `transaction` event

wolverineks: make transaction manager?

banklesss: @wolverineks any name suggestion?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  await updateMemos()
+
+  const saveMemo = async (txId: string, memo: string): Promise<void> => {
+    await storage.setItem(txId, memo)
+    await updateMemos()
+  }
+
+  return {
+    get memos() {
```
wolverineks: why is this a getter?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,64 @@
...
+    saveMemo: (txId: string, memo: string): Promise<void> => memosManager.saveMemo(txId, memo),
+  } as const
+}
+
+export type TransactionManager = Awaited<ReturnType<typeof makeTransactionManager>>
+
+
+const makeMemosManager = async (storage: Storage) => {
+  const getMemos = async () => (await storage.getAllKeys().then(storage.multiGet)) as unknown as Array<[string, string]>
+  let memos: Array<[string, string]> = []
```
wolverineks: why array of tuples and not a `{ txid: memo }`?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,64 @@
...
+    // memo api
+    saveMemo: (txId: string, memo: string): Promise<void> => memosManager.saveMemo(txId, memo),
+  } as const
+}
+
+export type TransactionManager = Awaited<ReturnType<typeof makeTransactionManager>>
+
+
+const makeMemosManager = async (storage: Storage) => {
+  const getMemos = async () => (await storage.getAllKeys().then(storage.multiGet)) as unknown as Array<[string, string]>
```
wolverineks: parse, dont cast

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,64 @@
+import {Storage} from '../storage'
+import {BackendConfig} from '../types'
+import {TransactionCache} from './shelley'
+
+export const makeTransactionManager = async (storage: Storage) => {
+  const transactionCache = await TransactionCache.create(storage.join('txs/'))
+  const memosManager = await makeMemosManager(storage.join('memos/'))
+  await memosManager.updateMemos()
```
wolverineks: isnt this already taken care of in `makeMemosManager`?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  return {
+    // transactionCache api
+    get transactions() {
+      const txs = {...transactionCache.transactions}
+      memosManager.memos.forEach(([address, memo]) => {
+        txs[address].memo = memo
```
wolverineks: dont mutate

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: `memos = await getMemos()`
and get rid of `await updateMemos`
?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,64 @@
...
+  const saveMemo = async (txId: string, memo: string): Promise<void> => {
+    await storage.setItem(txId, memo)
+    await updateMemos()
+  }
+
+  return {
+    get memos() {
+      return memos
+    },
+    updateMemos,
```
wolverineks: is there a need to expose this?

storybook/mocks/wallet.ts
```diff
@@ -171,6 +171,13 @@ const wallet: YoroiWallet = {
   save: async (...args) => {
     action('save')(...args)
   },
+  saveMemo: async (...args) => {
+    action('saveMemo')(...args)
+  },
+  readMemo: async (...args) => {
```
wolverineks: is this needed anymore?

banklesss: Removed

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: no

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: 👍🏻

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: to be able to provide the updated memos. Tested locally.

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -11,11 +11,11 @@ export const makeTransactionManager = async (storage: Storage) => {
     get transactions() {
       const {memos} = memosManager
       return Object.keys(memos).reduce(
-        (allTransactions, currentMemoAddress) => ({
+        (allTransactions, currentMemoTxAddress) => ({
```
wolverineks: (result, current)
start with an empty {}

wolverineks: iterate over that transactions instead of the memos

banklesss: could you explain the reason? I iterate over the memos because this is the minimum number of iterations

wolverineks: don't bother considering performance optimizations. write code with the clearest intent.
`transactions -> transactions with memos`

michalszorad (dismissed):

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: no event subscriptions
load them into memory when creating the transaction manager

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,62 @@
...
+import {BackendConfig} from '../types'
+import {TransactionCache} from './shelley'
+
+export const makeTransactionManager = async (storage: Storage) => {
+  const transactionCache = await TransactionCache.create(storage.join('txs/'))
+  const memosManager = await makeMemosManager(storage.join('memos/'))
+
+  return {
+    // transactionCache api
+    get transactions() {
```
stackchain: do we need these `get`s? I'm wondering why TS is not complaining

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,62 @@
...
+  const getMemos = async () => storage.getAllKeys().then(storage.multiGet).then(Object.fromEntries)
+  let memos = await getMemos()
+
+  const saveMemo = async (txId: string, memo: string): Promise<void> => {
+    await storage.setItem(txId, memo)
+    memos = await getMemos()
+  }
+
+  return {
+    get memos() {
```
stackchain: when deleting the wallet the memos must be gone altogether.

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: `memo` getter provides the memos updated, if it is not a getter it just provides the memos gotten when the instance was created

In the case of `transactions ` Maybe we can avoid it by refactoring

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,62 @@
...
+    get perAddressTxs() {
+      return transactionCache.perAddressTxs
+    },
+    get confirmationCounts() {
+      return transactionCache.confirmationCounts
+    },
+    clear: () => transactionCache.clear(),
+    resetState: () => transactionCache.resetState(),
+    subscribe: (handler) => transactionCache.subscribe(handler),
+    doSync: async (addressesByChunks: Array<Array<string>>, backendConfig: BackendConfig) =>
```
stackchain: can we move the `backendConfig` to the initialization?

stackchain (commented):
⬆️

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: I have checked it I have noticed that the transaction manager instance is created and passed as an argument for the shelley wallet constructor

Do you think should the transaction manager instance be created after the shelley wallet instance is created?

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: Should i refactor it ? thanks!

stackchain: I think you can just remove all `get` and add `as const` in the return.

src/yoroi-wallets/cardano/transactionManager.ts
stackchain: I think this should be part of `commonCreate()` as UtxoManager is.

stackchain: backend config is synchronously available there, so we don't need to provide it every time.

banklesss: Correct me if I am wrong! but `commonCreate` is a static method used in static methods and `getBackendConfig` is a private method of `ShelleyWallet` so it is not possible to get the backend config unless its obtained independently of `getBackendConfig`

src/yoroi-wallets/cardano/ShelleyWallet.test.ts
```diff
@@ -103,10 +103,10 @@ describe('migration', () => {
       `"stake1u8q3auyvgnekzzm72m2xuzrtjqvxcyhf568s2gdhcnrjujcr25kkv"`,
     )
 
-    expect(wallet.transactionCache?.perRewardAddressCertificates).toMatchInlineSnapshot(`Object {}`)
-    expect(wallet.transactionCache?.confirmationCounts).toMatchInlineSnapshot(`Object {}`)
-    expect(wallet.transactionCache?.perAddressTxs).toMatchInlineSnapshot(`Object {}`)
-    expect(wallet.transactionCache?.transactions).toMatchInlineSnapshot(`Object {}`)
+    expect(wallet.transactionManager?.getPerRewardAddressCertificates()).toMatchInlineSnapshot(`Object {}`)
```
wolverineks: does this still need to be `?`?

src/yoroi-wallets/cardano/ShelleyWallet.ts
```diff
@@ -274,7 +276,7 @@ export class ShelleyWallet implements WalletInterface {
     const rewardAddressHex = await deriveRewardAddressHex(accountPubKeyHex, networkId)
     const apiUrl = getCardanoNetworkConfigById(networkId).BACKEND.API_ROOT
     const utxoManager = await makeUtxoManager({storage: storage.join('utxoManager/'), apiUrl})
-    const transactionCache = await TransactionCache.create(storage.join('txs/'))
+    const transactionManager = await makeTransactionManager(storage.join('tx-manager/'))
```
wolverineks: follow conventions, `'txManager'`

src/yoroi-wallets/cardano/ShelleyWallet.ts
```diff
@@ -386,11 +388,24 @@ export class ShelleyWallet implements WalletInterface {
...
-    await this.transactionCache?.clear()
+    await this.transactionManager.clear()
     await this.utxoManager.clear()
   }
 
+  saveMemo(txId: string, memo: string): Promise<void> {
+    return this.transactionManager.saveMemo(txId, memo)
+  }
+
+  clearMemos(): Promise<void> {
```
wolverineks: ~what is this used for?~

banklesss: To remove the memos when a wallet is removed

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,17 @@
+import {storage as rootStorage} from '../storage'
+import {makeMemosManager} from './transactionManager'
+
+describe('transaction manager', () => {
```
wolverineks: this is not a transaction manager test

src/yoroi-wallets/cardano/ShelleyWallet.ts
wolverineks: move this into `remove()`
updates types

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,66 @@
...
+export const makeMemosManager = async (storage: Storage) => {
+  const getMemos = async () => storage.getAllKeys().then(storage.multiGet).then(Object.fromEntries)
+  let memos = await getMemos()
+
+  const saveMemo = async (txId: string, memo: string): Promise<void> => {
+    await storage.setItem(txId, memo)
+    memos = await getMemos()
+  }
+
+  const clear = async () => storage.getAllKeys().then(storage.multiRemove)
```
wolverineks: `storage.clear()`?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,66 @@
...
+    // memo api
+    saveMemo: (txId: string, memo: string): Promise<void> => memosManager.saveMemo(txId, memo),
+    clearMemos: (): Promise<void> => memosManager.clear(),
+  } as const
+}
+
+export type TransactionManager = Awaited<ReturnType<typeof makeTransactionManager>>
+
+export const makeMemosManager = async (storage: Storage) => {
+  const getMemos = async () => storage.getAllKeys().then(storage.multiGet).then(Object.fromEntries)
```
wolverineks: nice

wolverineks:
```suggestion
  const getMemos = () => storage.getAllKeys().then(storage.multiGet).then(Object.fromEntries)
```

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks:
```suggestion
  const clear = () => storage.getAllKeys().then(storage.multiRemove)
```

src/yoroi-wallets/cardano/types.ts
```diff
@@ -100,8 +100,14 @@ export interface WalletInterface {
 
   save(): Promise<void>
 
+  remove(): Promise<void>
```
wolverineks: whats the difference between `remove()` and `clear()`?

banklesss: When the wallet is removed from `walletManager` the wallet is cleared and the memos are removed.

On resync the wallet is cleared but the memos are not removed.

So I think we have to make a distinction here.

Also maybe is a good starting point to move all the storage removing stuff from `walletManager.removeWallet` to the wallet itself, is it ok?

wolverineks: eventually, not this pr

src/yoroi-wallets/cardano/transactionManager.ts
stackchain: there is a version, `getNetworkConfigById`, I believe they return the same thing based on networkId

banklesss: you are right ! 👍🏻

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+      return Object.keys(transactionCache.transactions).reduce(
+        (result, current) => ({
+          ...result,
+          [current]: {
+            ...transactionCache.transactions[current],
+            memo: memos[current],
+          },
+        }),
+        {},
+      )
```
wolverineks: `Object.entries`

wolverineks: and make sure its properly typed

src/yoroi-wallets/cardano/shelley/transactionCache.ts
```diff
@@ -398,7 +406,7 @@ export type TimestampedCertMeta = {
   epoch: number
   certificates: Array<RemoteCertificateMeta>
 }
-type PerAddressCertificatesDict = Record<string, Record<string, TimestampedCertMeta>>
+export type PerAddressCertificatesDict = Record<string, Record<string, TimestampedCertMeta>>
```
wolverineks:
```suggestion
type PerAddressCertificatesDict = Record<string, Record<string, TimestampedCertMeta>>
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+import {BackendConfig, Transactions} from '../types'
+import {PerAddressCertificatesDict, TransactionCache} from './shelley'
+
+export const makeTransactionManager = async (storage: Storage, backendConfig: BackendConfig) => {
+  const transactionCache = await TransactionCache.create(storage.join('txs/'))
+  const memosManager = await makeMemosManager(storage.join('memos/'))
+
+  return {
+    // transactionCache api
+    getTransactions(): Transactions {
```
wolverineks:
```suggestion
    getTransactions() {
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+      const memos = memosManager.getMemos()
+      return Object.entries(transactionCache.transactions).reduce(
+        (result, [txId, tx]) => ({
+          ...result,
+          [txId]: {
+            ...tx,
+            memo: memos[txId],
+          },
+        }),
+        {},
```
wolverineks:
```suggestion
        {} as Transactions,
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+      )
+    },
+    getPerRewardAddressCertificates(): PerAddressCertificatesDict {
```
wolverineks:
```suggestion
    getPerRewardAddressCertificates() {
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+      return transactionCache.perRewardAddressCertificates
+    },
+    getPerAddressTxs(): Record<string, string[]> {
```
wolverineks:
```suggestion
    getPerAddressTxs() {
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+      return transactionCache.perAddressTxs
+    },
+    getConfirmationCounts(): {[x: string]: number | null} {
```
wolverineks:
```suggestion
    getConfirmationCounts() {
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+      return transactionCache.confirmationCounts
+    },
+    clear: (): Promise<void> => transactionCache.clear(),
```
wolverineks:
```suggestion
    clear: () => transactionCache.clear(),
```

banklesss:
thanks @wolverineks

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+      return transactionCache.confirmationCounts
+    },
+    clear: () => transactionCache.clear(),
+    resetState: () => transactionCache.resetState(),
+    subscribe: (handler) => transactionCache.subscribe(handler),
+    doSync: async (addressesByChunks: Array<Array<string>>) =>
+      transactionCache.doSync(addressesByChunks, backendConfig),
+
+    // memo api
+    saveMemo: (txId: string, memo: string): Promise<void> => memosManager.saveMemo(txId, memo),
```
wolverineks:
```suggestion
    saveMemo: (txId: string, memo: string) => memosManager.saveMemo(txId, memo),
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+    clearMemos: (): Promise<void> => memosManager.clear(),
```
wolverineks:
```suggestion
    clearMemos: () => memosManager.clear(),
```

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: how did you test this?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+    memos = await getMemos()
+  }
+
+  const clear = async () => {
+    await storage.getAllKeys().then(storage.multiRemove)
+    memos = {}
+  }
+
+  return {
+    getMemos() {
```
wolverineks: if you mutate memos, then i think this wouldnt have to be a getter
something like
```
memos = Object.assign(memos, await getMemos())
```

wolverineks:
```
Object.keys(memos).forEach(key => delete obj[key])
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+          ...result,
+          [txId]: {
+            ...tx,
+            memo: memos[txId] ?? null,
+          },
+        }),
+        {} as Transactions,
+      )
+    },
+    getPerRewardAddressCertificates() {
```
wolverineks: why are these getters?

banklesss: How to expose these values?

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: why not use a getter?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+export type TransactionManager = Awaited<ReturnType<typeof makeTransactionManager>>
+
+export const makeMemosManager = async (storage: Storage) => {
+  const getMemos = () => storage.getAllKeys().then(storage.multiGet).then(Object.fromEntries) ?? {}
+  let memos = await getMemos()
+
+  const saveMemo = async (txId: string, memo: string): Promise<void> => {
+    await storage.setItem(txId, memo)
+    memos = await getMemos()
+  }
```
stackchain: switch to optimistic update when saving, that we don't have to wait for the storage, make it readonly to avoid mutations

```suggestion
export const makeMemosManager = async (storage: Storage) => {
  const getMemos = () => storage.getAllKeys().then(storage.multiGet).then(Object.fromEntries)
  let memos: Readonly<....> = await getMemos()

  const updateMemos = (txId, memo) => memos = {...memos, {[txId]: memo}}  

  const saveMemo = async (txId: string, memo: string): Promise<void> => {
    updateMemos(txId, memo) 
    storage.setItem(txId, memo)
  }
```

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  const clear = async () => {
+    await storage.getAllKeys().then(storage.multiRemove)
+    memos = {}
```
stackchain:
```suggestion
    memos = {}
    storage.getAllKeys().then(storage.multiRemove)

```

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: ?
```
perAddressTxs: transactionCache.perAddressTxs
```

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: if it works, it works

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: premature performance optimization

wolverineks: plus there is no way to know if the memos have change until the resolve of the function anyway

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: premature performance optimization
there is no way to know if the memos have changed until the resolve of the function

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: So should i mutate in this case?

src/yoroi-wallets/cardano/transactionManager.ts
stackchain: where do we need to know if the memos have changed?

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: getter is fine

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: i would like to test this before pushing. Maybe it will happen like it happened here https://github.com/Emurgo/yoroi-mobile/pull/2306#discussion_r1090826052

banklesss: some testnet apis are down

banklesss: It needs to be a getter to be able to get the updated values. Tested locally using a similar approach as in this branch https://github.com/Emurgo/yoroi-mobile/commit/2ac2021cd0d07f8fe4ec1668dc77188fd82f3181

cc @stackchain @wolverineks

stackchain (approved):
not sure why need to wrap the getters to make it work (need to test it), lgtm

michalszorad (approved):

src/yoroi-wallets/cardano/transactionManager.ts
wolverineks: are you going to change this?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,78 @@
...
+import {BackendConfig, Transactions} from '../types'
+import {TransactionCache} from './shelley'
+
+export const makeTransactionManager = async (storage: YoroiStorage, backendConfig: BackendConfig) => {
+  const transactionCache = await TransactionCache.create(storage.join('txs/'))
+  const memosManager = await makeMemosManager(storage.join('memos/'))
+
+  return {
+    // transactionCache api
+    getTransactions() {
```
wolverineks: are you going to change this?

src/yoroi-wallets/cardano/transactionManager.ts
```diff
@@ -0,0 +1,78 @@
...
+        }),
+        {} as Transactions,
+      )
+    },
+    getPerRewardAddressCertificates() {
+      // to get the updates
+      return transactionCache.perRewardAddressCertificates
+    },
+    getPerAddressTxs() {
+      // to get the updates
```
wolverineks: are you going to these?

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,23 @@
+import {storage as rootStorage} from '../storage'
+import {makeMemosManager} from './transactionManager'
+
```
wolverineks, line 61: where are the tx manager tests?

wolverineks (changes requested):
comments

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: I tested that https://github.com/Emurgo/yoroi-mobile/pull/2306#discussion_r1091809650

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: I tested that https://github.com/Emurgo/yoroi-mobile/pull/2306#discussion_r1091809650

src/yoroi-wallets/cardano/transactionManager.ts
banklesss: I tested that https://github.com/Emurgo/yoroi-mobile/pull/2306#discussion_r1091809650

banklesss:
> not sure why need to wrap the getters to make it work (need to test it), lgtm

 I tested that https://github.com/Emurgo/yoroi-mobile/pull/2306#discussion_r1091809650

I there is any other alternative please let me know and I would be happy to apply it

@stackchain @wolverineks @MichalSzorad

wolverineks:
> 

why not include the tests on the pr?

jemubm:
> > 
> 
> why not include the tests on the pr?

banklesss:
> why not include the tests on the pr?

Test related to the getters ? I will do

banklesss:
Tests updated.

In this branch the tests do not pass if we do without the getters: https://github.com/Emurgo/yoroi-mobile/commit/fe06f8ff2f8164511e5a6ab2aa616222ffda089e

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,148 @@
...
+  it('stores memos', async () => {
+    expect(txManager.getTransactions()).toEqual({[mockTx.id]: mockTx})
+
+    await txManager.saveMemo(mockTx.id, 'memo 1')
+
+    expect(txManager.getTransactions()).toEqual({[mockTx.id]: {...mockTx, memo: 'memo 1'}})
+  })
+
+  it('gets updated tx cache values', async () => {
```
banklesss: Testing that the getters are tracking the changes in tx cache state.

Through doing something like `perAddressTxs: transactionCache.perAddressTxs` the changes aren't tracked as can be verified by making the appropriate modifications in the tests and in tx manager

src/yoroi-wallets/cardano/shelley/__snapshots__/transactionCache.test.ts.snap
```diff
@@ -13,7 +13,12 @@ Object {
   "0a8962dde362eef1f840defe6f916fdf9701ad53c7cb5dd4a74ab85df8e9bffc": Object {
     "blockHash": "fb418acaa29c66e799a16b594f7beedfe2ef53413e9863b61a418f2df1ff1442",
     "blockNum": 2909238,
-    "certificates": Array [],
+    "certificates": Array [
```
wolverineks: why did this change?

src/yoroi-wallets/cardano/shelley/__snapshots__/transactionCache.test.ts.snap
```diff
@@ -111,7 +117,22 @@ Object {
 }
 `;
 
-exports[`transactionCache create loads from storage 3`] = `Object {}`;
+exports[`transactionCache create loads from storage 3`] = `
+Object {
```
wolverineks: why did this change?

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,148 @@
...
+import {makeMemosManager, makeTransactionManager} from './transactionManager'
+
+jest.mock('./api', () => mockApi)
+
+describe('transaction manager', () => {
+  DeviceInfo.getVersion = () => '9.9.9'
+
+  let txManager
+  beforeEach(async () => {
+    txManager = await makeTransactionManager(mockStorage, mockedBackendConfig)
```
wolverineks: if you dont have to share the txManager, then create a new one inside each test

src/yoroi-wallets/cardano/shelley/__snapshots__/transactionCache.test.ts.snap
banklesss: To test the changes in the variable `perRewardAddressCertificates`

src/yoroi-wallets/cardano/shelley/__snapshots__/transactionCache.test.ts.snap
banklesss: To test the changes in the variable `perRewardAddressCertificates`

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    })
+
+    // tx cache mutation
+    await txManager.doSync(mockedAddressesByChunks)
+
+    expect(txManager.getTransactions()).toEqual({
+      [mockTx.id]: mockTx,
+      ...fromPairs(mockedHistoryResponse.transactions.map((t) => [t.hash, toCachedTx(t)])),
+    })
+    expect(txManager.getPerAddressTxs()).toEqual(
```
wolverineks: use `toMatchSnapshot` for complex assertions

wolverineks: or `toEqual` with a mock defined somewhere else

src/yoroi-wallets/cardano/transactionManager.test.ts
wolverineks: the tests should resemble how you intend the class to be used
if you dont intend for users of this class to use selectors, dont use them in the tests

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        },
+      },
+    })
+
+    // tx cache mutation
+    await txManager.doSync(mockedAddressesByChunks)
+
+    expect(txManager.getTransactions()).toEqual({
+      [mockTx.id]: mockTx,
+      ...fromPairs(mockedHistoryResponse.transactions.map((t) => [t.hash, toCachedTx(t)])),
```
wolverineks: too much logic. we want to avoid needed tests for our tests

banklesss: copied from` transactionCache.test.ts`. Should it be changed there too?

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    await memosManager.clear()
+
+    expect(memosManager.getMemos()).toEqual({})
+  })
+})
+
+// mocks
+
+const mockStorage: YoroiStorage = {
```
wolverineks: can you create a different storage for each test, and seed it with whatever data it needs?
something like:
```
storage.join("txs/").setItem("txids", [mockTx.id])
storage.join("txs/").multiSet([[txids[0], mockTx])
```

banklesss: `transactionCache` and `memosManager` storages derive from the same storage

src/yoroi-wallets/cardano/shelley/__snapshots__/transactionCache.test.ts.snap
banklesss: Also, `transactionCache.test.ts` and `transactionManager.test.ts` are using the same mocks now.

src/yoroi-wallets/cardano/shelley/__snapshots__/transactionCache.test.ts.snap
banklesss: Also, `transactionCache.test.ts` and `transactionManager.test.ts` are using the same mocks now.

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: it is no longer necessary. using snapshots

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: it is no longer necessary. using snapshots.

src/yoroi-wallets/cardano/shelley/transactionCache.ts
```diff
@@ -375,7 +383,7 @@ type TimeForTx = {
   txOrdinal: number
 }
 
-const perAddressTxsSelector = (state: TransactionCacheState) => {
+export const perAddressTxsSelector = (state: TransactionCacheState) => {
```
banklesss:
```suggestion
const perAddressTxsSelector = (state: TransactionCacheState) => {
```

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: And the root storage is an argument of `makeTransactionManager`. Once `transactionManager` instantiates there is no control over it.

wolverineks: can you do this?
```
const txStorage = storage.join("txs/")
txStorage.setItem(...)
txStorage.multiSet(...)
// ...whatever setup....

const transactionManager = makeTransactionManager(txStorage)
```
so each test has a different txManager and each storage is customized directly in the test

src/yoroi-wallets/cardano/mocks/index.ts
```diff
@@ -0,0 +1,815 @@
...
+    withdrawals: [],
+    certificates: [],
+    validContract: true,
+    scriptSize: 0,
+    collateralInputs: [],
+    memo: null,
+  },
+}
+
+export const mockApi = {
```
wolverineks, line 808: delete mockApi, 
create the mock in the test

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,86 @@
+import DeviceInfo from 'react-native-device-info'
+
+import {storage as rootStorage, YoroiStorage} from '../storage'
+import {Transaction} from '../types'
+import {mockApi, mockedAddressesByChunks, mockedBackendConfig, mockTx} from './mocks'
+import {makeMemosManager, makeTransactionManager} from './transactionManager'
+
+jest.mock('./api', () => mockApi)
```
wolverineks: jest.mock('./api', () => ({
  ...create mock here...
})

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+describe('transaction manager', () => {
+  DeviceInfo.getVersion = () => '9.9.9'
+
+  afterEach(rootStorage.clear)
```
wolverineks: beforeEach(() => AsyncStorage.clear())

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: It's not possible to access out-of-scope variables from there.

banklesss: So the mocked api responses can't be imported from the common mocks. Test it if you have time.

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    await txManager.saveMemo(mockTx.id, 'memo 1')
+
+    expect(txManager.getTransactions()).toEqual({[mockTx.id]: {...mockTx, memo: 'memo 1'}})
+  })
+
+  it('gets updated tx cache values', async () => {
+    const txManager = await makeTransactionManager(mockStorage, mockedBackendConfig)
+
+    expect(txManager.getTransactions()).toMatchSnapshot()
```
wolverineks: can you pick something from each return value to check specifically?
```
expect(transactions.length).toBe(1)
await txManager.doSync(mockedAddressesByChunks)
expect(transactions.length).toBe(2)
```

banklesss: Should the snapshots be removed?

banklesss: Or should that be added as an extra check?

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: The storage needs to be cleared after the last test because it's going to be used in the memo manager tests.

banklesss: Maybe clearing the storage before starting the memos manager tests its fine

src/yoroi-wallets/cardano/mocks/index.ts
banklesss: It is not posible to use out-of-scope variables inside the mock factory function so I would need to add everything from there in the factory function.

banklesss: The tx history mock is a big one mock

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: The tx history mock is very big!

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: Using `afterAll` should be fine too

src/yoroi-wallets/cardano/transactionManager.test.ts
wolverineks: import the responses and build the mock here

src/yoroi-wallets/cardano/transactionManager.test.ts
wolverineks: clear the storage before each of the memo tests too

wolverineks: I prefer beforeEach because it's defensive and protects the tests in case someone else forgot to clear

src/yoroi-wallets/cardano/transactionManager.test.ts
wolverineks: I'm fine either way, just make the tests read well

src/yoroi-wallets/cardano/transactionManager.test.ts
wolverineks: it worked for me

banklesss: yes it works thanks!

But when you define the variables in the file it doesn't work https://github.com/Emurgo/yoroi-mobile/commit/91a2fa1e29081b6fe53256068391af55bced3a68

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: I cant get to model this to work yet.  I need to test it further anyway.

stackchain (approved):
lgtm

src/yoroi-wallets/cardano/ShelleyWallet.ts
```diff
@@ -264,7 +263,10 @@ export class ShelleyWallet implements WalletInterface {
     const rewardAddressHex = await deriveRewardAddressHex(accountPubKeyHex, networkId)
     const apiUrl = getCardanoNetworkConfigById(networkId).BACKEND.API_ROOT
     const utxoManager = await makeUtxoManager({storage: storage.join('utxoManager/'), apiUrl})
-    const transactionCache = await TransactionCache.create(storage.join('txs/'))
+    const transactionManager = await makeTransactionManager(
+      storage.join('txManager/'),
+      getCardanoNetworkConfigById(networkId).BACKEND,
```
wolverineks: why move this here?

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: @wolverineks tested. And its not possible its implementation because its creating a path like: `.../join/join`

Let me know if you have another idea to implement this.

src/yoroi-wallets/cardano/ShelleyWallet.ts
banklesss: Discussed here: https://github.com/Emurgo/yoroi-mobile/pull/2306#discussion_r1088177239

michalszorad (approved):

src/yoroi-wallets/cardano/ShelleyWallet.test.ts
```diff
@@ -101,10 +101,10 @@ describe('migration', () => {
       `"stake1u8q3auyvgnekzzm72m2xuzrtjqvxcyhf568s2gdhcnrjujcr25kkv"`,
     )
 
-    expect(wallet.transactionCache?.perRewardAddressCertificates).toMatchInlineSnapshot(`Object {}`)
-    expect(wallet.transactionCache?.confirmationCounts).toMatchInlineSnapshot(`Object {}`)
-    expect(wallet.transactionCache?.perAddressTxs).toMatchInlineSnapshot(`Object {}`)
-    expect(wallet.transactionCache?.transactions).toMatchInlineSnapshot(`Object {}`)
+    expect(wallet.transactionManager.perRewardAddressCertificates).toMatchInlineSnapshot(`Object {}`)
```
wolverineks: these were migration tests
you can get rid of anything testing private wallet values

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    .fn()
+    .mockResolvedValueOnce(mockedHistoryResponse)
+    .mockResolvedValueOnce(mockedEmptyHistoryResponse)
+    .mockResolvedValueOnce(mockedEmptyHistoryResponse),
+}))
+
+beforeEach(() => AsyncStorage.clear())
+
+describe('transactionManager', () => {
+  describe('create', () => {
```
wolverineks: you dont need "create"

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    it('loads from storage', async () => {
```
wolverineks: "restore, resetState"

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,303 @@
...
+      const txManager = await TransactionManager.create(mockStorage)
+
+      expect(txManager.transactions).toMatchSnapshot()
+      expect(txManager.perAddressTxs).toMatchSnapshot()
+      expect(txManager.perRewardAddressCertificates).toMatchSnapshot()
+      expect(txManager.confirmationCounts).toMatchSnapshot()
+
+      txManager.resetState()
+
+      expect(txManager.transactions).toMatchSnapshot()
```
wolverineks: if the snapshot is `{}`, then `.toEqual({})`

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,303 @@
...
+      expect(txManager.perAddressTxs).toMatchSnapshot()
+      expect(txManager.perRewardAddressCertificates).toMatchSnapshot()
+      expect(txManager.confirmationCounts).toMatchSnapshot()
+    })
+
+    it('syncs', async () => {
```
wolverineks: merge this into previous test

src/yoroi-wallets/cardano/transactionManager.test.ts
wolverineks: "restore, resetState, sync"

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,303 @@
...
+
+      const txManager = await TransactionManager.create(mockStorage)
+
+      expect(txManager.transactions).toMatchSnapshot()
+      expect(txManager.perAddressTxs).toMatchSnapshot()
+      expect(txManager.perRewardAddressCertificates).toMatchSnapshot()
+      expect(txManager.confirmationCounts).toMatchSnapshot()
+    })
+
+    it('drops transaction if invalid format', async () => {
```
wolverineks: this is a txstorage test

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: do you mean this test can be removed?

src/yoroi-wallets/cardano/transactionManager.test.ts
```diff
@@ -0,0 +1,303 @@
...
+      expect(txManager.perRewardAddressCertificates).toMatchSnapshot()
+
+      await txManager.doSync(mockedAddressesByChunks, mockedBackendConfig)
+
+      expect(txManager.transactions).toMatchSnapshot()
+      expect(txManager.perAddressTxs).toMatchSnapshot()
+      expect(txManager.perRewardAddressCertificates).toMatchSnapshot()
+    })
+
+    it('starts fresh if txids is invalid format', async () => {
```
wolverineks: this is a txstorage test

src/yoroi-wallets/cardano/transactionManager.test.ts
banklesss: ok i got it

wolverineks (changes requested):
comments

wolverineks (approved):
comments

PR #2309, opened 2023-01-23
feat(search): Add context to search

michalszorad:
https://emurgo.atlassian.net/browse/YOMO-454


banklesss (approved):

src/App.tsx
```diff
@@ -27,15 +28,17 @@ const App = () => {
...
-      <RNP.Provider>
-        <AuthProvider>
-          <SelectedWalletMetaProvider>
-            <SelectedWalletProvider>
-              <AppNavigator />
-            </SelectedWalletProvider>
-          </SelectedWalletMetaProvider>
-        </AuthProvider>
-      </RNP.Provider>
+      <SearchProvider>
```
wolverineks: move search provider as close to consumer as possible

src/Nfts/Nfts.tsx
```diff
@@ -8,15 +8,13 @@ import noNftsImage from '../assets/img/no-nft.png'
...
 import {ImageGallery, SkeletonGallery} from './ImageGallery'
 
-type Props = {
-  search?: string
-}
-
-export const Nfts = ({search}: Props) => {
-  const searchTermLowerCase = (search ?? '').toLowerCase()
+export const Nfts = () => {
+  const {search} = useSearch()
```
wolverineks: search is a verb, not a noun

wolverineks: this is filter, not a search

wolverineks: maybe something like?
```
const {results, setFilter} = useFilter(items)
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,27 @@
...
+    throw new Error('useSearch must be used within a SearchProvider')
+  }
+  return value
+}
+
+export const SearchProvider = ({children, value}: {children: ReactNode; value?: Partial<SearchContextApi>}) => {
+  const [search, setSearch] = useState('')
+  const [visible, setVisible] = useState(false)
+  const clearSearch = () => setSearch('')
+  const defaultValue = {search, setSearch, clearSearch, visible, setVisible}
```
wolverineks: what is default value?

wolverineks: is visibility toggled from anywhere except the header?
if not, would it make sense to move visibility out of the provider and into the header?

wolverineks: or maybe into the navigator file

src/Nfts/Nfts.tsx
```diff
@@ -8,15 +8,13 @@ import noNftsImage from '../assets/img/no-nft.png'
...
+  const searchTermLowerCase = search.toLowerCase()
   const wallet = useSelectedWallet()
   const {nfts, isLoading, refetch, isRefetching, isError} = useNfts(wallet)
   const navigation = useNavigation<WalletStackRouteNavigation>()
```
wolverineks: filtering should be done in the hook/context so that the end result it declarative
keep busy logic out of the ui, something like...
```
const filteredNfts = useFilteredNfts()
```

src/Nfts/NftsNavigator.tsx
```diff
@@ -36,29 +27,18 @@ export const NftsNavigator = () => {
...
-                <SearchBar
-                  placeholder={strings.search}
-                  onChangeText={(search) => setSearch(search)}
-                  value={search}
-                  onClearPress={handleClearPress}
-                  onBackPress={handleBackPress}
-                />
-              )
-            : undefined,
+          header: searchVisible ? () => <SearchHeader placeholder={strings.search} /> : undefined,
```
wolverineks: move `visible, setVisible` out of context and into this component

src/App.tsx
wolverineks: each search context should scoped, not reused
if we need more searching, then add multiple providers
searching on one screen shouldnt affect results on another screen

src/Nfts/NftsNavigator.tsx
wolverineks: `<SearchHeader onClear={() => setVisible(false)} ...`

wolverineks: ~how does the user show the search again, after its been cleared/hidden?~
nvm, got it

src/Nfts/Nfts.tsx
michalszorad: According to Google, "search" can be a verb and also a noun. The word filter could indicate there are also other than text search filters. Eg, filter by date minted, metadata, transaction volume, etc. In this case, I think it would be better to use `useSearch` or `useSearchFilter`, or `useTextSearch`. What do you think?

src/Search/SearchContext.tsx
michalszorad: Default value is an empty string, I'll make it clear in the code. I'll put the visibility toggle into navigator

michalszorad:
@wolverineks I've addressed the issues raised. Please let me know your thoughts

src/Nfts/Nfts.tsx
```diff
@@ -6,27 +6,16 @@ import {SafeAreaView} from 'react-native-safe-area-context'
...
-  const {nfts, isLoading, refetch, isRefetching, isError} = useNfts(wallet)
+export const Nfts = () => {
   const navigation = useNavigation<WalletStackRouteNavigation>()
-  const filteredNFTs =
-    searchTermLowerCase.length > 0 && nfts.length > 0
-      ? nfts.filter((n) => n.name.toLowerCase().includes(searchTermLowerCase))
-      : nfts
+  const {search, filteredNfts, isLoading, refetch, isRefetching, isError} = useFilteredNfts()
 
   const showDetails = (id: string) => navigation.navigate('nft-details-routes', {screen: 'nft-details', params: {id}})
```
stackchain:
```suggestion
  const navigateToDetails = (id: string) => navigation.navigate('nft-details-routes', {screen: 'nft-details', params: {id}})
```

src/Nfts/Nfts.tsx
```diff
@@ -6,27 +6,16 @@ import {SafeAreaView} from 'react-native-safe-area-context'
...
-  const handleNFTSelect = (index: number) => showDetails(nfts[index].id)
+  const handleNFTSelect = (index: number) => showDetails(filteredNfts[index].id)
```
stackchain:
```suggestion
  const handleNFTSelect = (index: number) => navigateToDetails(filteredNfts[index].id)
```

src/Nfts/Nfts.tsx
```diff
@@ -37,16 +26,16 @@ export const Nfts = ({search}: Props) => {
           {isError ? (
             <ErrorScreen onRefresh={refetch} isRefreshing={isRefetching} />
           ) : isLoading ? (
-            <LoadingScreen nftsCount={filteredNFTs.length} onRefresh={refetch} isRefreshing={isRefetching} />
-          ) : searchTermLowerCase.length > 0 && filteredNFTs.length === 0 ? (
+            <LoadingScreen nftsCount={filteredNfts.length} onRefresh={refetch} isRefreshing={isRefetching} />
```
stackchain: can we branch some of this logic earlier? I.e
```
if (isError) return (<View style={styles.root}><ErrorScreen....</View>)

if (isLoading) return (<View style={styles.root}><LoadingScreen....</View>)
```

src/Nfts/NftsNavigator.tsx
```diff
@@ -3,63 +3,47 @@ import React, {useState} from 'react'
...
       >
-        {(props) => <Nfts {...props} search={search} />}
-      </Stack.Screen>
-    </Stack.Navigator>
+        <Stack.Screen
+          name="nfts"
+          options={{
+            title: strings.title,
+            headerTitleAlign: 'center',
+            header: searchVisible
```
stackchain: I'd wrap it in a `ToogleabeSearchBar`, but by changing the state within the navigator, all stack screens will rerender, by moving the state deeper only its children will be affected.

src/hooks/index.ts
```diff
@@ -23,6 +23,8 @@ import {processTxHistoryData} from '../legacy/processTransactions'
 import {WalletMeta} from '../legacy/state'
 import storage from '../legacy/storage'
 import {cardanoValueFromRemoteFormat} from '../legacy/utils'
+import {useSearch} from '../Search'
+import {useSelectedWallet} from '../SelectedWallet'
```
wolverineks: `hooks.index.ts` will be extracted into another library so it can be shared with extension
so, no additions that are tied to mobile

src/hooks/index.ts
```diff
@@ -913,3 +915,15 @@ export const useNft = (wallet: YoroiWallet, {id}: {id?: string} = {}): YoroiNFT
   }
   return nft
 }
+
+export const useFilteredNfts = () => {
```
wolverineks: move this to nfts directory

src/Nfts/Nfts.tsx
michalszorad: Thanks, I've just merged feat/nft-gallery into this branch and that has fixed this issue

src/Nfts/Nfts.tsx
michalszorad: Thanks, I've just merged feat/nft-gallery into this branch and that has fixed this issue

src/Nfts/NftsNavigator.tsx
michalszorad: We can use `memo(Nfts)`  and this should solve the issue, right?

michalszorad: Hmm, using memo does not seem to help

src/hooks/index.ts
stackchain: ➕

src/Nfts/NftsNavigator.tsx
michalszorad: @wolverineks, what are your thoughts?

src/hooks/index.ts
michalszorad: That's now moved to Nfts/hooks.ts

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,26 @@
+import React, {createContext, ReactNode, useContext, useState} from 'react'
+
+interface SearchContextApi {
+  search: string
+  setSearch: (search: string) => void
+  clearSearch: () => void
+}
```
stackchain, line 9: it is looking like such great candidates for `actions` -> `useReduce`.

src/Nfts/Nfts.tsx
```diff
@@ -1,70 +1,84 @@
...
+        <ScrollView
+          style={styles.scrollView}
+          contentContainerStyle={styles.scrollViewError}
+          refreshControl={<RefreshControl onRefresh={refetch} refreshing={isRefetching} />}
+        >
+          <NoNftsScreen count={<NftCount count={filteredNfts.length} />} />
+        </ScrollView>
+      </ScreenWrapper>
+    )
+  }
```
stackchain: these two states for `NoNftsScreen` are looking so similar, maybe the design could update the screen to contain a message when there are no NFTS for a particular searching term, like, `isSearching` -> no NFTs were found based on your search. ???

stackchain (commented):
⬆️

src/Nfts/NftsNavigator.tsx
michalszorad: @stackchain I think we discussed with @wolverineks  moving the toggle out of context for the Search context to have a single responsibility, as the toggle is only being used here.

src/Nfts/Nfts.tsx
michalszorad: @wolverineks can I close this conversation?

src/Search/SearchContext.tsx
michalszorad: I've just added `searchReducer` please let me know your thoughts

src/Nfts/Nfts.tsx
michalszorad: I'll check with Product

michalszorad: Good point, I've missed the message in the designs. That's now updated

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
+import React, {createContext, ReactNode, useContext, useReducer} from 'react'
+
+interface SearchContextApi {
+  search: string
+  setSearch: (search: string) => void
+  clearSearch: () => void
+}
```
stackchain:
```suggestion
type SearchState {
  search: string
}

type SearchActions {
  searchChanged: (search: string) => void
  clearSearch: () => void
}
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const SearchContext = createContext<SearchContextApi | null>(null)
```
stackchain:
```suggestion
const SearchContext = createContext<undefined | (SearchState & SearchActions))>(undefined)
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const useSearch = () => {
+  const value = useContext(SearchContext)
+  if (!value) {
+    throw new Error('useSearch must be used within a SearchProvider')
+  }
+  return value
+}
+
+export const SearchProvider = ({children, value}: {children: ReactNode; value?: Partial<SearchContextApi>}) => {
```
stackchain:
```suggestion
export const SearchProvider = ({children, initialState}: {children: ReactNode; initialState?: Partial<SearchState>}) => {
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const [search, dispatch] = useReducer(searchReducer, DEFAULT_SEARCH_VALUE)
```
stackchain:
```suggestion
  const [state, dispatch] = useReducer(searchReducer, initialState)
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const clearSearch = () => dispatch({type: 'CLEAR'})
+  const setSearch = (search: string) => dispatch({type: 'SET', payload: search})
+  const defaultValue = {search, setSearch, clearSearch}
```
stackchain:
```suggestion
  const actions = React.useRef<SearchActions>({
    clearSearch: () => dispatch({type: 'clear'}),
    searchChanged: (search: string) => dispatch({type: 'searchChanged', payload: search})  
  }).current
  
  const context = React.useMemo(() => ({...state, ...actions}), [state, actions])
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return <SearchContext.Provider value={{...defaultValue, ...value}}>{children}</SearchContext.Provider>
```
stackchain:
```suggestion
  return <SearchContext.Provider value={context}>{children}</SearchContext.Provider>
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type ClearSearchAction = {
+  type: 'CLEAR'
+}
+
+type SetSearchAction = {
+  type: 'SET'
+  payload: string
+}
```
stackchain:
```suggestion
type SearchAction = 
  |  {
        type: 'clear'
    }
  |  {
        type: 'searchChanged'
        search: string
    }
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+type SearchAction = ClearSearchAction | SetSearchAction
```
stackchain:
```suggestion
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+function searchReducer(state: string, action: SearchAction) {
```
stackchain:
```suggestion
function searchReducer(state: SearchState, action: SearchAction) {
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  switch (action.type) {
+    case 'CLEAR':
```
stackchain:
```suggestion
    case 'clear':
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+      return DEFAULT_SEARCH_VALUE
+    case 'SET':
```
stackchain:
```suggestion
    case 'searchChanged':
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+      return action.payload
```
stackchain: maybe keep the state as string, just map type
```suggestion
      return {
          ...state,
          search,
      }
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+    default:
+      return state
```
stackchain:
```suggestion
      throw new Error(`searchReducer invalid action`)
```

src/Search/SearchContext.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+}
+
+const DEFAULT_SEARCH_VALUE = ''
```
stackchain:
```suggestion
const initialState: SearchState = Object.freeze({search: ''})
```

stackchain (changes requested):

banklesss:
Hi @MichalSzorad 

Recently a context was added to the `Send`  feature folder. There the context was included inside the folder `Context`. Would be useful do the same here and apply the same idea in future contexts?

michalszorad:
Hello @jemubm , I've always tried to keep the folder structure flat for more straightforward navigation through files and folders. In this specific case, there's only one file in the `Context` folder, which makes me wonder if the `Context` folder really needs to be there. However, I'm happy to go with one or another approach if we keep it consistent. What do you suggest?

stackchain (approved):

PR #2311, opened 2023-01-27
cleanup tokens

wolverineks:


src/Send/AssetSelectorScreen/AssetSelectorScreen.tsx
```diff
@@ -118,8 +112,7 @@ const AssetSelectorItem = ({wallet, tokenInfo, onSelect, matcher}: AssetSelector
 
         <View style={{flex: 1, padding: 4}}>
           <Text numberOfLines={1} ellipsizeMode="middle" style={{color: COLORS.BLUE_LIGHTER}} testID="tokenInfoText">
-            {/* eslint-disable-next-line @typescript-eslint/strict-boolean-expressions */}
-            {tokenInfo.ticker || tokenInfo.name || '-'}
+            {tokenInfo.ticker ?? tokenInfo.name ?? '-'}
```
wolverineks, line 115: better types

src/Send/AssetSelectorScreen/AssetSelectorScreen.tsx
```diff
@@ -60,7 +54,7 @@ export const AssetSelectorScreen = () => {
       </View>
 
       <FlatList
-        data={sortedTokenInfos}
+        data={sortTokenInfos({wallet, tokenInfos})}
```
wolverineks, line 57: shared sorting

src/Send/AssetSelectorScreen/AssetSelectorScreen.tsx
```diff
@@ -47,7 +41,7 @@ export const AssetSelectorScreen = () => {
   return (
     <SafeAreaView edges={['left', 'right', 'bottom']} style={{flex: 1, backgroundColor: 'white'}}>
       <View style={{paddingTop: 16, paddingHorizontal: 16}}>
-        <SearchInput onChangeText={(text) => onChangeMatcher(normalize(text))} />
+        <SearchInput onChangeText={(text) => onChangeMatcher(text)} />
```
wolverineks, line 44: moved into `matches`

src/TxHistory/AssetList/AssetList.tsx
```diff
@@ -55,7 +49,7 @@ export const AssetList = (props: Props) => {
 
       <FlatList
         {...props}
-        data={sortedTokenInfos}
+        data={sortTokenInfos({wallet, tokenInfos})}
```
wolverineks, line 52: shared sorting

src/TxHistory/AssetList/AssetList.tsx
```diff
@@ -87,7 +81,7 @@ const AssetItem = ({tokenInfo, onPress}: AssetItemProps) => {
 
       <Middle>
         <Text numberOfLines={1} ellipsizeMode="middle" style={styles.tokenInfo} testID="tokenInfoText">
-          {tokenInfo.ticker ?? tokenInfo.name}
+          {tokenInfo.ticker ?? tokenInfo.name ?? '-'}
```
wolverineks, line 84: better types

src/yoroi-wallets/cardano/api/api.ts
```diff
@@ -113,42 +110,6 @@ export const getTokenInfo = async (tokenId: string, apiUrl: string) => {
   return entry ? tokenInfo(entry) : fallbackTokenInfo(tokenId)
 }
```
wolverineks, line 112: moved to utils

src/yoroi-wallets/cardano/api/api.ts
```diff
@@ -166,58 +127,13 @@ export const fetchCurrentPrice = async (currency: CurrencySymbol, config: Backen
   return response.ticker.prices[currency]
 }
 
-export const hexToAscii = (hex: string) => Buffer.from(hex, 'hex').toLocaleString()
```
wolverineks, line 169: moved to utils

src/yoroi-wallets/cardano/api/api.ts
```diff
@@ -233,13 +149,25 @@ type Signature = {
...
+  value: T | undefined
 }
 
 const parseTokenRegistryEntry = (data: unknown) => {
   return isTokenRegistryEntry(data) ? data : undefined
 }
 
 const isTokenRegistryEntry = (data: unknown): data is TokenRegistryEntry => {
-  return !!data && typeof data === 'object' && 'name' in data && 'assetName' in data && 'policyId' in data
+  const candidate = data as TokenRegistryEntry
```
wolverineks, line 160: fixed type check

src/Send/AssetSelectorScreen/AssetSelectorScreen.tsx
```diff
@@ -152,16 +146,18 @@ const Icon = (props) => (
...
 const SearchInput = (props) => {
   const strings = useStrings()
 
   return <TextInput {...props} label={strings.searchLabel} />
 }
 
 const matches = (tokenInfo: TokenInfo, matcher: string) =>
-  Object.values(tokenInfo).some((key) => normalize(String(key)).includes(normalize(matcher)))
+  Object.values(tokenInfo)
+    .filter((value): value is string | number => value != null)
```
michalszorad, line 157: let's use strict checks and also would recommend to use `typeof value === 'string' || typeof value === 'number'` to make the TypeGuard more future proof if we add a new property to `TokenInfo` that has a different type from `string` `number` and `undefined`

src/yoroi-wallets/cardano/api/utils.ts
```diff
@@ -0,0 +1,123 @@
...
+  const tokenSubject = toTokenSubject(tokenIdentifier)
+  const assetName = hexToAscii(tokenSubject.slice(56)) || undefined
+  return assetName
+}
+
+export const toTokenSubject = (tokenIdentifier: string) => tokenIdentifier.replace('.', '')
+export const toTokenId = (tokenIdentifier: string) => {
+  const tokenSubject = toTokenSubject(tokenIdentifier)
+  return `${tokenSubject.slice(0, 56)}.${tokenSubject.slice(56)}`
+}
```
michalszorad, line 64: Is this just an abstraction so we can move to use AssetId as a token id? Also, would it be worth re-using the existing functions? Eg

```
export const toTokenId = (tokenIdentifier: string) => {
  const tokenSubject = toTokenSubject(tokenIdentifier)
  return `${toPolicyId(tokenIdentifier)}.${toAssetNameHex(tokenIdentifier)}`
}
```

michalszorad (approved):

src/utils/sorting.ts
```diff
@@ -0,0 +1,23 @@
+import {YoroiWallet} from '../yoroi-wallets'
+import {TokenInfo} from '../yoroi-wallets/types'
+
+export const sortTokenInfos = ({wallet, tokenInfos}: {wallet: YoroiWallet; tokenInfos: Array<TokenInfo>}) =>
```
stackchain, line 4: @wolverineks in my head is something that we should export within `yoroi-wallets`, so we sort on all platforms in the same way. Maybe moving the `/utils` to `yoroi-wallets/` ?

src/Send/AssetSelectorScreen/AssetSelectorScreen.tsx
stackchain: not sure about it, `NaN` will resolve to "NaN" anyways. maybe a clear parser?

src/TxHistory/AssetList/AssetList.tsx
stackchain: mental note: come with something to unify this logic, maybe a utility function `displayName`

src/yoroi-wallets/cardano/api/utils.ts
stackchain: `toPolicyId` makes sense, but I think the `tokenIdentifier` name is already in hex.

stackchain (commented):
⬆️

src/yoroi-wallets/cardano/api/utils.ts
wolverineks: i was going for...
```
tokenIdentifier: tokenId | tokenSubject
tokenId:         policyId . assetNameHex
tokenSubject:    policyId   assetNameHex
```

wolverineks: ultimately i want to deprecate `tokenIdentifier`, `tokenSubject`, and `tokenId`
and use `token.id = toAssetFingerprint(policyId, assetNameHex)`

PR #2356, opened 2023-02-24
keyboard spacer on android fix

banklesss:
Related to [YOMO-476](https://emurgo.atlassian.net/browse/YOMO-476)

[YOMO-476]: https://emurgo.atlassian.net/browse/YOMO-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wolverineks (approved):

src/components/KeyboardSpacer.tsx
```diff
@@ -1,25 +1,29 @@
 import * as React from 'react'
-import {Animated, Keyboard} from 'react-native'
+import {Animated, Keyboard, Platform} from 'react-native'
+
+const keyboardShowEvent = Platform.OS === 'android' ? 'keyboardDidShow' : 'keyboardWillShow'
+const keyboardHideEvent = Platform.OS === 'android' ? 'keyboardDidHide' : 'keyboardWillHide'
```
michalszorad, line 5: Should we also use this in the confirm screen?

banklesss: Hi!

`KeyboardSpacer` component is used in the confirm screen so the fix is applied there. That means ios and android keyboard animations have the same behaviour.

michalszorad (approved):

PR #2359, opened 2023-02-28
Fix NFT search bar position and metadata spacing

michalszorad:

PR #2363, opened 2023-03-01
fix header jump

wolverineks:


src/NftDetails/NftDetails.tsx
```diff
@@ -145,8 +144,36 @@ const NftMetadataPanel = ({nft}: {nft: YoroiNft}) => {
         </Link>
       </MetadataRow>
 
-      <Spacer height={24} />
-    </>
```
michalszorad: The spacer was added after a design review from Dima to ensure the HR is visible

stackchain:
is it related to #2359 ?

src/NftDetails/NftDetails.tsx
michalszorad: @wolverineks I think this is good to be merged after we add back the spacer

michalszorad (approved):

PR #2375, opened 2023-03-06
upgrade storybook

wolverineks:


package.json
```diff
@@ -96,7 +96,7 @@
     "@formatjs/intl-pluralrules": "^4.1.3",
     "@formatjs/intl-relativetimeformat": "^9.2.3",
     "@ledgerhq/react-native-hw-transport-ble": "6.27.3",
-    "@react-native-async-storage/async-storage": "1.15.5",
+    "@react-native-async-storage/async-storage": "^1.17.11",
```
stackchain, line 99: is it required for this PR ?

.storybook/stories/Button/ExampleButton.stories.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
```
stackchain, line 1: we will keep these examples?

stackchain (approved):

.storybook/stories/Button/ExampleButton.stories.js
wolverineks: it's a whole new story format going forward. thought an example would be worthwhile for now

package.json
wolverineks: dunno, that was part of the init script

wolverineks: does this break anything?

stackchain: minor version, not supposed to, just curious why is it bumping it 🤔

banklesss (approved):

PR #2379, opened 2023-03-08
Add moderation status behind a feature flag

michalszorad:
https://emurgo.atlassian.net/browse/YOMO-485
https://emurgo.atlassian.net/browse/YOMO-484


src/NftDetails/NftDetails.tsx
```diff
@@ -26,7 +27,7 @@ export const NftDetails = () => {
   return (
     <FadeIn style={styles.container}>
       <ScrollView contentContainerStyle={styles.contentContainer}>
-        <NftImage nft={nft} />
+        {MODERATING_NFTS_ENABLED ? <ModeratedNftImage nft={nft} /> : <UnModeratedNftImage nft={nft} />}
```
banklesss, line 30: I think eslint is going to complain if `MODERATING_NFTS_ENABLED === true` is not used

src/Nfts/ImageGallery/ImageGallery.tsx
```diff
@@ -30,7 +31,13 @@ export const ImageGallery = ({nfts = [], onSelect, onRefresh, isRefreshing}: Pro
       data={nfts}
       onRefresh={onRefresh}
       refreshing={isRefreshing}
-      renderItem={(nft) => <ModeratedImage onPress={() => onSelect(nfts.indexOf(nft))} nft={nft} key={nft.id} />}
+      renderItem={(nft) =>
+        MODERATING_NFTS_ENABLED ? (
```
banklesss, line 35: I think eslint is going to complain if `MODERATING_NFTS_ENABLED === true` is not used

src/NftDetails/NftDetails.tsx
michalszorad: eslint is not complaining as `MODERATING_NFTS_ENABLED` is expected to be boolean

michalszorad:
@stackchain is the SonarQube checking coverage on all components and if so, should we start adding unit tests for components?

src/Nfts/ImageGallery/ImageGallery.tsx
michalszorad: eslint is not complaining as `MODERATING_NFTS_ENABLED` is expected to be boolean

stackchain:
> @stackchain is the SonarQube checking coverage on all components and if so, should we start adding unit tests for components?

we are still configuring it.

PR #2390, opened 2023-03-13
clean up config envs

wolverineks:
move feature flags from `config.ts`  to `feature.ts`
clean up imports


src/features.ts
```diff
@@ -11,4 +8,33 @@ export default {
     search: __DEV__ || false,
     nfts: __DEV__ || false,
   },
+  useTestnet: __DEV__ ? false : false,
```
stackchain, line 11: for folders `useX` is ok, for vars maybe it can cause some lint issues when importing within react component lets see.

stackchain (approved):

src/features.ts
```diff
@@ -11,4 +8,33 @@ export default {
...
+  startWithIndexScreen: __DEV__ ? false : false,
+  prefillWalletInfo: __DEV__ ? false : false,
+  showProdPoolsInDev: __DEV__ ? false : false,
+  showNftGallery: __DEV__ ? true : false,
+  moderatingNftsEnabled: __DEV__ ? false : false,
```
michalszorad, line 16: Let's say someone wants to override these values for local development, and the only way to do it is to modify this file tracked by git. When creating a commit, the person needs to remember not to commit this file. Will it not add extra mental overhead?

e2e/createWallet.spec.js
```diff
@@ -1,6 +1,6 @@
 // @flow
```
banklesss:
```suggestion
```

michalszorad (approved):

src/features.ts
stackchain: I'm ok with that, as long as the PROD is always in the elses' case :D

stackchain: working with feature branches will make this concern go away, also when moving to prod the flag will be gone (at least it's supposed to)

wolverineks: gitignore features?

PR #2417, opened 2023-03-31
Extract search logic from nfts part2

banklesss:
Related to [YOMO-502](https://emurgo.atlassian.net/browse/YOMO-502)

[YOMO-502]: https://emurgo.atlassian.net/browse/YOMO-502?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Nfts/Nfts.tsx
```diff
@@ -4,23 +4,31 @@ import {RefreshControl, ScrollView, StyleSheet, Text, View} from 'react-native'
...
 
-  const {search, nfts, isLoading, refetch, isError} = useFilteredNfts({
+  const {isLoading, nfts, refetch, isError} = useNfts(wallet, {
     onSettled: () => {
       if (isManualRefreshing) setIsManualRefreshing(false)
     },
   })
 
+  const {search: nftsSearchTerm} = useSearch()
+  const nftsSearchResult = filterNfts(nftsSearchTerm, nfts).sort((NftA, NftB) => NftA.name.localeCompare(NftB.name))
```
stackchain: save capitals for classes/modules, still I'll be fine with `a,b`.
```suggestion
  const nftsSearchResult = filterNfts(nftsSearchTerm, nfts).sort((nftA, nftB) => nftA.name.localeCompare(nftB.name))
```

stackchain: sorting after searching, implies that only filtered nfts will be sorted, I'd suggest splitting into `sortedNfts` and applying the filter on top of the sorted nfts, or even sorted right before adding the nfts into the data for the list.

src/Nfts/Nfts.tsx
```diff
@@ -37,12 +45,12 @@ export const Nfts = () => {
...
     return (
       <ScreenWrapper>
-        <LoadingScreen nftsCount={nfts.length} />
+        <LoadingScreen nftsCount={nftsSearchResult.length} />
       </ScreenWrapper>
     )
   }
 
-  if (search.length > 0 && nfts.length === 0) {
+  if (nftsSearchTerm.length > 0 && nftsSearchResult.length === 0) {
```
stackchain: const `hasEmptySearchResult` ?

src/Nfts/filterNfts.ts
```diff
@@ -0,0 +1,10 @@
+import {YoroiNft} from '../yoroi-wallets'
+
+export const filterNfts = (nftsSearchTerm: string, nfts: YoroiNft[]): YoroiNft[] => {
```
stackchain: `search: string` or `term: string`, `searchTerm`  would be fine for me, the function name is `filterNfts` so I think nft its implied

src/Nfts/filterNfts.ts
```diff
@@ -0,0 +1,10 @@
...
+  const searchTermLowerCase = nftsSearchTerm.toLowerCase()
+  const filteredNfts =
+    searchTermLowerCase.length > 0 && nfts.length > 0
```
stackchain: I think `&& nfts.leng..` is not necessary

stackchain (commented):
⬆️

src/Nfts/Nfts.tsx
wolverineks: why sort nfts that are not displayed?

src/Nfts/Nfts.tsx
```diff
@@ -4,23 +4,34 @@ import {RefreshControl, ScrollView, StyleSheet, Text, View} from 'react-native'
...
-  const {search, nfts, isLoading, refetch, isError} = useFilteredNfts({
+  const {isLoading, nfts, refetch, isError} = useNfts(wallet, {
     onSettled: () => {
       if (isManualRefreshing) setIsManualRefreshing(false)
     },
   })
 
+  const {search: nftsSearchTerm} = useSearch()
+  const sortedNfts = sortNfts(nfts)
+  const nftsSearchResult = filterNfts(nftsSearchTerm, sortedNfts)
```
wolverineks, line 30: extract the functions used to sort/filter, but don't wrap the sort/filter. something like:
`const filteredNfts = nfts. filter(match(search))`
`const sortedNfts = nfts.sort(byName)`

src/Nfts/Nfts.tsx
```diff
@@ -4,23 +4,34 @@ import {RefreshControl, ScrollView, StyleSheet, Text, View} from 'react-native'
...
+
+  const handleNftSelect = (index: number) => navigateTo.nftDetails(nftsSearchResult[index].id)
```
wolverineks: don't use index

src/Nfts/Nfts.tsx
wolverineks: filter then sorry

src/Nfts/Nfts.tsx
stackchain: yap, that makes sense, sort for last right before adding to the list component.

src/Nfts/Nfts.tsx
stackchain: yap, I think it can receive the `id` directly.

src/Nfts/Nfts.tsx
```diff
@@ -68,7 +79,7 @@ export const Nfts = () => {
             message={strings.noNftsInWallet}
             heading={
               <View>
```
wolverineks, line 81: why does NoNFTsScreen need an nft count? shouldn't it always be 0?

wolverineks: "*Screens" should be directly used in navigators, not inside other components

src/Nfts/Nfts.tsx
banklesss: what name would you suggest? it is the index of the nfts array

wolverineks: not the name, i mean literally, dont use indexes

wolverineks: `onSelect(index: number)` -> `onSelect(id: string)`

wolverineks: the selection process is `nft -> index -> nft -> id`
just pass the `id`

stackchain (approved):
lgtm

src/Nfts/Nfts.tsx
```diff
@@ -4,61 +4,72 @@ import {RefreshControl, ScrollView, StyleSheet, Text, View} from 'react-native'
...
   })
 
+  const {search: nftsSearchTerm} = useSearch()
+  const filteredNfts = filterNfts(nftsSearchTerm, nfts)
+  const sortedNfts = filteredNfts.sort(sortNfts)
+  const nftsSearchResult = filterNfts(nftsSearchTerm, sortedNfts)
+
+  const hasEmptySearchResult = nftsSearchTerm.length > 0 && nftsSearchResult.length === 0
+  const hasNotNfts = nftsSearchTerm.length === 0 && nftsSearchResult.length === 0
+
```
wolverineks, line 34: why `useCallback`

src/Nfts/Nfts.tsx
```diff
@@ -4,61 +4,72 @@ import {RefreshControl, ScrollView, StyleSheet, Text, View} from 'react-native'
...
     },
   })
 
+  const {search: nftsSearchTerm} = useSearch()
+  const filteredNfts = filterNfts(nftsSearchTerm, nfts)
+  const sortedNfts = filteredNfts.sort(sortNfts)
+  const nftsSearchResult = filterNfts(nftsSearchTerm, sortedNfts)
+
+  const hasEmptySearchResult = nftsSearchTerm.length > 0 && nftsSearchResult.length === 0
+  const hasNotNfts = nftsSearchTerm.length === 0 && nftsSearchResult.length === 0
```
wolverineks: why do you include search term when the wallet doesn't have any nfts?

src/Nfts/Nfts.tsx
```diff
@@ -68,35 +79,40 @@ export const Nfts = () => {
...
     )
   }
 
   return (
-    <ScreenWrapper>
+    <Wrapper>
       <View style={styles.galleryContainer}>
-        {search.length === 0 && (
+        {nftsSearchTerm.length === 0 && (
           <View>
```
wolverineks: `<View>` -> `<>`

src/Nfts/Nfts.tsx
```diff
@@ -140,7 +156,7 @@ function ErrorScreen({onRefresh, isRefreshing}: {onRefresh: () => void; isRefres
   )
 }
 
-function NftCount({count}: {count?: number | string}) {
+const NftCount = ({count}: {count?: number | string}) => {
```
wolverineks, line 159: inline countText

wolverineks: redundant `<View>`

src/Nfts/Nfts.tsx
wolverineks: if the wallet doesn't have any nfts, don't show search

src/Nfts/Nfts.tsx
```diff
@@ -4,61 +4,72 @@ import {RefreshControl, ScrollView, StyleSheet, Text, View} from 'react-native'
...
-        <LoadingScreen nftsCount={nfts.length} />
-      </ScreenWrapper>
+      <Wrapper>
+        <LoadingScreen nftsCount={nftsSearchResult.length} />
+      </Wrapper>
     )
   }
 
-  if (search.length > 0 && nfts.length === 0) {
+  if (hasEmptySearchResult) {
```
wolverineks, line 56: (isError, isLoading) = 1 scope
(hasNotNfts, hasEmptySearchResults) = different scope
don't mix scopes

src/Nfts/Nfts.tsx
```diff
@@ -165,6 +181,8 @@ function LoadingScreen({nftsCount}: {nftsCount: number}) {
   )
 }
 
+const sortNfts = (nftA: YoroiNft, nftB: YoroiNft): number => nftA.name.localeCompare(nftB.name)
```
wolverineks: don't pass in specific objects
pass in primitives
make it generic

src/Nfts/Nfts.tsx
banklesss: could you specify please?

src/Nfts/Nfts.tsx
banklesss: What would be the implementation that you suggest? Could you specify?

Thanks!

src/Nfts/Nfts.tsx
banklesss: > if the wallet doesn't have any nfts, don't show search

 I think this is something to be considered by product

src/Nfts/Nfts.tsx
banklesss: got it

src/Nfts/Nfts.tsx
banklesss: I created a ticket to implement the rescoping https://emurgo.atlassian.net/browse/YOMO-518

Thanks for the suggestion.

banklesss:
Rebase here https://github.com/Emurgo/yoroi-mobile/pull/2424

PR #2422, opened 2023-04-04
feat(nft-gallery): Show skeleton when loading image

michalszorad:
Resolves [YOMO-511](https://emurgo.atlassian.net/browse/YOMO-511)

[YOMO-511]: https://emurgo.atlassian.net/browse/YOMO-511?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/NftPreview/NftPreview.tsx
```diff
@@ -34,37 +36,47 @@ export const NftPreview = ({
       isSvgMediaType(getNftFilenameMediaType(nft, uri)))
   const shouldShowPlaceholder = !isString(uri) || showPlaceholder || (isUriSvg && blurRadius !== undefined) || error
 
+  useEffect(() => {
```
stackchain, line 39: is it really necessary?

wolverineks:
why does this use Error Boundary?

src/components/NftPreview/NftPreview.tsx
michalszorad: When the parent changes the NFT to a NFT with a different `uri` or changes the `showThumbnail` the state will be kept as `loaded: true`, but the image is not loaded yet

michalszorad:
@wolverineks to catch errors inside `SvgUri` for example. See here https://github.com/software-mansion/react-native-svg/issues/2011

src/NftDetails/NftDetailsNavigator.tsx
```diff
@@ -25,7 +25,7 @@ export const NftDetailsNavigator = () => {
         component={NftDetails}
       />
 
-      <Stack.Screen name="image-zoom" options={{headerShown: false}} component={NftDetailsImage} />
+      <Stack.Screen name="image-zoom" options={{headerTitle: () => null}} component={NftDetailsImage} />
```
michalszorad, line 28: Resolves YOMO-513

src/components/NftPreview/NftPreview.tsx
banklesss: What if the nft id is set as the key of this component so that when the nft changes the component is rendered again? The loading is `true` as default and when `onLoad` triggers is set to `false`

banklesss: Not sure if it will work

stackchain: I'd say that it should receive the uri instead of the nft here.

stackchain: cuz the nft info is just to figure the mimetype

stackchain:
created a ticket to move from instead of passing nft -> uri + mimetype, move nftpreview to imagepreview.

stackchain (approved):

PR #2442, opened 2023-04-19
Fix always allow select the primary token

stackchain:
- MAT should allow the primary token to be added regardless of the secondary tokens limit.


src/features/Send/useCases/ListAmountsToSend/AddToken/SelectTokenFromListScreen.tsx
```diff
@@ -57,7 +57,11 @@ export const SelectTokenFromListScreen = () => {
         data={sortedTokenInfos}
         renderItem={({item: tokenInfo}: {item: TokenInfo}) => (
           <Boundary>
-            <SelectableAssetItem tokenInfo={tokenInfo} disabled={!canAddAmount} wallet={wallet} />
+            <SelectableAssetItem
+              tokenInfo={tokenInfo}
+              disabled={!canAddAmount && tokenInfo.id !== wallet.primaryTokenInfo.id}
```
stackchain, line 62: primary token always should be able to be added

banklesss (approved):

PR #2463, opened 2023-05-10
chore: bumped bridge + lib

stackchain:


apps/wallet-mobile/src/Catalyst/QrCode.tsx
```diff
@@ -16,7 +16,7 @@ const {FlagSecure} = NativeModules
 
 export const QrCode = ({onNext, votingKeyEncrypted}: {onNext: () => void; votingKeyEncrypted: string}) => {
   useBlockGoBack()
-  useAllowScreenshot()
+  // useAllowScreenshot()
```
michalszorad: This seems to crash the screen. I'll investigate

michalszorad: This should be now resolved

apps/wallet-mobile/src/yoroi-wallets/cardano/constants/common.ts
```diff
@@ -41,6 +41,14 @@ export const BIP44_DERIVATION_LEVELS = {
   ADDRESS: 5,
 } as const
 
+export const VOTING_KEY_PATH = [
+  1694 + HARD_DERIVATION_START,
+  1815 + HARD_DERIVATION_START,
+  0 + HARD_DERIVATION_START,
+  0,
+  0,
+]
```
stackchain, line 50:
```suggestion
] as const
```

stackchain:
@MichalSzorad does it fix for `iOS` too?

michalszorad:
@stackchain I couldn't find a way to block screenshots on iOS. There are several conversations this may not be possible

- https://github.com/hawkup/react-native-prevent-screenshot-poc 
- https://stackoverflow.com/questions/66113565/is-there-any-way-to-prevent-screenshot-in-react-native-ios-version-0-63
- https://docs.expo.dev/versions/latest/sdk/screen-capture/

PR #2464, opened 2023-05-11
chore: metrics/fixes

stackchain:
#### Blocked by

- [ ] Product : Design for opt-out & Settings
- [ ] Legal : ToS update


lefthook.yml
```diff
@@ -0,0 +1,35 @@
+# EXAMPLE USAGE:
```
michalszorad: If everything is commented out, can we remove this file?

packages/metrics-node/package.json
```diff
@@ -0,0 +1,141 @@
+{
+  "name": "@yoroi/metrics-node",
```
michalszorad, line 2: Will it make sense to have readme in each package folder?

packages/types/src/metrics/nft.ts
```diff
@@ -0,0 +1,115 @@
+import {MetricsTrackProperties} from './properties'
+
+export type MetricsNftEvent =
+  | 'nft_click_navigate'
+  | 'nft_list_load'
+  | 'nft_list_click_grid_big'
+  | 'nft_list_click_grid_small'
```
michalszorad, line 7: Do we have any documentation for all these events?

packages/metrics-react/scripts/flowgen.sh
```diff
@@ -0,0 +1,3 @@
+for i in $(find lib -type f -name "*.d.ts");
+  do sh -c "npx flowgen $i -o ${i%.*.*}.js.flow";
```
michalszorad, line 2: I've noticed this file is used across multiple packages, but is flow ever used or are there any plans on using it?

packages/metrics-node/lefthook.yml
```diff
@@ -0,0 +1,16 @@
...
+  parallel: true
+  commands:
+    lint:
+      files: git diff --name-only @{push}
+      glob: "*.{js,ts,jsx,tsx}"
+      run: npx eslint {files}
+    types:
+      files: git diff --name-only @{push}
+      glob: "*.{js,ts,jsx,tsx}"
+      run: npx tsc --noEmit
```
michalszorad: `files` and `glob` may not be needed here as `npx tsc --noEmit` does not use the variable. Same applies to other lefthook.yml files

packages/metrics-react/src/adapters/amplitude.ts
```diff
@@ -0,0 +1,36 @@
...
+    setUserId: (userId: string) => {
+      deps.analytics.setUserId(userId)
+    },
+    setDeviceId: (deviceId: string) => {
+      deps.analytics.setDeviceId(deviceId)
+    },
+    setSessionId: (sessionId: number) => {
+      deps.analytics.setSessionId(sessionId)
+    },
+  } as const as Metrics.Module<Amplitude.Types.EventOptions>
```
michalszorad, line 35: Can't wait for TS 4.9 so we can use `satisfies` here

apps/wallet-mobile/.env
```diff
@@ -1,12 +1,16 @@
 BUILD_VARIANT=STAGING
-USE_TESTNET=true
+
 COMMIT=5caf3ead
-SHOW_PROD_POOLS_IN_DEV=true
 SENTRY=https://59637d7dcb274b6a9e9f708e0082b3d5@o1138840.ingest.sentry.io/6260389
+AMPLITUDE_API_KEY=a518e0983cb9e4f1cccb9edcd66b3897
```
michalszorad, line 5: Is this key used for all environments?

lefthook.yml
stackchain: it comes by default with `bob` builder, yes we can.

packages/metrics-node/package.json
stackchain: I believe so since the installation instructions would be a little bit different.

packages/types/src/metrics/nft.ts
stackchain: It was agreed with Product, for now, it's placeholders of the events that we can track, the final specs are coming and break it down in Product (focused on funnels) and Business (metrics and KPIs)

packages/metrics-react/scripts/flowgen.sh
stackchain: it's here cuz this package is exported for extension and it's used over there. https://github.com/Emurgo/yoroi-frontend/pull/3211

packages/metrics-react/src/adapters/amplitude.ts
stackchain: yap :S

apps/wallet-mobile/.env
stackchain: it's for dev now, we would be touching it again, is blocked the `nightly` and `production` new pipelines. it's behind a feature flag.

michalszorad (approved):

banklesss (approved):

PR #2468, opened 2023-05-12
header title fix

banklesss:
Fixes this

Screenshot_20230512-124422_Yoroi


apps/wallet-mobile/src/navigation.tsx
```diff
@@ -38,6 +39,8 @@ export const defaultStackNavigationOptionsV2: StackNavigationOptions = {
   headerTitleStyle: {
     fontSize: 16,
     fontFamily: 'Rubik-Medium',
+    width: WIDTH - 75,
```
michalszorad, line 42: Why 75? It may be worth creating a constant if this refers to some other element width

banklesss: got the fix here https://stackoverflow.com/a/57727917

Do you suggest another solution?

michalszorad: Hmm yeah I see. Maybe it would be worth adding a comment explaining why 75 with a link to the StackOverflow issue @stackchain what are your thoughts?

stackchain: 👀

michalszorad (approved):

PR #2476, opened 2023-05-19
Use intl to format date and remove moment

michalszorad:
Resolves YOMO-470


apps/wallet-mobile/package.json
```diff
@@ -112,7 +112,6 @@
     "expo-camera": "^13.2.1",
     "jsc-android": "241213.1.0",
     "lodash": "^4.17.21",
-    "moment": "2.29.1",
```
michalszorad, line 118: Remove moment

apps/wallet-mobile/src/legacy/format.ts
```diff
@@ -129,25 +128,37 @@ export const formatAdaWithText = (amount: Quantity, defaultAsset: DefaultAsset)
...
-  return moment(ts).format((moment(0) as any)._locale._format.timeToSeconds)
+export const formatTime = (timestamp: string, intl: IntlShape) => {
+  return intl.formatTime(new Date(timestamp), {
+    timeStyle: 'medium',
+  })
 }
 
-export const formatDateToSeconds = (ts: string | any) => {
-  return moment(ts).format((moment(0) as any)._locale._format.dateToSeconds)
+export const formatDateAndTime = (timestamp: string, intl: IntlShape) => {
```
michalszorad, line 158: Refactored function name, removed `any` and changed variable name from `ts` to `timestamp`

apps/wallet-mobile/src/legacy/format.ts
```diff
@@ -129,25 +128,37 @@ export const formatAdaWithText = (amount: Quantity, defaultAsset: DefaultAsset)
...
-    sameElse: 'L',
+  const inputDateString = new Date(ts).toDateString()
+  const today = new Date().toDateString()
+  const yesterday = new Date(Date.now() - 24 * 60 * 60 * 1000).toDateString()
+
+  if (inputDateString === today) {
+    return intl.formatMessage(messages.today)
+  }
+
+  if (inputDateString === yesterday) {
```
michalszorad, line 180: Not sure if there's a better way to check today / yesterday without using moment

banklesss (approved):

apps/wallet-mobile/src/legacy/format.ts
```diff
@@ -129,25 +128,37 @@ export const formatAdaWithText = (amount: Quantity, defaultAsset: DefaultAsset)
   return `${formatAda(amount, defaultAsset)}${utfSymbols.NBSP}${defaultAssetMeta.ticker}`
 }
 
-export const formatTimeToSeconds = (ts: string | any) => {
-  return moment(ts).format((moment(0) as any)._locale._format.timeToSeconds)
+export const formatTime = (timestamp: string, intl: IntlShape) => {
```
stackchain, line 149: since it's expecting a string and it can be empty, I'd say it should be validated here, `.length === 0 => ''`, and not in every call.

apps/wallet-mobile/src/legacy/format.ts
stackchain: Why `.toDateString()` ?
```
function getYesterday() {
    const date = new Date();
    date.setDate(date.getDate() - 1);
    return date.toISOString().split('T')[0];
}
```

stackchain (changes requested):
⬆️

apps/wallet-mobile/src/legacy/format.ts
michalszorad: I've added an empty string check. However, the caller will need to check for null | undefined, as this fn accepts string only

PR #2479, opened 2023-05-24
fix(debugger): Override buffer's subarray and remove react-native-buffer

michalszorad:
Launching a debugger makes the app crash.

The issue was the `react-native-quick-base64` used by `@craftzdog/react-native-buffer` initializes a module synchronously which is not possible in debugger. I’ve removed `@craftzdog/react-native-buffer` and added a polyfill `Buffer.prototype.subarray = … `(a copy from `@craftzdog/react-native-buffer`) and this seems to solve the issue. I’ve tested the ledger bt connection and it works well. I would have kept the `@craftzdog/react-native-buffer`  if they did not import the `react-native-quick-base64` but this seems to be the core issue in the module.

Resolves YOMO-551
Screenshot 2023-05-18 at 16 42 28 (1)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1449,14 +1449,6 @@
   dependencies:
     chalk "^4.1.0"
 
-"@craftzdog/react-native-buffer@^6.0.5":
-  version "6.0.5"
-  resolved "https://registry.yarnpkg.com/@craftzdog/react-native-buffer/-/react-native-buffer-6.0.5.tgz#0d4fbe0dd104186d2806655e3c0d25cebdae91d3"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -16361,7 +16353,7 @@ react-native-qrcode-svg@6.0.6:
     prop-types "^15.5.10"
     qrcode "^1.3.2"
 
-react-native-quick-base64@^2.0.5, react-native-quick-base64@^2.0.6:
+react-native-quick-base64@^2.0.6:
```
banklesss, line 16356: should this be uninstalled?

michalszorad: ~My bad, let me fix this~

banklesss (approved):

yarn.lock
michalszorad: Actually this is only the lock file and v2.0.5 got deleted because it was only used by react-native-buffer. V2.0.6 is still there.

PR #2480, opened 2023-05-25
feat(mat): Navigate to transaction start screen from token select

michalszorad:
Resolves YOMO-515

Navigates to transaction start screen from asset select when the user has no selected assets and clicks on back. Otherwise, navigates to the list


apps/wallet-mobile/src/features/Send/useCases/StartMultiTokenTx/StartMultiTokenTxScreen.tsx
```diff
@@ -24,7 +23,7 @@ export const StartMultiTokenTxScreen = () => {
 
   const {targets, selectedTargetIndex, receiverChanged, memo, memoChanged, addressChanged} = useSend()
   const {address, amounts} = targets[selectedTargetIndex].entry
-  const shouldOpenSelectToken = Object.keys(amounts).length === 0
+  const shouldOpenAddToken = Object.keys(amounts).length === 0
```
michalszorad: Changed the name to match the `navigateTo.addToken()`

apps/wallet-mobile/src/features/Send/common/navigation.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const useNavigateTo = () => {
+  const navigation = useNavigation<TxHistoryRouteNavigation>()
+
+  return {
+    selectedTokens: () => navigation.navigate('send-list-amounts-to-send'),
+    addToken: () => navigation.navigate('send-select-token-from-list'),
+    startTx: () => navigation.navigate('send-start-tx'),
+  }
+}
```
michalszorad: Moved this to a new file

apps/wallet-mobile/src/features/Send/useCases/ListAmountsToSend/AddToken/SelectTokenFromListScreen.tsx
```diff
@@ -28,11 +29,24 @@ export type FungibilityFilter = 'all' | 'ft' | 'nft'
...
+  const navigateTo = useNavigateTo()
+  const {amounts} = targets[selectedTargetIndex].entry
+  const hasTokensSelected = Object.keys(amounts).length > 0
+
+  const navigateBack = () => {
+    if (hasTokensSelected) {
+      navigateTo.selectedTokens()
+    } else {
+      navigateTo.startTx()
+    }
```
michalszorad: If there are any selected tokens, it will navigate to `selectedTokens` otherwise `startTx`

apps/wallet-mobile/src/Search/SearchContext.tsx
```diff
@@ -93,15 +93,23 @@ export const useSearchOnNavBar = ({
...
   noBack?: boolean
+  navigateBack?(): boolean
 }) => {
   const navigation = useNavigation()
 
   const {search, visible, showSearch, hideSearch, clearSearch} = useSearch()
 
+  useEffect(() => {
+    if (!navigateBack) return
+    const subscription = BackHandler.addEventListener('hardwareBackPress', navigateBack)
```
michalszorad: Handles native gestures and buttons

banklesss (approved):

apps/wallet-mobile/src/Search/SearchContext.tsx
```diff
@@ -113,7 +121,13 @@ export const useSearchOnNavBar = ({
      *   1) go back when the search input is not visible
      *   2) close the search input when the search input is visible
      */
-    if (!visible) navigation.goBack()
+    if (visible) return
+
+    if (navigateBack) {
```
stackchain, line 122: I believe this will cause the navigation to never go back to the receiver when there is at least one amount added.

stackchain (commented):
⬆️

apps/wallet-mobile/src/Search/SearchContext.tsx
michalszorad: I'm having a look

michalszorad: @stackchain that should be fixed now

apps/wallet-mobile/src/Search/SearchContext.tsx
stackchain: hey @MichalSzorad in that case, it should hideSearch and clearSearch, otherwise it will present a different behaviour. Tap back (close the search), Swipe back will be navigating back, but won't close and clear the search.

stackchain (changes requested):

apps/wallet-mobile/src/utils/navigation.ts
```diff
@@ -0,0 +1,28 @@
...
+export function useOverrideBackNavigate(goBack: () => boolean) {
+  const navigation = useNavigation()
+  const [initialRouteId] = useState(() => getNavigationRouteId(navigation))
+
+  const handleGoBack = useCallback(() => {
+    const currentRouteId = getNavigationRouteId(navigation)
+
+    if (currentRouteId !== initialRouteId) {
+      return false
+    }
```
michalszorad, line 14: Since routes from the same stack are kept in memory, we need to check whether we're on the route where the hook is called

apps/wallet-mobile/src/Search/SearchContext.tsx
michalszorad: @stackchain We now pass `handleGoBack` to `useOverrideBackNavigate` which should ensure the functionality is the same

apps/wallet-mobile/src/features/Send/useCases/ListAmountsToSend/ListAmountsToSendScreen.tsx
```diff
@@ -189,6 +203,7 @@ const useNavigateTo = () => {
     addToken: () => navigation.navigate('send-select-token-from-list'),
     editAmount: () => navigation.navigate('send-edit-amount'),
     confirmTx: () => navigation.navigate('send-confirm-tx'),
+    startTx: () => navigation.navigate('send-start-tx'),
```
stackchain: update and import from common.

apps/wallet-mobile/src/features/Send/useCases/StartMultiTokenTx/StartMultiTokenTxScreen.tsx
stackchain: my bad here, it should be using `Amounts.toArray` helper, amounts must be manipulated with this helper always.

stackchain (changes requested):

stackchain (approved):

PR #2494, opened 2023-06-12
fix(transactions): Use local timezone to display transaction time

michalszorad:
Resolves YOMO-569

expo-localization  does not seem to work in dev mode https://github.com/expo/expo/issues/20026
react-native-localize breaks down the debugger (same issue as we experienced with buffer libraries)
react-native-timezone seems to work well https://www.npmjs.com/package/react-native-timezone


michalszorad:
I need to check whether this does not break debugger

apps/wallet-mobile/package.json
```diff
@@ -4,10 +4,6 @@
   "private": true,
   "scripts": {
     "android-bundle": "react-native bundle --platform android --dev false --entry-file index.js --bundle-output android/app/src/main/assets/index.android.bundle --assets-dest android/app/src/main/res",
-    "prerelease:android:nightly": "bundle install && ./scripts/update-commit.sh",
-    "release:android:nightly": "bundle exec fastlane android release --env nightly",
-    "prerelease:android:production": "bundle install && ./scripts/update-commit.sh",
-    "release:android:production": "bundle exec fastlane android release --env production",
```
michalszorad, line 10: This got reformatted automatically after running `yarn add ...` let's keep it this way, so it does not cause diff in future PRs?

apps/wallet-mobile/src/i18n/LanguageProvider.tsx
```diff
@@ -12,16 +13,38 @@ const LanguageContext = React.createContext<undefined | LanguageContext>(undefin
...
 
+const useTimezone = () => {
+  const query = useQuery({
+    queryKey: ['timeZone'],
+    queryFn: async () => {
+      const defaultTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone
+      const timeZone = await TimeZone.getTimeZone()
+      return timeZone ?? defaultTimeZone
+    },
+    suspense: true,
```
michalszorad, line 39: Uses suspense as `useLanguageCode`

stackchain:
@MichalSzorad I think is missing the pod install -> Podfile.lock update.

michalszorad:
@stackchain that's now updated

banklesss (approved):

PR #2505, opened 2023-06-20
fix: ledger catalyst voting

stackchain:
Resolves YOMO-564

Please note this needs to be tested using:

- Cardano APP v6 on Ledger
- Ledger needs to have experimental / developer features enabled


banklesss (approved):

apps/wallet-mobile/src/yoroi-wallets/cardano/shelley/ShelleyWallet.ts
```diff
@@ -702,6 +702,12 @@ export const makeShelleyWallet = (constants: typeof MAINNET | typeof TESTNET) =>
       })
     }
 
+    async getFirstPaymentAddress() {
+      const externalAddress = this.externalAddresses[0]
+      const addr = await Cardano.Wasm.Address.fromBech32(externalAddress)
+      return Cardano.Wasm.BaseAddress.fromAddress(addr)
+    }
```
michalszorad, line 709: @stackchain please let me know if this is what you meant. Thanks!

stackchain: yap that is the idea, than later you can`this.getAddressing` and return the addressing obj to use it inside the voting.

michalszorad: So instead of `SHELLEY_VOTING_KEY_PATH` it would be `addressing.path`?

stackchain: yap, can you write some tests too?

apps/wallet-mobile/src/yoroi-wallets/cardano/shelley/tests/mainnet/ShelleyWallet.test.ts
```diff
@@ -332,6 +335,62 @@ describe('ShelleyWallet', () => {
...
+    const isReadOnly = false
+
+    const wallet: YoroiWallet & Record<string, any> = await ShelleyWalletMainnet.createBip44({
+      id: walletMeta.id,
+      accountPubKeyHex,
+      storage: storage.join(`${walletMeta.id}/`),
+      hwDeviceInfo,
+      isReadOnly,
+    })
+    wallet._utxos = [
```
michalszorad, line 361: A hacky way here to modify utxos

michalszorad: @stackchain is there any better way?

stackchain: I believe we can use the current mocks, or also inject the `UtxoManager`, for now it's ok, we can improve the tests later.

PR #2520, opened 2023-06-29
fix qr reader when fails

banklesss:
Related to [YOMO-587](https://emurgo.atlassian.net/browse/YOMO-587)

There are an issue open in the expo-camera repo: [Barcode Scanner does not stop scanning · Issue #345 · expo/expo](https://github.com/expo/expo/issues/345) 

There is no way to stop the scan, so a loop is generated.

The idea I had is to remove the camera component when the qr has been read first time, either reading good or bad code. If it is correct, the user will go to the next screen. If it is bad, the user can go back and retry as I show in the video shared in the ticket.

[YOMO-587]: https://emurgo.atlassian.net/browse/YOMO-587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/components/QRCodeScanner/QRCodeScanner.tsx
+++ b/apps/wallet-mobile/src/components/QRCodeScanner/QRCodeScanner.tsx
@@ -19,12 +19,13 @@ export const QRCodeScanner = ({onRead}: {onRead: ({data}: {data: string}) => voi
     setQrScanned(true)
   }
 
-  if (!granted || qrScanned) {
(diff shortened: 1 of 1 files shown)


michalszorad (commented):
I'm wondering if it would be a slightly better UX if we redirect them back when the QR code is invalid instead of them having to press the back button. What are your thoughts? Otherwise good job!

stackchain (approved):

PR #2532, opened 2023-07-17
chore: Added analytics (feature flag)

jorbuedo:
- Adds `ampli` tool and scripts
- Init amplitude with 5 sec flush
- Opt out functionality (settings only, missing UI)
- Adds NFT Gallery tracking

- [ ] Blocked by Opt out login screen (check with product)

image


jorbuedo:
I've been trying to use plugins (recently discovered them browsing the docs) that sdk supports to do the mock thing instead of the Proxy. But there seems to be some issue with it. In the end, the behavior would be the same, so it can be changed later in any case.

apps/wallet-mobile/.env
```diff
@@ -1,7 +1,6 @@
 COMMIT=9121d1a5
 
 BUILD_VARIANT=STAGING
-AMPLITUDE_API_KEY=52a980fd5fb8da5fc680687d7e991e18
```
michalszorad, line 4: Instead of the `AMPLITUDE_API_KEY` are we using the config from ampli.json file?

apps/wallet-mobile/src/WalletNavigator.tsx
```diff
@@ -76,9 +75,7 @@ const WalletTabNavigator = () => {
           listeners={() => {
             return {
               focus: () => {
-                track({
-                  event: 'nft_click_navigate',
-                })
+                metrics.nftGalleryPageViewed({nft_count: 0})
```
michalszorad: This will need to be moved to `NFTs.tsx` into an `useEffect`

Eg

```tsx
useEffect(() => {
 if (isLoading) return;
  metrics.nftGalleryPageViewed({nft_count: nfts.length})
}, [isLoading, nfts]);
```

Eventually, `filteredNFTs` could be used, but this is to be discussed.

apps/wallet-mobile/src/metrics/README.md
```diff
@@ -0,0 +1,30 @@
+# Metrics
+
+Metrics feature uses Amplitude service to send analytics of the app.
+
+## Preparation
+
+To update the metrics with new Amplitude events, run script `metrics-pull`.
```
michalszorad, line 7: Can we add a note that the user may be prompted to log in and need to choose "us" server? At least this was my case

apps/wallet-mobile/src/metrics/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+import {features} from '../features'
+import {parseBoolean} from '../yoroi-wallets/utils'
+import {ampli} from './ampli'
+
+const featureFlag = features.analytics
+
+const environment = __DEV__
+  ? 'development'
+  : ({NIGHTLY: 'production', PROD: 'production', STAGING: 'staging', DEV: 'development'} as const)[
```
michalszorad: Small improvement 

```
const environmentMap = {
  DEV: 'development',
  PROD: 'production',
  STAGING: 'staging',
  NIGHTLY: 'production',
} as const
```

apps/wallet-mobile/src/metrics/index.ts
```diff
@@ -0,0 +1,64 @@
...
+      Config.BUILD_VARIANT ?? 'DEV'
+    ] ?? 'development'
+
+export const metrics = new Proxy(ampli, {
+  get(target, prop) {
+    const original = target[prop]
+    if (typeof original !== 'function' || prop === 'track' || prop === 'isInitializedAndEnabled') return original
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    return function (this: any, ...args: any[]) {
+      if (__DEV__ || !featureFlag) console.debug('[metrics-react-native] ', prop, ...args)
```
michalszorad: Maybe `Logger.info` would be better than `console.debug`. 

Do we need to log when the feature is disabled?

apps/wallet-mobile/src/metrics/index.ts
```diff
@@ -0,0 +1,64 @@
...
+const initMetrics = async () => {
+  if (metrics.isLoaded) return
+  const enabled = await isMetricsEnabled()
+  metrics.load({environment, client: {configuration: {optOut: !enabled, flushIntervalMillis: 5000}}})
+}
+
+const metricsStorageEnabledKey = 'metrics-enabled'
+export const isMetricsEnabled = async () => {
+  try {
+    return parseBoolean(await AsyncStorage.getItem(metricsStorageEnabledKey)) ?? __DEV__
```
michalszorad: I think we were trying to move away from `AsyncStorage` and use `YoroiStorage`

apps/wallet-mobile/src/YoroiApp.tsx
```diff
@@ -38,44 +36,39 @@ setLogLevel(CONFIG.LOG_LEVEL)
...
 
 const queryClient = new QueryClient()
-const amplitudeClient = getMetricsFactory(features.analytics ? 'amplitude' : 'mock')({
-  apiKey: Config.AMPLITUDE_API_KEY ?? '',
-})
-const metricsStorage = makeMetricsStorage()
 
 export const YoroiApp = () => {
   const migrated = useMigrations(storage)
+  const metrics = useInitMetrics()
```
michalszorad: Could we move `useInitMetrics` into `InitApp` component to have all initialization together?

michalszorad (commented):
I've added a few minor comments/questions. Otherwise, good job!

apps/wallet-mobile/.env
jorbuedo: Yes, `ampli.json` specifies the company/project it represents. The actual public keys to use per environment are added automatically to the generated file.

apps/wallet-mobile/src/WalletNavigator.tsx
jorbuedo: Yes! makes sense, let's discuss it and make the change with all the others.

apps/wallet-mobile/src/metrics/README.md
jorbuedo: Correct, that's where the amplitude servers for the account are. I'll add to the docs `us` has to be selected.

apps/wallet-mobile/src/metrics/index.ts
jorbuedo: To be decided, the package Juliano made before this refactor used debug. But there's no reason one way or another.
If it should console it or not based on the feature being disabled or just on dev, not sure. Depends on how you usually do it.

apps/wallet-mobile/src/metrics/index.ts
jorbuedo: Not sure, Juliano didn't mention anything. Is YoroiStorage mocked in storybook as well? What's the difference with just the native one?

michalszorad: I think the idea was to namespace settings using `YoroiStorage`. Eg, `appSettings/pin/`, `appSettings/installationId/`, `wallet/:id/memos/`. So, in this case, it could be `appSettings/metrics-enabled/`.  `YoroiStorage` uses `AsyncStorage` in the background and I believe should work out of the box in Storybook.

apps/wallet-mobile/src/Nfts/filterNfts.ts
```diff
@@ -6,3 +7,16 @@ export const filterNfts = (searchTerm: string, nfts: TokenInfo[]): TokenInfo[] =
     searchTermLowerCase.length > 0 ? nfts.filter((nft) => nft.name?.toLowerCase().includes(searchTermLowerCase)) : nfts
   return filteredNfts
 }
+
+let timeoutId: ReturnType<typeof setTimeout>
+let lastSearch = ''
+export const filterNftsMetrics = (searchTerm: string, resultCount: number) => {
```
michalszorad: I'm thinking there would be no need for global variables if we used the hook below. 

` if (!searchTerm) return;` may be needed, as we don't want to fire the event when someone clicks on the clear search button.

What are your thoughts?

```

const useTrackNftGallerySearchActivated = (searchTerm: string, resultCount: number) => {
  useEffect(() => {
   if (!searchTerm) return;
   const timeout = setTimeout(() => ampli.nftGallerySearchActivated({nft_search_term: searchTerm, nft_count: resultCount}), 500);
  
  return () => clearTimeout(timeout);
  }, [searchTerm, resultCount]);

}
```

michalszorad (approved):

michalszorad:
Looks good to me, good job 👍

apps/wallet-mobile/src/Nfts/filterNfts.ts
jorbuedo: My initial though would be that useEffects tend to rerun more than necessary.
And that those variables are not global. They are scoped to the file, which modern javascript treats as an independent module. So they can't be accessed anywhere else but that file.

I'm gonna test it but I think it will either run more than necessary, or the timeout won't clear when it should.

jorbuedo: Tested. It doesn't rerun for no reason, which on thinking now makes sense since string and number don't have the same issues as new objects that the other useEffect had.

But the debounce doesn't work. It calls every keystroke. You need the timeoutId outside of the function, otherwise it gets created every time and the clear is called on unmount, which doesn't really happen.

apps/wallet-mobile/src/YoroiApp.tsx
```diff
@@ -38,18 +38,16 @@ setLogLevel(CONFIG.LOG_LEVEL)
 if (Boolean(Config.DISABLE_LOGBOX)) LogBox.ignoreAllLogs()
 
 const queryClient = new QueryClient()
-const amplitudeClient = getMetricsFactory(features.analytics ? 'amplitude' : 'mock')({
-  apiKey: Config.AMPLITUDE_API_KEY ?? '',
-})
-const metricsStorage = makeMetricsStorage()
+
+const metricsManager = features.analytics ? makeMetricsManager() : mockMetricsManager()
```
stackchain, line 42: feature flag will go away eventually -> module for now is behind it, only the mock version is injected.

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,188 @@
...
+import Config from 'react-native-config'
+
+import {Logger} from '../yoroi-wallets/logging'
+import {storage, YoroiStorage} from '../yoroi-wallets/storage'
+import {parseBoolean} from '../yoroi-wallets/utils'
+import {Ampli, ampli} from './ampli'
+import {mockMetricsManager} from './mocks'
+
+const buildVariants = {
+  NIGHTLY: 'production',
```
stackchain, line 12: - [ ] Check with product -> `nightly` should have a specific view on amp?

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,188 @@
...
+  STAGING: 'staging',
+  DEV: 'development',
+} as const
+type MetricsEnv = (typeof buildVariants)[keyof typeof buildVariants]
+const currentBuildVariant = Config.BUILD_VARIANT ?? 'DEV'
+const environment: MetricsEnv = Object.keys(buildVariants).includes(currentBuildVariant)
+  ? buildVariants[currentBuildVariant]
+  : buildVariants.DEV
+
+export const makeMetricsStorage = (yoroiStorage: YoroiStorage = storage) => {
```
stackchain, line 33: colocate storage

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,188 @@
...
+    write: (enable: boolean) => settingsStorage.setItem(enabledKey, JSON.stringify(enable)),
+  } as const
+
+  return {
+    enabled,
+  } as const
+}
+export type MetricsStorage = ReturnType<typeof makeMetricsStorage>
+
+export const makeMetricsManager = (
```
stackchain, line 48: colocate manager

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,188 @@
...
+  } as const
+}
+
+export type MetricsManager = ReturnType<typeof makeMetricsManager>
+export type MetricsManagerContext = Omit<MetricsManager, 'enable' | 'disable'> &
+  Readonly<{
+    enable: () => void
+    disable: () => void
+  }>
+export type MetricsState = Readonly<{
```
stackchain, line 126: cache is handled by context so we don't need to await for the storage

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,188 @@
...
+  const init = () =>
+    enabled()
+      .then(
+        (isEnabled) =>
+          metricsModule.load({
+            environment,
+            client: {configuration: {optOut: !isEnabled, flushIntervalMillis: flushIntervalMs}},
+          }).promise,
+      )
+      .catch((e) => {
```
stackchain, line 83: avoid crashing in case ampli init fails

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,188 @@
...
+}
+type MetricsContextType = MetricsManagerContext & MetricsState
+
+const defaultState: MetricsState = {
+  isLoaded: false,
+  isEnabled: false,
+} as const
+const defaultActions: MetricsActions = {
+  isLoadedChanged: (_loaded: boolean) => console.error('[metrics-react] missing initialization'),
+  isEnabledChanged: (_enabled: boolean) => console.error('[metrics-react] missing initialization'),
```
stackchain:
```suggestion
  isLoadedChanged: (_loaded: boolean) => Logger.error('[metrics-react] missing initialization'),
  isEnabledChanged: (_enabled: boolean) => Logger.error('[metrics-react] missing initialization'),
```

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const {disable: managerDisable, enable: managerEnable} = metricsManager
+
+  const actions = React.useRef<MetricsActions>({
+    isLoadedChanged: (isLoaded) => dispatch({type: 'isLoadedChanged', isLoaded}),
+    isEnabledChanged: (isEnabled) => dispatch({type: 'isEnabledChanged', isEnabled}),
+  }).current
+
+  const disable = React.useCallback(() => {
+    actions.isEnabledChanged(false)
+    managerDisable()
```
stackchain, line 185: optimistic update

apps/wallet-mobile/src/metrics/metricsManager.tsx
```diff
@@ -0,0 +1,189 @@
...
+  }, [actions, managerDisable])
+  const enable = React.useCallback(() => {
+    actions.isEnabledChanged(true)
+    managerEnable()
```
stackchain, line 189: optimistic update

PR #2582, opened 2023-08-03
Swap - API integration

SorinC6:


packages/swap/src/translators/reactjs.tsx
```diff
@@ -131,6 +131,29 @@ export const useOrderByStatusOpen = (
...
+    ['persist:', 'usePairListByToken', string]
+  >,
+) => {
+  const {pairs} = useSwap()
+  const query = useQuery({
+    suspense: true,
+    ...options,
+    queryKey: ['persist:', 'usePairListByToken', tokenIdBase],
+    queryFn: () => pairs.list.byToken(tokenIdBase),
+  })
```
SorinC6, line 151: @stackchain not sure about the `tokenIdBase` , from what I see in the
` byToken: getTokens,`  . The `getTokens` function in the openswap is receiving `{policyId = '', assetName = ''} ` not a string but maybe I understand it wrong

PR #2584, opened 2023-08-03
E2e reporter test results

rahulnr7:
1) Create html report for test results
2) Add util functions to attach screenshots to the html report

https://emurgo.atlassian.net/browse/YOMO-684


michalszorad:
Hello @rahulnair87 it looks like the test is failing in `react-swap`, because it's trying to use package `@nicolo-ribaudo/semver-v6` which is not listed in package.json. If you add the dependency, it should resolve the issue.

rahulnr7:
will create a new PR as this was created based on an older branch

PR #2592, opened 2023-08-07
chore(tests): Add E2E NFT gallery tests

rahulnr7:
1) test for searching in gallery by nft name
2) add test IDs
3) add utils


apps/wallet-mobile/e2e/tests/_android/search-nft-gallery.test.ts
```diff
@@ -0,0 +1,51 @@
...
+       
+        const nftCount =await nftGalleryScreen.countNftsDisplayedAndroid()
+        await utils.takeScreenshot(`NFT Gallery !! Total number of NFTs : ${nftCount}` )
+    })    
+
+    it('should be able to seaarch for sample test NFT', async () => {
+        await nftGalleryScreen.iconSearch().tap()
+        await nftGalleryScreen.inputSearch().tap()
+        await nftGalleryScreen.inputSearch().typeText(nftToSearch)
+        await expect(nftGalleryScreen.cardNFT(nftToSearch)).toBeVisible()
```
Nebyt: This check will pass even the search will not work.
I mean search will not search anything and displayed nfts will stay the same as before.
I would suggest to add check for amount of found NFTs and after that check that the found nft is that we are looking for.

apps/wallet-mobile/e2e/tests/_ios/search-nft-gallery.test.ts
```diff
@@ -0,0 +1,49 @@
...
+        await walletMenuScreen.menuNFTGallery().tap()
+        await waitFor(nftGalleryScreen.iconSearch()).toBeVisible().withTimeout(10000)
+        await utils.takeScreenshot(`NFT Gallery !!` )
+    })    
+
+    it('should be able to seaarch for sample test NFT', async () => {
+        await nftGalleryScreen.iconSearch().tap()
+        await nftGalleryScreen.inputSearch().tap()
+        await nftGalleryScreen.inputSearch().typeText(nftToSearch)
+        await expect(nftGalleryScreen.cardNFT(nftToSearch)).toBeVisible()
```
Nebyt: The same remark as for android test.

Nebyt (changes requested):

apps/wallet-mobile/e2e/tests/_android/search-nft-gallery.test.ts
rahulnr7: I thought it will say 'No NFTs found' if no match, no? or is there a case when the search inouts does not result in any filtering?

Nebyt: > a case when the search inouts does not result in any filtering

This one

rahulnr7: new approach was adopted

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
```diff
@@ -0,0 +1,64 @@
...
+   let nftAttributes
+
+
+   while (!exit && i < max) {
+      try {
+         await expect(element(by.id(/^card_nft_[a-zA-Z0-9_ ]+$/)).atIndex(i)).toExist()
+         nftAttributes = await element(by.id(/^card_nft_[a-zA-Z0-9_ ]+$/)).atIndex(i).getAttributes()
+         if (nftAttributes.label === nftLabel) {
+            matched = true
+            exit = true
```
Nebyt: It can be replaced with `return true`

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
```diff
@@ -0,0 +1,64 @@
...
+         }
+         i++
+      } catch (e) { // stop checking for any matches on any failures
+         exit = true
```
Nebyt: It can be replaced `return false`

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
```diff
@@ -0,0 +1,64 @@
...
+      }
+   }
+   if (matched) { console.log(' Match found : ', nftAttributes.label) }
+
+   return matched
```
Nebyt: This part can be removed after adding `return` statement before.
I would suggest to use a proper logging tool into file instead of console logging.

Nebyt (changes requested):

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
rahulnr7: Done. Yes, we have one to add it to the html report - addMsgToReport() in utils.

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
rahulnr7: Done

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
rahulnr7: Done

Nebyt (approved):

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+   }
+   await utils.addMsgToReport(`Total Number of NFTs counted: ${i}`)
+   return i
+}
+
+export const checkAttributeNFTAndroid = async (nftLabel: string, max = 10) => {
+   let i = 0
+
+   /** @type {Detox.ElementAttributes} */
```
michalszorad: This type annotation can be removed

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
```diff
@@ -0,0 +1,90 @@
...
+   let nftAttributes
+
+   while (i < max) {
+      try {
+         await expect(element(by.id(/^card_nft_[a-zA-Z0-9_ ]+$/)).atIndex(i)).toExist()
+         nftAttributes = await element(by.id(/^card_nft_[a-zA-Z0-9_ ]+$/)).atIndex(i).getAttributes()
```
michalszorad: We won't have to define the `nftAttributes` type if we use const inside the loop

```
const nftAttributes = ...
if ('label' in nftAttributes) {
 // ... 
}
```

apps/wallet-mobile/e2e/tests/_android/search-nft-gallery.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        await expect(walletMenuScreen.menuNFTGallery()).toBeVisible()
+        await utils.takeScreenshot('Wallet 2 Home screen')
+        
+    })    
+
+    it('should be able to open "NFT Gallery" and verify the total count of NFTs displayed', async () => {
+        await walletMenuScreen.menuNFTGallery().tap()
+        await waitFor(nftGalleryScreen.iconSearch()).toBeVisible().withTimeout(10000)
+       
+        const nftCount =await nftGalleryScreen.countNftsDisplayedAndroid()
```
michalszorad: It looks like the file is not formatted

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
```diff
@@ -0,0 +1,90 @@
...
+export const countNftsDisplayedAndroid = async (max = 10) => {
+   let i = 0
+   let exit = false
+
+   while (!exit && i < max) {
+      try {
+         await expect(element(by.id(/^card_nft_[a-zA-Z0-9_ ]+$/)).atIndex(i)).toExist()
+         i++
+      } catch (e) { // stop checking for any matches on all failures
+         exit = true
```
michalszorad: would a `break` instead of `exit = true` work here?

apps/wallet-mobile/e2e/screens/devOptionsScreen.screen.ts
```diff
@@ -0,0 +1,6 @@
+import {by, element} from 'detox'
+
+export const buttonRestoreWallet1 = () => element(by.id('btnRestoreWallet1'))
+export const buttonRestoreWallet2 = () => element(by.id('btnRestoreWallet2'))
+export const buttonRestoreWallet3 = () => element(by.id('btnRestoreWallet3'))
+export const btnResetAnalyticsConsent = () => element(by.id('btnResetAnalyticsConsent'))
```
michalszorad: I see sometimes we use `button` and sometimes `btn`. Would it be better to use always one?

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
```diff
@@ -0,0 +1,90 @@
...
+
+import * as utils from '../utils'
+
+export const iconSearch = () => element(by.id('iconSearch'))
+export const inputSearch = () => element(by.id('inputSearch'))
+
+export const cardNFT = (nftName: string) => element(by.id(`card_nft_${nftName}`))
+export const txtNftCount = () => element(by.id('txtNftCount'))
+
+export const countNftsDisplayedAndroid = async (max = 10) => {
```
michalszorad, line 12: Why is there a different function for Android and iOS? Is it possible to reuse one function?

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
rahulnr7: Done. Thanks!

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
rahulnr7: Done, Thanks

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
rahulnr7: Done, Thanks

apps/wallet-mobile/e2e/tests/_android/search-nft-gallery.test.ts
rahulnr7: Done, Thanks

apps/wallet-mobile/e2e/screens/devOptionsScreen.screen.ts
rahulnr7: Done, Thanks

rahulnr7: Yes will stick to using button and camel case

apps/wallet-mobile/e2e/screens/nftGallery.screen.ts
rahulnr7: Detox behaves differently for iOS and Android in certain instances(not fully evolved for Android yet). 
In this case, if we try to find the matching elements for a locator that is a regular expression, then for iOS, the getAttributes() returns an array of objects while for Android it fails with multiple matching elements found error. 
So for Android the way to find the total number of matching elements will be to index the returned value iteratively till it fails to find next one.

michalszorad (approved):

PR #2631, opened 2023-08-21
refactor: Started extracting wallets - storage

stackchain:


apps/wallet-mobile/src/yoroi-wallets/cardano/nfts.ts
```diff
@@ -1,13 +1,13 @@
...
 
 import {features} from '../../features'
 import {getAssetFingerprint} from '../../legacy/format'
-import {createTypeGuardFromSchema, isArrayOfType, isString} from '../utils'
 import {utf8ToHex} from './api/utils'
-
 export const convertNft = (options: {
-  metadata?: unknown
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  metadata?: any
```
michalszorad, line 10: Why use `any` instead of `unknown`?

apps/wallet-mobile/src/yoroi-wallets/cardano/shelley/ShelleyWallet.ts
```diff
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import {Balance} from '@yoroi/types'
+import {App, Balance} from '@yoroi/types'
+import {parseSafe} from '@yoroi/wallets'
```
michalszorad, line 3: I've noticed multiple parsing functions are imported from `@yoroi/wallets`. Should we consider moving parsing logic out of `@yoroi/wallets` in the future and have something like `@yoroi/parsing`?

apps/wallet-mobile/src/yoroi-wallets/cardano/nfts.ts
stackchain: basically to not cast once, I've added a generic to the TokenInfo, but didn't want to update in this PR.

apps/wallet-mobile/src/yoroi-wallets/cardano/shelley/ShelleyWallet.ts
stackchain: everything that is inside `/yoroi-wallets` will be moved into `yoroi/wallets` eventually

banklesss (approved):

michalszorad (approved):

PR #2635, opened 2023-08-22
feature: Banxa module

stackchain:


packages/banxa/package.json
```diff
@@ -0,0 +1,209 @@
...
+  },
+  "license": "Apache-2.0",
+  "author": "EMURGO Fintech <support@emurgo.com> (https://github.com/Emurgo/yoroi)",
+  "contributors": [
+    {
+      "name": "Juliano Lazzarotto",
+      "email": "30806844+stackchain@users.noreply.github.com"
+    }
+  ],
+  "main": "lib/commonjs/index",
```
banklesss, line 30: A local build is needed to be able to use the package

banklesss (approved):

michalszorad (approved):

packages/banxa/package.json
stackchain: yap, doing that for all now, cuz it needs to resolve the package dependency order, I've added lerna that does so. There is a new script on the root that bootstrap everything and watch for changes, so while working on the packages it will watch and rebuild pkgs in the right order.

PR #2636, opened 2023-08-23
feature: Banxa referral redirect

stackchain:


apps/wallet-mobile/index.js
```diff
@@ -1,4 +1,5 @@
 import './polyfills'
+import 'react-native-url-polyfill/auto'
```
stackchain, line 2: required for `URL` + `URLSearchParams`

apps/wallet-mobile/src/TxHistory/ActionsBanner.tsx
```diff
@@ -22,7 +23,20 @@ export const ActionsBanner = ({disabled = false}: {disabled: boolean}) => {
...
+    const url = banxa.createReferralUrl({
+      coinType: 'ADA',
+      fiatType: 'USD',
+      blockchain: 'ADA',
+      walletAddress,
+    })
+    Linking.openURL(url.toString())
+  }
+
+  const handleOnSend = () => {
```
banklesss, line 39: `onPressSend`/`onPressBuy` ? or `handleOnBuy`/`handleOnSend`?

banklesss: and from now on we can follow the same pattern, because I have seen both types in the repository

banklesss (approved):

apps/wallet-mobile/src/TxHistory/ActionsBanner.tsx
stackchain: we usually stick with `onPress` to keep consistency, yet, when as props we use event name `on{Action,Buy,Sell}` in case `onPress` doesn't "fit", we use `handleXXX` as internal functionals, and never as props.

PR #2654, opened 2023-08-31
fix (swap)

SorinC6:


packages/swap/src/adapters/transformers.ts
```diff
@@ -113,13 +113,16 @@ export const asYoroiAmount = (openswapAmount: {
...
-  const {amount, token} = openswapAmount
-  const [assetName = '', policyId = ''] = token.split('.')
-  const subject = `${assetName}.${policyId}`
-  return {
-    quantity: amount ? (amount as Balance.Quantity) : '0',
-    tokenId: subject.length === 1 ? '' : subject,
-  } as const
+  if (openswapAmount?.amount) {
+    const {amount, token} = openswapAmount
+    const [assetName = '', policyId = ''] = token.split('.')
```
stackchain: is it in the right order? usually is `policyId.assetName` just wondering

packages/swap/src/adapters/transformers.ts
```diff
@@ -113,13 +113,16 @@ export const asYoroiAmount = (openswapAmount: {
...
+    const subject = `${assetName}.${policyId}`
+    return {
+      quantity: amount ? (amount as Balance.Quantity) : '0',
```
stackchain: there is a helper `asQuantity` that will do it.

packages/swap/src/adapters/transformers.ts
SorinC6: will change the order 😁

packages/swap/src/adapters/transformers.ts
SorinC6: will need to bring the helper into the package 🤔

jorbuedo (approved):

packages/swap/src/adapters/transformers.ts
stackchain: aaaa it's not exposed yet in the `yoroi/wallets`

stackchain (approved):

PR #2661, opened 2023-09-05
fix(swap): Fix asset selection list bugs

jorbuedo:
related:

- https://emurgo.atlassian.net/browse/YOMO-760
- https://emurgo.atlassian.net/browse/YOMO-761


SorinC6 (commented):
@jorbuedo when accessing the token list I still can see a spinner https://recordit.co/ZDwyozO2p7 and the skeleton loader after but for a very short time 🤔

jorbuedo:
> @jorbuedo when accessing the token list I still can see a spinner https://recordit.co/ZDwyozO2p7 and the skeleton loader after but for a very short time 🤔

Yes, I didn't change the main spinner, wasn't sure if that was part of the design or not. That one retrieves the list, then the individual ones each get the tokenInfo use the skeleton design.
Maybe I could substitute the spinner for a fixed array of skeletons

SorinC6:
> > @jorbuedo when accessing the token list I still can see a spinner https://recordit.co/ZDwyozO2p7 and the skeleton loader after but for a very short time 🤔
> 
> Yes, I didn't change the main spinner, wasn't sure if that was part of the design or not. That one retrieves the list, then the individual ones each get the tokenInfo use the skeleton design. Maybe I could substitute the spinner for a fixed array of skeletons

@jorbuedo  can you check with Rahul how the expected behaviour will be. I think they expect to see only the skeleton loading there but not 100% sure of it

apps/wallet-mobile/src/components/AmountItem/AmountItem.tsx
```diff
@@ -8,7 +8,7 @@ import {isEmptyString} from '../../utils'
 import {YoroiWallet} from '../../yoroi-wallets/cardano/types'
 import {useTokenInfo} from '../../yoroi-wallets/hooks'
 import {Quantities} from '../../yoroi-wallets/utils'
-import {Boundary, Icon, Placeholder, Spacer, Text, TokenIcon} from '..'
+import {Boundary, Icon, Spacer, Text, TokenIcon, TokenIconPlaceholder} from '..'
```
stackchain, line 11: Skeletons?

jorbuedo: What do you mean? The name of the Placeholder/Skeleton?

TokenIcon was exporting Placeholder to the global component scope, which was very generic, so I renamed it TokenIconPlaceholder. And created a different one for the AmountItemPlaceholder

Called them placeholders instead of skeletons because that's the name it had before

jorbuedo:
> > > @jorbuedo when accessing the token list I still can see a spinner https://recordit.co/ZDwyozO2p7 and the skeleton loader after but for a very short time 🤔
> > 
> > 
> > Yes, I didn't change the main spinner, wasn't sure if that was part of the design or not. That one retrieves the list, then the individual ones each get the tokenInfo use the skeleton design. Maybe I could substitute the spinner for a fixed array of skeletons
> 
> @jorbuedo can you check with Rahul how the expected behaviour will be. I think they expect to see only the skeleton loading there but not 100% sure of it

I agree it should be just skeletons, it just can't be the same skeletons. At least how it's structured now. There's 2 loading boundries, one for the initial load, then for each entry. I've put a list of skeletons now. But the header of the list doesn't show until the list is loaded. I don't think it's too much of a problem.

apps/wallet-mobile/src/components/AmountItem/AmountItem.tsx
stackchain: I think `TokenIconPlaceholder` is used to replace the original image/logo when not present, `Amount` looks like is used only as fallback's boundary so I'd say he is the candidate.

stackchain:
> > > > @jorbuedo when accessing the token list I still can see a spinner https://recordit.co/ZDwyozO2p7 and the skeleton loader after but for a very short time 🤔
> > > 
> > > 
> > > Yes, I didn't change the main spinner, wasn't sure if that was part of the design or not. That one retrieves the list, then the individual ones each get the tokenInfo use the skeleton design. Maybe I could substitute the spinner for a fixed array of skeletons
> > 
> > 
> > @jorbuedo can you check with Rahul how the expected behaviour will be. I think they expect to see only the skeleton loading there but not 100% sure of it
> 
> I agree it should be just skeletons, it just can't be the same skeletons. At least how it's structured now. There's 2 loading boundries, one for the initial load, then for each entry. I've put a list of skeletons now. But the header of the list doesn't show until the list is loaded. I don't think it's too much of a problem.

It's simpler to reproduce it on the `stories.` by injecting timed api responses, design the screen state there, so we can present to QA/Product, and ask for confirmation

jorbuedo:
> It's simpler to reproduce it on the `stories.` by injecting timed api responses, design the screen state there, so we can present to QA/Product, and ask for confirmation

It is there, I've added fetching state to both. I'm trying to figure out why the `sell` one doesn't work though, something with the suspense I think. The `buy` one shows it correctly (more or less). Looking into it

SorinC6:
> > It's simpler to reproduce it on the `stories.` by injecting timed api responses, design the screen state there, so we can present to QA/Product, and ask for confirmation
> 
> It is there, I've added fetching state to both. I'm trying to figure out why the `sell` one doesn't work though, something with the suspense I think. The `buy` one shows it correctly

The `Boundary` component is adding the loading spinner as well when fetching - maybe got something to do with it as well

stackchain:
> > > It's simpler to reproduce it on the `stories.` by injecting timed api responses, design the screen state there, so we can present to QA/Product, and ask for confirmation
> > 
> > 
> > It is there, I've added fetching state to both. I'm trying to figure out why the `sell` one doesn't work though, something with the suspense I think. The `buy` one shows it correctly
> 
> The `Boundary` component is adding the loading spinner as well when fetching - maybe got something to do with it as well

I'd suggest to download Flipper, it allows to see the react tree, then you can toggle suspense on any component and it will show you which of the `Boundary` is kicking it.

stackchain:
> > It's simpler to reproduce it on the `stories.` by injecting timed api responses, design the screen state there, so we can present to QA/Product, and ask for confirmation
> 
> It is there, I've added fetching state to both. I'm trying to figure out why the `sell` one doesn't work though, something with the suspense I think. The `buy` one shows it correctly (more or less). Looking into it

I remember @SorinC6 mocked some responses but dunno.

jorbuedo:
@SorinC6 @stackchain Well they work differently because one has `usePairListByToken` and the other doesn't. There's a parade of hooks calling each other, but only the hooks that actually use useQuery directly, and have suspense to true, will suspend.
Both suspend for the individual elements of the list, but on the parent level, only `buy` suspends because of the `usePairListByToken` . `sell` is briefly on a state of empty list instead, while the parade of hooks figure out that the list is indeed not empty.
I think current state is stable enough to merge. Even if it has some quirk, doesn't really look wrong. To have proper bondaries working means refactoring those hooks.

SorinC6 (approved):

banklesss (approved):

PR #2673, opened 2023-09-11
fix(swap): (auto) should be shown only for the default liquidity pool

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-772


apps/wallet-mobile/src/features/Swap/common/SelectPool/SelectPoolFromList/SelectPoolFromList.tsx
```diff
@@ -25,6 +27,7 @@ export const SelectPoolFromList = ({data = []}: Props) => {
     track.swapPoolChanged()
     selectedPoolChanged(pool)
     setSelectedCardIndex(pool.poolId)
+    poolTouched()
```
michalszorad, line 30: Will it be triggered even if I select the default pool?

stackchain: I think this is correct, cuz you touched the input, then it's not auto anymore.

apps/wallet-mobile/src/features/Swap/common/SelectPool/SelectPoolFromList/SelectPoolFromList.tsx
```diff
@@ -7,6 +7,7 @@ import LinearGradient from 'react-native-linear-gradient'
 import {Spacer} from '../../../../../components'
 import {useMetrics} from '../../../../../metrics/metricsManager'
 import {COLORS} from '../../../../../theme'
+import {useSwapTouched} from '../../../useCases/StartSwapScreen/CreateOrder/TouchedContext'
```
stackchain: if is used by more than 1 move to `common`. (-> `TouchedContext` -> `common/SwapFormProvider`)

apps/wallet-mobile/src/i18n/locales/en-US.json
```diff
@@ -149,6 +149,7 @@
   "swap.swapScreen.noAssetsFoundFor": "No assets found for \"{search}\"",
   "swap.swapScreen.defaultSlippage": "Default Slippage Tolerance",
   "swap.swapScreen.slippageInfo": "Slippage tolerance is set as a percentage of the total swap value. Your transactions will not be executed if the price moves by more than this amount.",
+  "swap.swapScreen.autoPool": " (auto)",
```
stackchain: add the space on the screen `${x} ${strings.xx}`

stackchain (changes requested):
minor CR

apps/wallet-mobile/src/features/Swap/common/SelectPool/SelectPoolFromList/SelectPoolFromList.tsx
jorbuedo: Yes, if you select the default pool then it's not selected because it's default, but because you selected it.

stackchain (approved):

PR #2678, opened 2023-09-11
fix(swap): Set pool auto when changing pool list

jorbuedo:
Missed this case before for https://emurgo.atlassian.net/browse/YOMO-772


apps/wallet-mobile/src/features/Swap/common/SwapFormProvider.tsx
```diff
@@ -10,6 +10,7 @@ type TouchedActions = {
   buyTouched: () => void
   switchTouched: () => void
   poolTouched: () => void
+  poolAuto: () => void
```
michalszorad: Most of these actions end with a verb, but `poolAuto` does not. Would it make sense to change it to `poolAutoChanged` or something similar?

michalszorad (approved):

PR #2681, opened 2023-09-12
fix(swap): Remove verified filter, API returns only verified ones

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-759


SorinC6 (approved):
@jorbuedo have you checked with product related to removing it? maybe if they will want to show it later we can just hide it until adding our own filters as well 🤔

jorbuedo:
> @jorbuedo have you checked with product related to removing it? maybe if they will want to show it later we can just hide it until adding our own filters as well 🤔

Waiting for answer

jorbuedo:
Looks like no one knows and it's gonna take a while, so I'm adding the `verified` filter to the list just in case the API changes suddenly again to include unverified ones. But assume that's not gonna be the case, so no UI switch to toggle.

stackchain:
in-fact is returning only verified, if they changed it later, we can revert the commit

PR #2685, opened 2023-09-13
fix(swap): Buy value recalculation race condition

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-762

I don't love the solution, not sure if it may break for slow connection either. 
Problem is recalculation of fields happens in the form, instead of in the swap package.

When buyTokenInfo?.id changes, I want to recalculate its value based on the sell quantity. But when running immediately, the createOrder values are not yet updated to do the calculation.

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditSellAmount/EditSellAmount.tsx
+++ b/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditSellAmount/EditSellAmount.tsx
@@ -68,7 +68,7 @@ export const EditSellAmount = () => {
   }
 
   React.useEffect(() => {
-    recalculateBuyValue(quantity)
+    setTimeout(() => recalculateBuyValue(quantity), 100)
     // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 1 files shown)


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditSellAmount/EditSellAmount.tsx
```diff
@@ -68,7 +68,7 @@ export const EditSellAmount = () => {
   }
 
   React.useEffect(() => {
-    recalculateBuyValue(quantity)
+    setTimeout(() => recalculateBuyValue(quantity), 100)
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [buyTokenInfo?.id])
```
michalszorad: If we add a `createOrder?.selectedPool` and `quantity` to the dependency array, does it help?

jorbuedo: That... makes sense 💯

stackchain (approved):

PR #2694, opened 2023-09-19
fixes(swap) fees calculation, slippage error, market price

SorinC6:
task:
- add total fees calculation
- reset market price when refresh swap
- add info modal for pool actions card 
- move helpers in swap package
- https://emurgo.atlassian.net/browse/YOMO-786
- https://emurgo.atlassian.net/browse/YOMO-793
- https://emurgo.atlassian.net/browse/YOMO-794


apps/wallet-mobile/src/features/Swap/useCases/ConfirmTxScreen/TransactionSummary.tsx
```diff
@@ -39,12 +41,22 @@ export const TransactionSummary = () => {
...
     },
     {
       label: strings.swapFeesTitle,
-      value: `${poolFee} ADA`,
+      value: `${getTotalFees(
+        selectedPool?.batcherFee.quantity,
+        asQuantity(poolFee),
+        wallet.primaryTokenInfo.decimals ?? 0,
+        numberLocale,
+      )} ADA`,
```
stackchain: `ADA` | `TADA` should come from the `wallet.primaryTokenInfo.ticker`

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -115,7 +115,7 @@ export const CreateOrder = () => {
       buy: amounts.buy,
       pools: poolList,
       selectedPool: createOrder.selectedPool,
-      slippage: createOrder.slippage,
+      slippage: createOrder.slippage * 100,
       address: addresses.used[0],
```
stackchain: if we want to always send the same address I'd say we should go for `wallet.externalChain[0]`, you don't necessarily consume the addresses in order, the first used can be the index 1/2/3, we don't know.

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditLimitPrice.tsx
```diff
@@ -42,29 +40,30 @@ export const EditLimitPrice = () => {
...
     const sellQuantityDenominated = Quantities.denominated(
       createOrder.amounts.sell.quantity,
       sellTokenInfo.decimals ?? 0,
     )
     buyAmountChanged({
-      quantity: getBuyQuantityForLimitOrder(sellQuantityDenominated, `${defaultPrice}`, buyTokenInfo.decimals ?? 0),
+      quantity: getBuyQuantityForLimitOrder(
+        sellQuantityDenominated,
+        asQuantity(defaultPrice ?? 0),
+        buyTokenInfo.decimals ?? 0,
```
stackchain, line 52: hmm, if pieces of data from the tokenInfo (decimals) are dependencies for the helpers to calculate the pair, it will make the offload of these functions to the module a little more hard, maybe instead of the amounts in the state, for the token, we should have also the tokenInfo or maybe it can be simplified and attach only wha t is actually needed if its only the decimals

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditPool/ShowPoolActions.tsx
```diff
@@ -35,7 +38,7 @@ export const ShowPoolActions = () => {
   const totalAmount = Quantities.format(amounts.buy.quantity, buyTokenInfo.decimals ?? 0)
   const protocolCapitalize = selectedPool.provider[0].toUpperCase() + selectedPool.provider.substring(1)
```
stackchain: can we avoid index accessing? maybe lodash has a function to handle capitalize already.

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/ShowTokenActions/TopTokenActions.tsx
```diff
@@ -1,36 +1,61 @@
...
+    } else {
+      orderTypeChanged('limit')
+    }
+  }
+
   const handleSync = () => {
+    orderTypeChanged('market')
     refetch()
+    if (poolList !== undefined) {
+      const bestPool = poolList.map((a) => a).sort((a, b) => a.price - b.price)[0]
```
stackchain, line 39: avoid index accessing -> `.find(() => true)` | `const [bestPool, _] = poolList.` it also looks like this is a helper for the package `findPoolByBestPrice(poolList)`

packages/swap/src/helpers/orders/getMinAdaReceiveAfterSlippage.ts
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+export const getMinAdaReceiveAfterSlippage = (
+  outputAmount: Balance.Quantity,
+  slippagePercentage: number,
+  decimals: number,
+  numberLocale: NumberLocale,
+): string => {
+  const slippageDecimal = slippagePercentage / 100
+  const result = Number(outputAmount) / (1 + slippageDecimal)
```
stackchain, line 23: `.Quantity` can hold big numbers, using `Number()` can lose precision, should it be `BigNumber` or Number is safe for this first parameter?, cuz if uncertain, `BN(outputAmount).div` ....

packages/swap/src/helpers/orders/getMinAdaReceiveAfterSlippage.test.ts
```diff
@@ -0,0 +1,44 @@
+import {asQuantity} from '../../utils/asQuantity'
+import {getMinAdaReceiveAfterSlippage} from './getMinAdaReceiveAfterSlippage'
+
+const mocknumberLocale = {
```
stackchain: maybe it can be moved into `adapters/intl/number-locale.mocks.ts` suggestion.

packages/swap/src/helpers/orders/getTotalFees.test.ts
```diff
@@ -0,0 +1,57 @@
+import {asQuantity} from '../../utils/asQuantity'
+import {getTotalFees} from './getTotalFees'
+
+const mocknumberLocale = {
```
stackchain: :arrow_up:

packages/swap/src/helpers/orders/getTotalFees.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  decimalSeparator: '.',
+  fractionGroupSeparator: ' ',
+  fractionGroupSize: 0,
+  groupSeparator: ',',
+  groupSize: 3,
+  prefix: '',
+  secondaryGroupSize: 0,
+  suffix: '',
+}
```
stackchain:
```suggestion
} as const
```

packages/swap/src/helpers/orders/getTotalFees.ts
```diff
@@ -0,0 +1,30 @@
...
+  proiderFee: Balance.Quantity,
+  decimals: number,
+  numberLocale: NumberLocale,
+): string => {
+  const result = Quantities.denominated(
+    Quantities.sum([batcherFee, proiderFee]),
+    decimals,
+  )
+
+  console.log('getTotalFees result', result)
```
stackchain:
```suggestion
```

packages/swap/src/helpers/orders/getMinAdaReceiveAfterSlippage.ts
jorbuedo: @stackchain I've seen in the package we've not make use of BigNumber, using BigInt (native one) instead. Is there a reason behind that?

packages/types/src/helpers/types.ts
```diff
@@ -1 +1,12 @@
 export type Nullable<T> = T | null | undefined
+
+export type NumberLocale = {
```
stackchain: maybe `src/intl/numbers.ts`

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
SorinC6: I can't find the `externalChain ` type in `YoroiWallet` 🤔 should we go with `wallet.externalAddresses[0]` ?

stackchain: wow sorry, yap, externalAddress[0] (here index accessing is safe)

stackchain: still `const [firstExtAddress, ] = wallet.externalAddresses`

packages/swap/src/helpers/orders/getMinAdaReceiveAfterSlippage.ts
stackchain: no, it was carried from old dex module, I've read that current `hermes` is supporting it, so maybe we can ditch BN slowly, still `BigInt` has some limitations, such as `JSON.strin..` commands can't serialized/deserialize it, as long as the in-outs are strings, for now looks safe to use internally, maybe we should add some edge cases on the client side for testing it.

stackchain: https://github.com/facebook/hermes/issues/510

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditLimitPrice.tsx
SorinC6: I see that functions `getBuyQuantityForLimitOrder ` and `getSellQuantityForLimitOrder` are used only on the frontend 🤔 once Jordi moves the logic of the inputs math values calculation in the swap module I think this will be moved in the swap package - will double-check with him

packages/swap/src/helpers/orders/getMinAdaReceiveAfterSlippage.ts
SorinC6: have used `BigInt` as I saw that is used for the other functions in the package. Yes! seems like we can use it as well from what I see

stackchain (approved):

PR #2698, opened 2023-09-20
chore: wallets->common + cardano

stackchain:
- [x] added Yoroi workspace (for those using vscode)
- [x] migrated Cardano stuff into Cardano package (not-in-use) for the next iteration 
- [x] migrated some functions from wallets->common
- [x] covered the new packages (common/cardano) with 100% of tests
- [x] added jest to openswap pkg (tests were randomly failing) cache issues
- [x] moving to ISA some updates on async storage


banklesss (approved):

packages/cardano/src/cardano/api/parsers.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    expect(isFtMetadata(true)).toBe(false)
+  })
+
+  it('should return false if name is not a string or is missing', () => {
+    expect(isFtMetadata({name: 123})).toBe(false)
+    expect(isFtMetadata({extra: 123})).toBe(false)
+  })
+
+  it('should return false if description exists but is not a string', () => {
+    expect(isFtMetadata({name: 'some-name', description: 123})).toBe(false)
```
michalszorad, line 94: Shouldn't we in this case, treat is valid, but only matching `{name: 'some-name'}` and ignore all invalid properties?

michalszorad: Also applies in all similar cases

packages/cardano/src/cardano/api/token-onchain-metadata.ts
```diff
@@ -0,0 +1,186 @@
...
+    possibleNftMetadataRecord !== undefined &&
+    isRecord(possibleNftMetadataRecord)
+  ) {
+    const parsedNftMetadataRecord = parseNftMetadataRecord(
+      possibleNftMetadataRecord,
+    )
+    if (parsedNftMetadataRecord !== undefined) {
+      return {
+        // record holds original tx mint metadata is safe to cast here
+        mintNftMetadata: record as ApiNftMetadataRecord,
```
michalszorad, line 151: If the record holds original TX mint metadata, it should not be safe to cast, right? The metadata on the blockchain can contain anything and don't need to follow any community recommendations.

michalszorad: Same as in `mintFtMetadata: record as ApiFtMetadataRecord,`

packages/cardano/src/cardano/api/token-supply.ts
```diff
@@ -0,0 +1,66 @@
...
+  supplies: z.record(z.number().nullable()),
+})
+
+export const isTokenSupplyResponse =
+  createTypeGuardFromSchema(TokenSupplySchema)
+
+export const parseTokenSupplyResponse = (
+  data: unknown,
+): TokenSupplyResponse | undefined => {
+  return isTokenSupplyResponse(data) ? (data as TokenSupplyResponse) : undefined
```
michalszorad: It should work without the type cast

packages/cardano/src/cardano/api/token-supply.ts
```diff
@@ -0,0 +1,66 @@
...
+      url: `${baseUrl}/multiAsset/supply`,
+      data: payload,
+      method: 'POST',
+      headers: {'Content-Type': 'application/json'},
+    }).then((response) => {
+      const parsedResponse = parseTokenSupplyResponse(response)
+      if (!parsedResponse)
+        return Promise.reject(new Error('Invalid asset supplies'))
+
+      const result: any = {} // need any here TS issues with key indexing
```
michalszorad: Does `  const result: Record<string, ApiTokenSupplyRecord> = {}` work?

packages/cardano/src/cardano/api/token-supply.ts
```diff
@@ -0,0 +1,66 @@
...
+      const supplies: Record<ApiTokenId, ApiTokenSupplyRecord | undefined> =
+        parsedResponse.supplies
+
+      Array.from(assetsMap.entries()).forEach(([id, {policy, name}]) => {
+        const tokenId: ApiTokenId = `${policy}.${name}`
+        result[id] = supplies[tokenId] ?? null
+      })
+
+      return Promise.resolve(result as ApiTokenSupplyResponse) // casting here
```
michalszorad: If `const result: Record<string, ApiTokenSupplyRecord> = {}` works, then we can remove this type cast as well

packages/cardano/src/cardano/api/parsers.ts
```diff
@@ -0,0 +1,70 @@
...
+  if (!isString(file.mediaType)) return false
+  if (!isStringOrArrayOfString(file.src)) return false
+
+  return true
+}
+
+export function isNftMetadata(data: unknown): data is ApiNftMetadata {
+  if (!isRecord(data)) return false
+
+  const metadata = data as Partial<ApiNftMetadata>
```
michalszorad, line 25: type cast can be removed

packages/cardano/src/cardano/api/parsers.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  if (!isString(metadata.name)) return false
+  if (!isStringOrArrayOfString(metadata.image)) return false
+  if (metadata.mediaType && !isString(metadata.mediaType)) return false
+  if (metadata.description && !isStringOrArrayOfString(metadata.description))
+    return false
+
+  if (
+    metadata.files &&
```
michalszorad, line 34: if `metadata.files` is `null` (or `0`, etc) then `isNftMetadata` will resolve to `true` while still having invalid type, but should resolve to `false`, right?

packages/cardano/src/cardano/api/parsers.ts
```diff
@@ -0,0 +1,70 @@
...
+  )
+    return false
+
+  return true
+}
+
+export function isFtMetadata(data: unknown): data is ApiFtMetadata {
+  if (!isRecord(data)) return false
+
+  const metadata = data as Partial<ApiFtMetadata>
```
michalszorad, line 46: typecast can be removed

packages/cardano/src/cardano/api/parsers.test.ts
stackchain: do you know if is it possible with `zod` to make it loose? but in that case I'd only use this feature if we can provide a fallback if that property didn't pass i.e `string` -> is `123` (invalid) defaulting to '' in the schema.

packages/cardano/src/cardano/api/token-onchain-metadata.ts
stackchain: the way it goes is, it validates the CIPs fields, and everything else would be falling into `{[string, unknown]}` https://github.com/Emurgo/yoroi/pull/2698/files/b5774a45b5eddcf086077aa3f9eb49e850a0f17c#diff-cca0d9e6e533dd4a20a4aa8d03d3eff64262ccbaa06c0a2f3b794c936ad48b1fR18

stackchain: so `mint***` is the original metadata (whatever it is) and the `selected` only exists if by parsing the expected CIP format did pass.

packages/cardano/src/cardano/api/token-supply.ts
stackchain: same issue when you don't have explicit keys

packages/cardano/src/cardano/api/token-supply.ts
stackchain:
```suggestion
  return isTokenSupplyResponse(data) ? data : undefined
```

packages/cardano/src/cardano/api/token-supply.ts
stackchain: the expected output is `Record<ApiTokenId, ApiTokenSupplyRecord>` if we use `Record<string...` it will work for the 1st complain, but is not the same type of the return `Record<ApiTokenId..` (will have to cast anyway again) no idea why TS can't handle typed keys.

stackchain: I've found the issue, the mapping was declared as `string`, and not as `ApiTokenId`.

packages/cardano/src/cardano/api/parsers.test.ts
michalszorad: Yes, there seems to be an API supporting this behavior.However both `.default()` and `.catch()` may need to be added

```
var zod = require("zod")

var schema = zod.object({name: zod.string(), age: zod.number().optional().default(0).catch(0)
})

const parsed = schema.safeParse({name: 'Michal', age: '123'});
console.log(parsed) // {success: true, data: {name: "Michal", age: 0}}
```

packages/cardano/src/cardano/api/parsers.ts
stackchain: this will cause metadata: unknown, TS doesn't know that !isRecords, make data a record. :/

packages/cardano/src/cardano/api/parsers.test.ts
michalszorad: Or maybe `var schema = zod.object({name: zod.string(), age: zod.number().catch(0)})` will work as well -> `.optional` and `.default` is removed

packages/cardano/src/cardano/api/parsers.ts
stackchain: this will cause metadata: unknown
!isRecord -> doesn't turn data into a Record<>

packages/cardano/src/cardano/api/parsers.ts
stackchain: I think it's covered here https://github.com/Emurgo/yoroi/pull/2698/files#diff-61ea1918772681dc1c2da59a60e08d3445570f4914b211ac63f5d44771ad0b53R9

stackchain: it covers for 2 scenarios: `files: file[]` and `files: file`(wrong metadata, yet loose check)

packages/cardano/src/cardano/api/parsers.test.ts
stackchain: k, gonna be moving this to product, right now it has the same behaviour as before, if it starts accepting invalid metadata it will be required some decisions around what to display `-` `invalid`, coerce into `string` i.e 123 -> `description: '123'` even though is wrong

michalszorad (approved):

PR #2702, opened 2023-09-22
chore: avoid notifying no changes

stackchain:
- [x] avoid firing 'utxos' event that causes many screens to rerender when there is no update in the utxos.

### DIFF
--- a/apps/wallet-mobile/src/yoroi-wallets/cardano/byron/ByronWallet.ts
+++ b/apps/wallet-mobile/src/yoroi-wallets/cardano/byron/ByronWallet.ts
@@ -1007,10 +1007,29 @@ export class ByronWallet implements YoroiWallet {
 
     await this.utxoManager.sync(addresses)
 
-    this._utxos = await this.utxoManager.getCachedUtxos()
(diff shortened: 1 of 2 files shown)


SorinC6 (approved):

jorbuedo (approved):

banklesss (approved):

apps/wallet-mobile/src/yoroi-wallets/cardano/byron/ByronWallet.ts
```diff
@@ -1007,10 +1007,29 @@ export class ByronWallet implements YoroiWallet {
...
+      this.notify({type: 'utxos', utxos: this.utxos})
+    }
+  }
+
+  private hasUtxoUpdated(oldUtxos: RawUtxo[], newUtxos: RawUtxo[]): boolean {
+    if (oldUtxos.length !== newUtxos.length) {
+      return true
+    }
+
+    const oldUtxoIds = new Set(oldUtxos.map((utxo) => utxo.utxo_id))
```
lisicky, line 1024: How often do we call hasUtxoUpdated ?  Creating a set from the scratch for each function call can be resource consuming

PR #2705, opened 2023-09-25
feat(swap): define base storage type to be used for web & mobile

neuodev:
```ts
// For mobile  
swapStorageMaker() // use async storage for react native 
// For web
swapStorageMaker({
   storage: {
     setItem: promisify(localStorage.setItem)
  }
})
```


stackchain:
hey @neuodev LGFN, later I'll be adding a storage `kind` to the types, so if `kind: 'local' | 'whatever'` we can defined `{kind: 'local', getItem(key: string): string | null} | {kind: 'async-storage', getItem:(key: string): Promise<string | null>` so you guys don't need to promisify everything.

PR #2707, opened 2023-09-25
fix/confirm tx types

stackchain:


apps/wallet-mobile/src/components/ConfirmTx/ConfirmTx.tsx
```diff
@@ -159,7 +158,7 @@ export const ConfirmTx = ({
       try {
         setIsProcessing(true)
 
-        let signedTx
+        let signedTx: YoroiSignedTx
```
michalszorad, line 161: `let signedTx: YoroiSignedTx | undefined`

michalszorad (approved):

banklesss (approved):

apps/wallet-mobile/src/components/ConfirmTx/ConfirmTx.tsx
stackchain: TS can infer if it was used before assigned, I believe this can cause confusion, cuz it is required to have `signedTx`.

PR #2716, opened 2023-09-26
chore: Add supported providers

SorinC6:
tasks:
- https://emurgo.atlassian.net/browse/YOMO-806

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
+++ b/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
@@ -132,7 +132,7 @@ export const CreateOrder = () => {
       const orderResult: Swap.CreateOrderData | undefined = makePossibleMarketOrder(
         orderDetails.sell,
         orderDetails.buy,
-        [orderDetails.selectedPool],
+        orderDetails.pools,
         orderDetails.slippage,
         orderDetails.address,
       )
--- a/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
+++ b/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
@@ -132,7 +132,7 @@ export const CreateOrder = () => {
       const orderResult: Swap.CreateOrderData | undefined = makePossibleMarketOrder(
         orderDetails.sell,
         orderDetails.buy,
-        orderDetails.pools,
+        [orderDetails.selectedPool],
         orderDetails.slippage,
         orderDetails.address,
       )
--- a/packages/swap/src/helpers/transformers.ts
+++ b/packages/swap/src/helpers/transformers.ts
@@ -157,7 +157,7 @@ export const transformersMaker = (
       ? openswapPools
           .filter((pool) =>
             supported_protocols.includes(
-              pool.provider as Swap.SupportedProtocol,
+              pool.provider as Swap.SupportedProvider,
             ),
           )
           .map(asYoroiPool)
--- a/packages/swap/src/helpers/transformers.ts
+++ b/packages/swap/src/helpers/transformers.ts
@@ -152,7 +152,24 @@ export const transformersMaker = (
   }
 
   const asYoroiPools = (openswapPools: OpenSwap.Pool[]): Swap.Pool[] => {
-    return openswapPools?.length > 0 ? openswapPools.map(asYoroiPool) : []
+    const supportedProviders: Swap.SupportedProtocol[] = [
+      'minswap',
+      'wingriders',
+      'sundaeswap',
+      'muesliswap',
+      'muesliswap_v2',
+      'vyfi',
+    ]
+
+    return openswapPools?.length > 0
+      ? openswapPools
+          .filter((pool) =>
+            supportedProviders.includes(
+              pool.provider as Swap.SupportedProtocol,
+            ),
+          )
+          .map(asYoroiPool)
+      : []
   }
 
   const asYoroiBalanceTokens = (
--- a/packages/swap/src/helpers/transformers.ts
+++ b/packages/swap/src/helpers/transformers.ts
@@ -3,7 +3,7 @@ import {Swap, Balance} from '@yoroi/types'
 import {OpenSwap} from '@yoroi/openswap'
 
 import {Quantities} from '../utils/quantities'
-import {supported_protocols} from '../translators/constants'
+import {supportedProtocols} from '../translators/constants'
 
 export const transformersMaker = (
   primaryTokenId: Balance.Token['info']['id'],
@@ -156,7 +156,7 @@ export const transformersMaker = (
     return openswapPools?.length > 0
       ? openswapPools
           .filter((pool) =>
-            supported_protocols.includes(
+            supportedProtocols.includes(
               pool.provider as Swap.SupportedProvider,
             ),
           )
--- a/packages/swap/src/helpers/transformers.ts
+++ b/packages/swap/src/helpers/transformers.ts
@@ -3,6 +3,7 @@ import {Swap, Balance} from '@yoroi/types'
 import {OpenSwap} from '@yoroi/openswap'
 
 import {Quantities} from '../utils/quantities'
+import {supported_protocols} from '../translators/constants'
 
 export const transformersMaker = (
   primaryTokenId: Balance.Token['info']['id'],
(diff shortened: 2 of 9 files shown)


packages/swap/src/helpers/transformers.ts
```diff
@@ -152,7 +152,24 @@ export const transformersMaker = (
   }
 
   const asYoroiPools = (openswapPools: OpenSwap.Pool[]): Swap.Pool[] => {
-    return openswapPools?.length > 0 ? openswapPools.map(asYoroiPool) : []
+    const supportedProviders: Swap.SupportedProtocol[] = [
```
stackchain: extract into `/src/translators/constants.ts` -> also export in the `src/index` so the client can query the supportedProviders.

stackchain: make it `ReadonlyArray<Swap.SupportedProtocol>`

packages/swap/src/helpers/transformers.ts
```diff
@@ -152,7 +152,24 @@ export const transformersMaker = (
...
+      'minswap',
+      'wingriders',
+      'sundaeswap',
+      'muesliswap',
+      'muesliswap_v2',
+      'vyfi',
+    ]
```
stackchain:
```suggestion
    ] as const
```

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -132,7 +132,7 @@ export const CreateOrder = () => {
       const orderResult: Swap.CreateOrderData | undefined = makePossibleMarketOrder(
         orderDetails.sell,
         orderDetails.buy,
-        orderDetails.pools,
+        [orderDetails.selectedPool],
```
jorbuedo: What's the point of `makePossibleMarketOrder` expecting a list of pools instead of just one, if in the only place we call it we only pass an array with one?

I see 3 options:
1. Keep this as it was
2. Modify `makePossibleMarketOrder` to expect use the one
3. Keep this change but justify in which other case we would call it with the whole list

packages/swap/src/helpers/transformers.ts
```diff
@@ -152,7 +152,24 @@ export const transformersMaker = (
...
+
+    return openswapPools?.length > 0
+      ? openswapPools
+          .filter((pool) =>
+            supportedProviders.includes(
+              pool.provider as Swap.SupportedProtocol,
```
jorbuedo: I'm a bit concerned about mixing `provider` and `protocol`.
Do openswapPools have a protocol? Otherwise, where are we getting the protocols from?

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
SorinC6: we had a bug here as basically when creating `makePossibleMarketOrder ` we already knew what pool was selected by the user, no point in sending the list again and filtering by the best pool again.. before we were sending the list and in some cases was not filtering right so the user was choosing a list -> but we were sending all the pool lists again and something weird was happening

packages/swap/src/helpers/transformers.ts
SorinC6: yeah, could be misleading. The protocol/providerName we are getting it from the pool list entry. Maybe we can rename it as `SupportedProvider` but not sure

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
SorinC6: I will assume we can go with the second one if needed

stackchain: this is the api to make a possible market order, which means that it should only operate on the pools that we support, ~~I'd say if "none" is provided, we should look into all the supported pools, if we provide we provide one only.~~

packages/swap/src/helpers/transformers.ts
stackchain: yap, so I'm

packages/swap/src/translators/constants.ts
```diff
@@ -0,0 +1,10 @@
+import {Swap} from '@yoroi/types'
+
+export const supported_protocols: ReadonlyArray<Swap.SupportedProtocol> = [
```
stackchain: for now, use camelCase to client, (I'm reconsidering that too)

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
stackchain: as long as we provide pools in the support list, I think it's fine.

packages/swap/src/translators/constants.ts
SorinC6: this is just for the package as I'm doing the filtering based on the supported ones in the package only - the app is just getting the supported list now

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
jorbuedo: I don't think this function can be called without a selectedPool present. And the selectedPool should already be a supported one, because they get filtered as soon as we get them. So I would say refactor `makePossibleMarketOrder` to remove the selection from array.

stackchain: k, I think it's a little bit more complex, I think we should ask product, cuz if the user didn't touch the selected pool, and is trying to create a market order, I'd say it should check all supported pools, get the best price, and then file the order. But if the user selected a pair, and has selected a specific pool, I'm assuming that pool is the one wanted, so we should create an order for that pool.

packages/swap/src/translators/constants.ts
stackchain: I think you meant, the client can see just the types?

packages/swap/src/translators/constants.ts
```diff
@@ -0,0 +1,10 @@
...
+  'minswap',
+  'wingriders',
+  'sundaeswap',
+  'muesliswap',
+  'muesliswap_v2',
+  'vyfi',
```
stackchain, line 9: are we supporting `vyfi` I thought it was using different address logic, and it was throwing some errors?

packages/types/src/index.ts
```diff
@@ -47,6 +47,7 @@ export namespace Swap {
   export type Pool = SwapPool
   export type PoolResponse = SwapPool[]
   export type PoolProvider = SwapPoolProvider
+  export type SupportedProtocol = SwapSupportedProvider
```
stackchain: maybe add some comment here cuz it looks odd, Protocol = Provider, ->

packages/swap/src/translators/constants.ts
SorinC6: yes, but have renamed it for now

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
SorinC6: yes, if the user is not touching the selected pool -> the default one with the best price will be used to create the order and if is choosing a specific one, that will be used.  No point in sending the list again when `makePossibleMarketOrder` as we already know the selected by default has the best price. 
- If we are sending the list when creating the  `makePossibleMarketOrder` again, even if user is choosing a specific pool - then the function will sort them and choose the best one anyway 🙃

SorinC6: for not touching the helper and calculation have just sent the best one in the array as an easy fix - but I think the best approach will be to refactor the `makePossibleMarketOrder` to always receive the default one or the one user chooses to create a order

packages/swap/src/translators/constants.ts
SorinC6: hmm, I thought we supported it as is officially supported by museliswap as well

packages/swap/src/helpers/transformers.ts
```diff
@@ -152,7 +153,15 @@ export const transformersMaker = (
   }
 
   const asYoroiPools = (openswapPools: OpenSwap.Pool[]): Swap.Pool[] => {
-    return openswapPools?.length > 0 ? openswapPools.map(asYoroiPool) : []
+    return openswapPools?.length > 0
+      ? openswapPools
+          .filter((pool) =>
+            supportedProtocols.includes(
+              pool.provider as Swap.SupportedProvider,
+            ),
```
stackchain:
```suggestion
          .filter((openswappool) =>
            supportedProtocols.includes(
              openswappool.provider as Swap.SupportedProvider,
            ),
```

packages/swap/src/translators/constants.ts
stackchain: supportedProviders (cuz protocol is used only in the create order) everywhere else they are using Providers.

packages/swap/src/translators/constants.ts
```diff
@@ -0,0 +1,10 @@
+import {Swap} from '@yoroi/types'
+
+export const supportedProtocols: ReadonlyArray<Swap.SupportedProvider> = [
```
stackchain:
```suggestion
export const supportedProviders: ReadonlyArray<Swap.SupportedProvider> = [
```

stackchain (approved):

PR #2717, opened 2023-09-26
fix: Show some fees and allow selected pool to be undefined

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-813


packages/swap/package.json
```diff
@@ -48,6 +48,7 @@
   "scripts": {
     "build": "yarn tsc && yarn lint && yarn test --ci --silent && yarn clean && bob build && yarn flow",
     "clean": "del-cli lib",
+    "dev": "bob build",
```
stackchain:
```suggestion
    "dev": "yarn clean && bob build",
```

packages/swap/src/helpers/orders/getTotalFees.ts
```diff
@@ -1,29 +1,62 @@
...
+  return [
+    Quantities.format(batcherFee.quantity, adaDecimals),
+    batcherFee.quantity,
+  ]
+}
+
+export const getMuesliswapFrontendFee = (
+  adaPrice: string,
+  milkInWallet: Balance.Quantity,
+  sell: Balance.Amount,
```
SorinC6: @jorbuedo  not sure what `milkInWallet` means or maybe the name is a bit weird, and where this function `getMuesliswapFrontendFee` is called in the app. 🤔

apps/wallet-mobile/src/features/Swap/useCases/ConfirmTxScreen/TransactionSummary.tsx
```diff
@@ -53,10 +49,10 @@ export const TransactionSummary = () => {
     {
       label: strings.swapFeesTitle,
       value: `${getTotalFees(
-        selectedPool?.batcherFee.quantity,
-        asQuantity(poolFee),
+        selectedPool.batcherFee,
+        selectedPool.fee,
+        createOrder.amounts.sell,
```
SorinC6: @jorbuedo  I thought `getTotalFees` will not calculate `pool fee` from percentage anymore

packages/swap/src/helpers/orders/getTotalFees.ts
jorbuedo: @SorinC6  milkInWallet would be how many milk tokens there are on the user wallet
It's part of the fee calculation for muesliswap frontend: https://docs.muesliswap.com/cardano/fee-structure

`getMuesliSwapFrontendFee` is not called anywhere yet. Discussed at https://emurgo.slack.com/archives/C02M83F3G72/p1695823475698869?thread_ts=1695822921.382099&cid=C02M83F3G72

apps/wallet-mobile/src/features/Swap/useCases/ConfirmTxScreen/TransactionSummary.tsx
jorbuedo: @SorinC6  The idea is to remove `getTotalFee`, this file is not updated yet. Need to be sure about the fee functions first, and clarify with product how to show the fees.

jorbuedo:
@stackchain @SorinC6 Current version works

- Shows just batcher fee in fees
- Liquidity pool fee quantity in sell token is not shown yet but it has a function to calculate it (pending UI design)
- Frontend fee is not shown (nor added with the batcher fee for a total), but there's a draft function to calculate it based on muesliswap algorithm (needs milk tokens in wallet, and the price in ada of the selling token). Probably needs more work
- Merged the selected pool undefined into this pr so it fixes the non ada pair final issues
- Test coverage temporary reduced to 90% so it compiles, but will need more tests for those fee functions and for every reducer bit that changed behavior based on selected pool undefined

We can either merge it as is, or wait for the tests. The tests will have to change again for:
- Frontend fee function usage clarified
- Move poolList into the state to select the best pool given price after fees, and thus selectedPool for market orders, for every amount change

stackchain:
The undefined pool was already defined so I'll be bringing it, will purpose something for the state during the weekend.

lisicky:
Frontend fee is fee of UI/API provider who submit a tx with order and do all calculations. It is the thing that we can charge from users if muesli doesn't require to charge it for them.  https://muesliswap.com/static/media/whitepaper.8e47144d4fe7c9f2447c.pdf

PR #2725, opened 2023-09-27
Revert "fix: add refresh btn logic"

jorbuedo:
Reverts Emurgo/yoroi#2722

```
asQuantity(formattedValue)
```
Is not ok. Formatted value can't be passed to asQuantity.

```
<AmountInput onChange={onChange} value={createOrder.limitPrice} editable={!disabled} />
```
createOrder.limitPrice is not directly assignable to the input

And

```
    {
      onSuccess: (poolList: Swap.Pool[]) => {
        const bestPool: Swap.Pool | undefined = poolList
          .sort((a: Swap.Pool, b: Swap.Pool) => a.price - b.price)
          .find(() => true)
        if (bestPool !== undefined) {
          const defaultPrice = createOrder.marketPrice
          const formattedValue = BigNumber(defaultPrice).decimalPlaces(PRECISION).toFormat(numberLocale)
          selectedPoolChanged(bestPool)
          limitPriceChanged(asQuantity(formattedValue))
        }
      },
    },
```
    
   
None of that should be modified from here. The pool change is already done on createOrder, and again limitPriceChanged can't receive that.


stackchain:
`Quantity = BigInt string` definitely not designed to be used as price TD to split it.

PR #2728, opened 2023-09-27
fix(swap): Non ada pair selection

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-815

When the selection screen wants to select a new token, the only action it can perform is `SellAmountChanged` or `BuyAmountChanged`, which not only serve to modify the quantity, but the tokenId as well.
The reducer for those actions takes care of updating the amounts, but expects the selectedPool to be correct.
When selecting new tokens, the pool is still the old one, so the getBuyAmount/getSellAmount set the other amount to the other amount in the pool, which was always ada because we started with it in the pair.

This fix prevents calling getBuyAmount/getSellAmount if we are setting a quantity = 0, since the other one will be 0 as well. After that, the pair has changed, and the app in createOrder will pick up the change, request new pools, and update the selectedPool. Or that's the idea, I only have ada on my prod wallet. And the testnet wallet has a mocked pool list with only ada pairs, so I can't verify it works.

### DIFF
--- a/packages/swap/package.json
+++ b/packages/swap/package.json
@@ -48,6 +48,7 @@
   "scripts": {
     "build": "yarn tsc && yarn lint && yarn test --ci --silent && yarn clean && bob build && yarn flow",
     "clean": "del-cli lib",
+    "dev": "yarn clean && bob build",
     "dgraph": "depcruise src --include-only \"^src\" --output-type dot | dot -T svg > dependency-graph.svg",
     "flow": ". ./scripts/flowgen.sh",
     "lint": "eslint \"**/*.{js,ts,tsx}\"",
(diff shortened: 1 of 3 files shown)


jorbuedo:
@stackchain @SorinC6  Manged to test it and it seems to work. Except for pairs that don't return pools, then it can't select a new pool (because there's none) and reuses the previous one, since selectedPool can't be undefined. Honestly, should probably allow undefined for selectedPool. Will have to refactor quite a bit of places to account for it being undefined, but not sure what else to do if the selected pair doesn't have pools.

stackchain:
> @stackchain @SorinC6 Manged to test it and it seems to work. Except for pairs that don't return pools, then it can't select a new pool (because there's none) and reuses the previous one, since selectedPool can't be undefined. Honestly, should probably allow undefined for selectedPool. Will have to refactor quite a bit of places to account for it being undefined, but not sure what else to do if the selected pair doesn't have pools.

Roger, A, B and P, otherwise 0.

stackchain (approved):

PR #2729, opened 2023-09-28
[Request for comment] Alow selected pool to be undefined

jorbuedo:
Let me know how you'd see this change.

WIP: No tests, no app changes yet.

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/common/useSwapTx.ts
+++ b/apps/wallet-mobile/src/features/Swap/common/useSwapTx.ts
@@ -16,14 +16,14 @@ export const useSwapTx = (options?: UseMutationOptions<YoroiUnsignedTx, Error, {
       data: {
         msg: splitStringInto64CharArray(
           JSON.stringify({
-            provider: createOrder.selectedPool.provider,
+            provider: createOrder.selectedPool?.provider,
             sellTokenId: createOrder.amounts.sell.tokenId,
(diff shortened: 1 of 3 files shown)


stackchain:
I was mapping all states yesterday, the suggested pool based on price is ok, but is not 100% correct it should be based on the price after the fee, cuz if you have pool products that are similar (bouncing around the same price), the fee + the amount that the user is trading are needed to actually find the best price, I'd say that the all pools calculations are also part of the state:
```
{
  createOrderData:...,
  unsignedTx:...,
  poolWithPrices: Swap.Pool[] & {calcs: {priceAfterFee: x, price: y, whateverelse: weneed}}
}
```

stackchain:
> I was mapping all states yesterday, the suggested pool based on price is ok, but is not 100% correct it should be based on the price after the fee, cuz if you have pool products that are similar (bouncing around the same price), the fee + the amount that the user is trading are needed to actually find the best price, I'd say that the all pools calculations are also part of the state:
> 
> ```
> {
>   createOrderData:...,
>   unsignedTx:...,
>   poolWithPrices: Swap.Pool[] & {calcs: {priceAfterFee: x, price: y, whateverelse: weneed}}
> }
> ```

So the list displayed of pools should come from here, and it can have more important info such as the affected price for all pools, and the selected pool will become the index of this list or undefined.

PR #2735, opened 2023-09-29
fix(swap): Fix price input format

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-825


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditLimitPrice.tsx
```diff
@@ -24,6 +23,7 @@ export const EditLimitPrice = () => {
   const {createOrder, limitPriceChanged} = useSwap()
   const sellTokenInfo = useTokenInfo({wallet, tokenId: createOrder.amounts.sell.tokenId})
   const buyTokenInfo = useTokenInfo({wallet, tokenId: createOrder.amounts.buy.tokenId})
+  const denomination = (sellTokenInfo.decimals ?? 0) - (buyTokenInfo.decimals ?? 0)
```
michalszorad, line 26: Will this work if sell token and buy token have different decimals?

jorbuedo: If they're different? Yes
If buy decimals are higher than sell decimals? Not sure... checking... maybe needs an absolute.

stackchain: I'd say that the denomination of the price should be fixed.

jorbuedo: It is right, denomination can be negative here

jorbuedo: > I'd say that the denomination of the price should be fixed.

Can't be fixed, depends on the pair you're trading

michalszorad: I think I understand now. The `createOrder.marketPrice` and `createOrder.limitPrice` are computed without domination, and the domination is added later for display

michalszorad (approved):

PR #2738, opened 2023-10-02
chore: impr ffee

stackchain:


packages/swap/src/translators/constants.ts
```diff
@@ -8,3 +11,52 @@ export const supportedProviders: ReadonlyArray<Swap.SupportedProvider> = [
...
+    secondaryTokenBalanceThreshold: Quantities.zero,
+    variableFeeMultiplier: 0.0,
+    variableFeeVisual: 0.0,
+    fixedFee: Quantities.zero,
+  },
+] as const
+
+export const milkTokenId = {
+  mainnet: '8a1cfae21368b8bebbbed9800fec304e95cce39a2a57dc35e2e3ebaa.4d494c4b',
+  preprod: '8a1cfae21368b8bebbbed9800fec304e95cce39a2a57dc35e2e3ebaa.4d494c4b',
```
stackchain, line 61: - [ ] get id `preprod`

packages/swap/src/helpers/orders/getBuyAmount.ts
```diff
@@ -30,8 +30,8 @@ export const getBuyAmount = (
     return {
       tokenId,
       quantity: asQuantity(
-        BigNumber(sell.quantity)
-          .dividedToIntegerBy(BigNumber(limit))
+        new BigNumber(sell.quantity)
```
jorbuedo, line 33: From their docs, `new is optional`
```
x = new BigNumber(123.4567)                // '123.4567'
// 'new' is optional
y = BigNumber(x)                           // '123.4567'
```

But we can add them if you like them

packages/swap/src/helpers/orders/getFrontendFee.ts
```diff
@@ -0,0 +1,65 @@
...
+  milkBalance,
+  sellInPrimaryTokenValue,
+  buyInPrimaryTokenValue,
+  primaryTokenInfo,
+  discountTiers = milkHoldersDiscountTiers,
+}: {
+  sellAmount: Balance.Amount
+  buyAmount: Balance.Amount
+  milkBalance: Balance.Quantity
+  sellInPrimaryTokenValue: Balance.Amount
```
jorbuedo, line 23: Do we get these from finding a pool between sell token and ada?

jorbuedo (approved):

packages/swap/src/helpers/orders/getBuyAmount.ts
stackchain: TS complains after BN 9.1

packages/swap/src/helpers/orders/getFrontendFee.ts
stackchain: these are received from the order + some metadata that it still don't have such as the sell/buy value in primary when they are not the primary. iUSD/DJED -> iUSD/ADA + DJED/ADA (based on what's selling)

packages/swap/src/helpers/orders/getBuyAmount.ts
jorbuedo: Uh, ok, I hadn't noticed anything but probably didn't do any install yet since 9.1 4 days ago

PR #2739, opened 2023-10-02
chore: Get best pool buy/sell + endpoint pools

lisicky:


packages/swap/src/helpers/orders/getBestBuyPoll.ts
jorbuedo, line 1: Typo in filename

packages/swap/src/helpers/orders/getBestBuyPoll.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @param pools - The liquidity pool list.
+ * @param sell - The desired sell amount.
+ *
+ * @returns The best pool to sell
+ * if the balance in the pool is insuficient it wont throw an error
+ * if the pools balance is 0 it will return undefined
+ * if the pool list is empty it will return undefined
+ */
+export const getBestBuyPool = (
```
jorbuedo: More of a personal preference but I find a reducer more readable

```ts
type Accumulator = {quantity: BalanceQuantity; pool?: Swap.Pool}
export const getBestBuyPool = (
  pools: Swap.Pool[],
  sell: Balance.Amount,
): Swap.Pool | undefined => {
  const result = pools.reduce(
    (best, pool): Accumulator => {
      const {quantity} = getBuyAmount(pool, sell)
      if (BigInt(quantity) > BigInt(best.quantity)) return {quantity, pool}
      return best
    },
    {quantity: Quantities.zero, pool: undefined} as Accumulator,
  )
  return result.pool
}
```

Haven't tested it

jorbuedo (commented):
The other thing I mentioned Juliano is that this algorithm doesn't really take into account the other fees paid in Ada.
Deposit is returned and frontend fee we set for all of them.
But the batcher(network) fee changes between pools and can make the difference between one being better than another one.
An easy check would be comparing the batcher fee between pools. If your current best pool has a lower batcher fee than the alternative, you know for sure yours is still better. Problem is when the current best has a higher fee, then you'd need the sell token price in ada to compare total values and it becomes a mess to calculate.
I'm pretty sure Muesli does this check it in their app because it interests them, since they have lower network fee than other pools (0.95 instead of 2 ada)

lisicky:
> The other thing I mentioned Juliano is that this algorithm doesn't really take into account the other fees paid in Ada.
> Deposit is returned and frontend fee we set for all of them.
> But the batcher(network) fee changes between pools and can make the difference between one being better than another one.
> An easy check would be comparing the batcher fee between pools. If your current best pool has a lower batcher fee than the alternative, you know for sure yours is still better. Problem is when the current best has a higher fee, then you'd need the sell token price in ada to compare total values and it becomes a mess to calculate.
> I'm pretty sure Muesli does this check it in their app because it interests them, since they have lower network fee than other pools (0.95 instead of 2 ada)

@jorbuedo  Yep you are right. Deposit and frontend fee won't make difference, but batcher fee probably will. But for the fair compare we need to cover batcher fee from ada to token or vice versa and then compare pools, because a token also has cost in a measurement of money. And we need to bring them to one unit of measurement. I'll investigate muesli api deeper. This PR is just starting point for our conversation about this topic.

packages/swap/src/helpers/orders/getBestBuyPoll.ts
```diff
@@ -0,0 +1,42 @@
...
+  pools: Swap.Pool[],
+  sell: Balance.Amount,
+): Swap.Pool | undefined => {
+  if (pools != null && pools.length === 0) {
```
michalszorad: This probably can't be `null`. Same applies in `getBestSellPool`

packages/swap/src/helpers/orders/getBestBuyPoll.ts
```diff
@@ -0,0 +1,42 @@
...
+    return undefined
+  }
+  let bestPool
```
michalszorad: `let bestPool: Swap.Pool | undefined = undefined` to add types

packages/swap/src/helpers/orders/getBestBuyPool.test.ts
```diff
@@ -0,0 +1,226 @@
...
+      price: 0,
+      batcherFee: {quantity: '2000000', tokenId: ''},
+      deposit: {quantity: '2000000', tokenId: ''},
+      poolId: '0',
+      lastUpdate: '0',
+      lpToken: {
+        quantity: '0',
+        tokenId: '0',
+      },
+    } as Swap.Pool
```
michalszorad: Instead of `as Swap.Pool` you can do `const pool3:  Swap.Pool = { ... }` so we get a TS error when any fields are missing

packages/swap/src/helpers/orders/getBestSellPool.test.ts
```diff
@@ -0,0 +1,116 @@
+import {Swap, Balance} from '@yoroi/types'
+
+import {getBestSellPool} from './getBestSellPool'
+import {getSellAmount} from './getSellAmount'
+
+describe('getBestSellPool', () => {
+  it('should return poll with min possible tokens to sell', () => {
```
michalszorad: `poll` -> `pool`

stackchain:
I believe these helpers will become sorters/finders, `findBy{BestX,BestY}` after #2740 we can process right after calculating all available pools.

packages/swap/src/helpers/orders/getBestBuyPoll.ts
lisicky: It's looks good. but in this case I prefer to use the old style code, it easier to support for me.

packages/swap/src/helpers/orders/getBestBuyPoll.ts
lisicky: [sound of pain on typescript]

packages/openswap/src/types.ts
```diff
@@ -71,30 +71,24 @@ export type CompletedOrder = {
...
   | 'sundaeswap'
   | 'wingriders'
+  | 'muesliswap'
   | 'muesliswap_v1'
   | 'muesliswap_v2'
   | 'muesliswap_v3'
   | 'muesliswap_v4'
   | 'vyfi'
   | 'spectrum'
+// | 'muesliswap_clp'
```
stackchain, line 85: @jorbuedo @lisicky I see this provider often in the requests while inspecting can we check with muesli if they have a list, endpoint some place to get the entire providers.

PR #2745, opened 2023-10-09
fix: Swap - confirm with password error reset

SorinC6:
part of the task: https://emurgo.atlassian.net/browse/YOMO-827
- disable sign button when user has empty input 
- clear error when the user is typing another password
demo: https://recordit.co/B8FFHUiLqw


apps/wallet-mobile/src/features/Swap/common/ConfirmWithSpendingPassword/ConfirmWithSpendingPassword.tsx
```diff
@@ -44,7 +54,13 @@ export const ConfirmWithSpendingPassword = ({onSubmit, isLoading, error}: Props)
 
       <Spacer fill />
 
-      <Button testID="swapButton" shelleyTheme title={strings.sign} onPress={() => onSubmit?.(spendingPassword)} />
+      <Button
+        testID="swapButton"
+        shelleyTheme
+        title={strings.sign}
+        onPress={() => onSubmit?.(spendingPassword)}
+        disabled={spendingPassword?.length === 0}
```
michalszorad: maybe instead of `?.` the `.` only will work here, as it should be always a string

apps/wallet-mobile/src/features/Swap/common/ConfirmWithSpendingPassword/ConfirmWithSpendingPassword.tsx
```diff
@@ -7,19 +7,29 @@ import {COLORS} from '../../../../theme'
...
-export const ConfirmWithSpendingPassword = ({onSubmit, isLoading, error}: Props) => {
+export const ConfirmWithSpendingPassword = ({onSubmit, isLoading, error, setErrorData}: Props) => {
   const spendingPasswordRef = React.useRef<RNTextInput>(null)
   const [spendingPassword, setSpendingPassword] = React.useState(
     features.prefillWalletInfo ? debugWalletInfo.PASSWORD : '',
   )
   const strings = useStrings()
 
+  React.useEffect(() => {
+    setErrorData && setErrorData({errorMessage: '', errorLogs: ''})
```
michalszorad: if it's just for clearing the error state, then maybe a `onPasswordChange()` prop should be good enough. In the parent then we can do `<ConfirmWithSpendingPassword onPasswordChange={() => setErrorData({errorMessage: '', errorLogs: ''})} />`

banklesss (approved):

PR #2751, opened 2023-10-10
fix: Show market price in pool selection

jorbuedo:
Price in `calculation` depends on which side we're selling. Is this the way we want it, or do we want to show A/B regardless of which side you're selling?

ps: All json and storybook things insist on regenerating itself, probably another PR missed them


apps/wallet-mobile/src/features/Swap/common/SelectPool/SelectPoolFromList/SelectPoolFromList.tsx
```diff
@@ -69,9 +74,17 @@ export const SelectPoolFromList = ({pools = []}: Props) => {
...
-                      <Boundary>
-                        <PriceInAda pool={pool} wallet={wallet} sell={orderData.amounts.sell} />
-                      </Boundary>
+                      <View style={styles.info}>
+                        <Text style={styles.infoLabel}>{strings.price}</Text>
+
+                        <Text style={styles.infoValue}>
+                          {Quantities.format(
+                            orderData.selectedPoolCalculation?.prices.market ?? Quantities.zero,
+                            denomination,
```
stackchain, line 83: does it handle negative denomination?

apps/wallet-mobile/src/features/Swap/common/SelectPool/SelectPoolFromList/SelectPoolFromList.tsx
```diff
@@ -28,6 +29,10 @@ export const SelectPoolFromList = ({pools = []}: Props) => {
   const navigate = useNavigateTo()
   const {track} = useMetrics()
 
+  const sellTokenInfo = useTokenInfo({wallet, tokenId: orderData.amounts.sell.tokenId})
```
stackchain, line 32: this way entire list can hang, please check the closer boundary to make sure it is displaying the expected waiting state

apps/wallet-mobile/src/features/Swap/common/SelectPool/SelectPoolFromList/SelectPoolFromList.tsx
jorbuedo: Yes, negative denomination is fine, just moves the comma the other way.

apps/wallet-mobile/src/features/Swap/common/SelectPool/SelectPoolFromList/SelectPoolFromList.tsx
jorbuedo: There's no boundary close. Both values should be cached by react-query though.

stackchain: k, I'm expecting the screens using this component to be handling the loading for it. triple checking all components tmr

PR #2755, opened 2023-10-11
chore(swap): handle swap balance errors

banklesss:
I refactored a bit to disable swap button in case of any balance error.

Related to [YOMO-819](https://emurgo.atlassian.net/browse/YOMO-819)

[YOMO-819]: https://emurgo.atlassian.net/browse/YOMO-819?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -239,6 +251,77 @@ export const CreateOrder = () => {
...
+
+  const hasBalance = !Quantities.isGreaterThan(
+    Quantities.sum([
+      tokenId === wallet.primaryTokenInfo.id ? orderData.amounts.sell.quantity : Quantities.zero,
+      orderData.selectedPoolCalculation?.cost.batcherFee.quantity ?? Quantities.zero,
+      orderData.selectedPoolCalculation?.cost.frontendFeeInfo.fee.quantity ?? Quantities.zero,
+      orderData.selectedPoolCalculation?.pool.deposit.quantity ?? Quantities.zero,
+    ]),
+    balance,
+  )
```
banklesss: Maybe here we need to separate primary token amount and other token amount right?

two types of balance: primary token balance + other types

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -239,6 +251,84 @@ export const CreateOrder = () => {
...
+
+  const hasSecondaryTokenBalance = !Quantities.isGreaterThan(
+    tokenId !== wallet.primaryTokenInfo.id ? orderData.amounts.sell.quantity : Quantities.zero,
+    balance,
+  )
```
banklesss: let me know what do you think about this

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
jorbuedo: @banklesss if you need all fees summed up, add them to makeOrderCalculations in another `cost` field, so you don't have to sum them in the app's front

jorbuedo: @stackchain we need to clarify if the frontendfee should be added or not
We said let's calculate it so we know how much it is, but we don't apply it
If we're not applying it it shouldn't be summed up here

banklesss: `ptTotalFee` added to `orderData.selectedPoolCalculation?.cost`

stackchain: the same way we have the other feeds, let's have the ptTotalFeeWithFeF and without

stackchain: and the client decides

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -70,7 +73,11 @@ export const CreateOrder = () => {
       setShowLimitPriceWarning(false)
     },
     onError: (error) => {
-      console.log(error)
+      if (error instanceof NotEnoughMoneyToSendError) {
+        setSellBackendError(strings.notEnoughBalance)
+      } else {
+        console.error(error)
+      }
```
stackchain, line 80:
```suggestion
      }
```

stackchain: maybe fallback to an `unkonwon error` if the api/network fails?

michalszorad (approved):

stackchain (approved):

PR #2758, opened 2023-10-12
fix(swap): infinite open orders loading

banklesss:


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/OpenOrders.tsx
```diff
@@ -201,6 +201,23 @@ export const OpenOrders = () => {
...
+      } catch (error) {
+        Alert.alert(strings.generalErrorTitle, strings.generalErrorMessage(error))
+      }
+
+      setIsLoading(false)
+      return fee
+    },
+    [cancelOrder, strings, wallet],
+  )
+
```
stackchain, line 220: can we turn it into a react-query and manually refetch on click (enable: false)? also disable cache.

stackchain: build fee as an Amount with PT.

PR #2759, opened 2023-10-12
fix(modal): Fix event handlers inside modal

michalszorad:
Resolves 

> Centralization of transport selection is not right

### DIFF
--- a/apps/wallet-mobile/src/components/Modal/ModalScreen.tsx
+++ b/apps/wallet-mobile/src/components/Modal/ModalScreen.tsx
@@ -2,7 +2,6 @@ import {useCardAnimation} from '@react-navigation/stack'
 import React from 'react'
 import {
   Animated,
-  GestureResponderEvent,
   KeyboardAvoidingView,
   NativeTouchEvent,
   Platform,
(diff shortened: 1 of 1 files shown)


banklesss (approved):

apps/wallet-mobile/src/components/Modal/ModalScreen.tsx
```diff
@@ -59,21 +61,24 @@ export const ModalScreen = () => {
             styles.animatedView,
           ]}
         >
-          <View style={styles.sheet} onResponderMove={onResponderMove} onStartShouldSetResponder={() => true}>
-            <Header />
+          <View style={styles.sheet}>
+            <Header onResponderMove={onResponderMove} onStartShouldSetResponder={() => true} />
```
michalszorad, line 65: The swipe down now only works in header, as this was preventing scrollable view to work

apps/wallet-mobile/src/components/Modal/ModalScreen.tsx
```diff
@@ -36,7 +36,9 @@ export const ModalScreen = () => {
   }
 
   return (
-    <Pressable style={styles.backdrop} onPress={closeModal}>
+    <View style={styles.backdrop}>
+      <Pressable style={styles.cancellableArea} onPress={closeModal} />
```
michalszorad, line 40: Presasble is now not the parent, as this was preventing scrollable view to work

PR #2768, opened 2023-10-12
fix(swap): fee balance errors

banklesss:
To address this comment:

https://emurgo.atlassian.net/browse/YOMO-819?focusedCommentId=18434

And to remove unsigned tx balance errors after the user retypes again in the sell input when he faces these errors

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
+++ b/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
@@ -75,7 +75,7 @@ export const CreateOrder = () => {
     },
     onError: (error) => {
       if (error instanceof NotEnoughMoneyToSendError) {
-        setSellBackendError(strings.notEnoughBalance)
+        setSellBackendError(strings.notEnoughFeeBalance)
         return
       }
 
@@ -365,7 +365,7 @@ const styles = StyleSheet.create({
   },
 })
 
-const useSellInput = ({onChange}: {onChange?: () => void}) => {
+export const useSellInput = ({onChange}: {onChange?: () => void}) => {
   const strings = useStrings()
   const {orderData, sellQuantityChanged} = useSwap()
   const {isSellTouched} = useSwapTouched()
--- a/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
+++ b/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
@@ -2,17 +2,19 @@ import {makeLimitOrder, makePossibleMarketOrder, useSwap, useSwapCreateOrder, us
 import {Swap} from '@yoroi/types'
 import BigNumber from 'bignumber.js'
 import * as React from 'react'
-import {Alert, KeyboardAvoidingView, Platform, StyleSheet, View, ViewProps} from 'react-native'
+import {Alert, KeyboardAvoidingView, Platform, StyleSheet, TextInput, View, ViewProps} from 'react-native'
 import {ScrollView} from 'react-native-gesture-handler'
 
 import {Button, Spacer} from '../../../../../components'
 import {LoadingOverlay} from '../../../../../components/LoadingOverlay'
+import {useLanguage} from '../../../../../i18n'
 import {useMetrics} from '../../../../../metrics/metricsManager'
 import {useSelectedWallet} from '../../../../../SelectedWallet'
 import {COLORS} from '../../../../../theme'
 import {isEmptyString} from '../../../../../utils'
 import {NotEnoughMoneyToSendError} from '../../../../../yoroi-wallets/cardano/types'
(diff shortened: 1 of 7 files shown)


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -74,7 +75,7 @@ export const CreateOrder = () => {
     },
     onError: (error) => {
       if (error instanceof NotEnoughMoneyToSendError) {
-        setSellBackendError(strings.notEnoughBalance)
+        setSellBackendError(strings.notEnoughFeeBalance)
```
banklesss, line 78: most "restrictive" message error

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditSellAmount/EditSellAmount.stories.tsx
banklesss, line 1: to do: take into consideration these missing states in `createOrder.stories.tsx` 👀

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -351,3 +364,39 @@ const styles = StyleSheet.create({
     borderTopColor: COLORS.BORDER_GRAY,
   },
 })
+
+export const useSellInput = ({onChange}: {onChange?: () => void}) => {
```
banklesss, line 368: needed to clear unsigned tx error

banklesss: open to better ideas of course

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditSellAmount/EditSellAmount.stories.tsx
banklesss: done

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
michalszorad: I was wondering whether `<EditSellAmount onChange={() => clearErrors()} />` would work. That way we would not need to create this hook. What are your thoughts?

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
```diff
@@ -220,7 +227,12 @@ export const CreateOrder = () => {
 
             <TopTokenActions />
 
-            <EditSellAmount error={sellError} />
+            <EditSellAmount
```
stackchain, line 230: `EditSellAmount` can change the `tokenId` and the `quantity`, if this one is just for quantity, add it to the API, `sellInputQuantity`, same for the hook `useSellQuantityInput`

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/CreateOrder.tsx
stackchain: EditSellAmount can update `tokenId` and `quantity` so the api would be onChange this and that, in the past we will be using  a controller to manage these behaviours and would be adding everything there, I'm ok with splitting into hooks to manage it, just believe we need to be more specific in what is changing and in the api.

jorbuedo: My thoughts are that more hooks is always worse.

stackchain: To me more props are always worst. Definitely don't want `<InputController><Input></></>`

jorbuedo: And also that I don't get why this logic that was in a component needs to be in createOrder now. There has to be a better way.

stackchain: on top of my head, EditSellAmount was playing the controller of AmountCard, yet CreateOrder is the one that should be doing it. So we can even discard EditSellAmount

stackchain: `<FormController><FormViewYadaYada><Input1><Input2>` form controller is responsible to deal with all errors otherwise how will it know if it can proceed or not?

banklesss:
Refactor here https://github.com/Emurgo/yoroi/pull/2786

PR #2770, opened 2023-10-13
fix(swap): reselect primary  token to sell when no buy token is selected

banklesss:
Related to [YOMO-878](https://emurgo.atlassian.net/browse/YOMO-878)

[YOMO-878]: https://emurgo.atlassian.net/browse/YOMO-878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditBuyAmount/SelectBuyTokenFromListScreen/SelectBuyTokenFromListScreen.tsx
+++ b/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/CreateOrder/EditBuyAmount/SelectBuyTokenFromListScreen/SelectBuyTokenFromListScreen.tsx
@@ -204,7 +204,7 @@ const SelectableToken = ({tokenForList, wallet}: SelectableTokenProps) => {
 
   return (
     <TouchableOpacity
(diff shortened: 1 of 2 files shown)


SorinC6 (approved):

michalszorad (approved):

jorbuedo (approved):
@stackchain This was caused by the default state having buy and sell as Ada, thus not allowing the sell to be set as Ada because the buy was already Ada.
Using the touched context to fix it works. But I think with the new state not reflecting 1:1 the createOrder argument, wouldn't it be better if buy started as undefined? Starting as both Ada is already a wrong state anyway.

stackchain:
> @stackchain This was caused by the default state having buy and sell as Ada, thus not allowing the sell to be set as Ada because the buy was already Ada. Using the touched context to fix it works. But I think with the new state not reflecting 1:1 the createOrder argument, wouldn't it be better if buy started as undefined? Starting as both Ada is already a wrong state anyway.

Yap, this is definitely right, in the next changes we definitely need to introduce, `unknown token`, `unknown decimals`, and allowing undefined on state. If we apply right now will create a cascade of `.?` everywhere.

PR #2771, opened 2023-10-13
fix(swap): Fix limit price precision / denomination

michalszorad:
Resolves [YOMO-855](https://emurgo.atlassian.net/browse/YOMO-855)

[YOMO-855]: https://emurgo.atlassian.net/browse/YOMO-855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -142,13 +142,13 @@ export const Quantities = {
     const parts = sanitized.split(decimalSeparator)
     const isDec = parts.length >= 2
```
stackchain: Branch all logic as earlier as possible to group up the logic, avoid index accessing.
```suggestion
    let fullDecValue = sanitized
    const isDec = parts.length >= 2
    if (isDec) {
        const [int, dec] = sanitized.split(decimalSeparator)
        fullDecValue = `${parts[0]}${decimalSeparator}${parts[1].slice(0, precision)}1`
        
    }

```

jorbuedo: Aren't you index accessing anyway? isDec already checks that parts indeed has length 2, so I don't see the harm in using the index

jorbuedo (approved):

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
stackchain: yap, but is just unreadable, part[0] (what is part), `1` hacky one to keep decimals, you have to scan the code and make an effort to understand. 
`const [whole, decimal] = sanitzedInput` it's read once, and you got it.

michalszorad: @stackchain @jorbuedo that's now updated, and also added comment above the `1` to explain the use-case

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,22 +133,31 @@ export const Quantities = {
 
     return absoluteQuantity.isEqualTo(minimalFractionalPart)
   },
-  parseFromText: (text: string, denomination: number, format: NumberLocale) => {
+  parseFromText: (text: string, denomination: number, format: NumberLocale, precision = denomination) => {
     const {decimalSeparator} = format
     const invalid = new RegExp(`[^0-9${decimalSeparator}]`, 'g')
     const sanitized = text === '' ? '' : text.replaceAll(invalid, '')
     if (sanitized === '') return ['', `0`] as [string, Balance.Quantity]
```
stackchain:
```suggestion
    if (sanitized === '') return ['', Quantities.zero] as [string, Balance.Quantity]
```

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,22 +133,31 @@ export const Quantities = {
...
     if (sanitized.startsWith(decimalSeparator)) return [`0${decimalSeparator}`, `0`] as [string, Balance.Quantity]
```
stackchain:
```suggestion
    if (sanitized.startsWith(decimalSeparator)) return [`0${decimalSeparator}`, Quantities.zero] as [string, Balance.Quantity]
```

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,22 +133,31 @@ export const Quantities = {
...
     const parts = sanitized.split(decimalSeparator)
```
stackchain, line 142:
```suggestion
```

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,22 +133,31 @@ export const Quantities = {
...
+
+    let fullDecValue = sanitized
+    let value = sanitized
+
     const isDec = parts.length >= 2
```
stackchain:
```suggestion
```

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,22 +133,31 @@ export const Quantities = {
...
 
-    const fullDecValue = isDec ? `${parts[0]}${decimalSeparator}${parts[1].slice(0, denomination)}1` : sanitized
+    if (isDec) {
```
stackchain:
```suggestion
    if (parts.length > 1) {
```

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,22 +133,31 @@ export const Quantities = {
...
+      const [int, dec] = parts
+      // trailing `1` is to allow user type `1.0` without losing the decimal part
+      fullDecValue = `${int}${decimalSeparator}${dec.slice(0, precision)}1`
+      value = `${int}${decimalSeparator}${dec.slice(0, precision)}`
+    }
+
     const fullDecFormat = new BigNumber(fullDecValue.replace(decimalSeparator, '.')).toFormat()
     const input = isDec ? fullDecFormat.slice(0, -1) : fullDecFormat
```
stackchain: move inside the if, add a comment that the trailing is removed inside the hasDecimals logic

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
stackchain: if possible remove the casting

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
jorbuedo: You'll have to declare input as `let` before and move both lines inside the if

Which is why I preferred everything as const and the ternaries 😅

jorbuedo (approved):

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
stackchain: BigNumber is object so const/let won't make that difference cuz you still can mutate it, but I also prefer and agree on not changing after declaring with initialization. To look even better I'd move everything dec / not decimal and return based on that, we will have some duplication, but readability is more important than duplication.

apps/wallet-mobile/src/yoroi-wallets/utils/utils.test.ts
```diff
@@ -129,6 +129,9 @@ describe('Quantities', () => {
     expect(Quantities.parseFromText('55.10', 3, english)).toEqual(['55.10', '55100'])
 
     expect(Quantities.parseFromText('ab1.5c,6.5', 3, english)).toEqual(['1.56', '1560'])
+
+    expect(Quantities.parseFromText('1.23456', 0, english, 3)).toEqual(['1.234', '1.234'])
+    expect(Quantities.parseFromText('1.23456', 2, english, 3)).toEqual(['1.234', '123.4'])
```
stackchain, line 134: ~~does it have the case for `1.0000` / `1.0`  ?~~

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,26 +133,43 @@ export const Quantities = {
...
+      // trailing `1` is to allow the user to type `1.0` without losing the decimal part
+      fullDecValue = `${int}${decimalSeparator}${dec?.slice(0, precision)}1`
+      value = `${int}${decimalSeparator}${dec?.slice(0, precision)}`
+      fullDecFormat = new BigNumber(fullDecValue.replace(decimalSeparator, '.')).toFormat()
+      // remove trailing `1`
+      input = fullDecFormat.slice(0, -1)
+    }
 
-    const value = isDec ? `${parts[0]}${decimalSeparator}${parts[1].slice(0, denomination)}` : sanitized
     const quantity = new BigNumber(value.replace(decimalSeparator, '.'))
```
stackchain:
```suggestion
    const quantity = asQuantity(new BigNumber(value.replace(decimalSeparator, '.'))
```

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,26 +133,43 @@ export const Quantities = {
...
-      .decimalPlaces(denomination)
+      .decimalPlaces(precision)
       .shiftedBy(denomination)
-      .toString(10)
+      .toString(10) as Balance.Quantity
```
stackchain:
```suggestion
      .toString(10))
```

apps/wallet-mobile/src/yoroi-wallets/utils/utils.test.ts
michalszorad: Yes, it has a test for ` expect(Quantities.parseFromText('55.0', 3, english)).toEqual(['55.0', '55000'])`

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
```diff
@@ -133,26 +133,43 @@ export const Quantities = {
...
-    const fullDecValue = isDec ? `${parts[0]}${decimalSeparator}${parts[1].slice(0, denomination)}1` : sanitized
-    const fullDecFormat = new BigNumber(fullDecValue.replace(decimalSeparator, '.')).toFormat()
-    const input = isDec ? fullDecFormat.slice(0, -1) : fullDecFormat
+    let fullDecValue = sanitized
+    let value = sanitized
+
+    let fullDecFormat = new BigNumber(fullDecValue.replace(decimalSeparator, '.')).toFormat()
+    let input = fullDecFormat
+
+    if (parts.length >= 2) {
```
stackchain: maybe if parts.length <= 1 return earlier and then process the decimals ?

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
jorbuedo: Maybe it's because I wrote it, but it looked more clear to me before. Now with the lets, an if dividing everything, the early return and a couple lines duplicated because of it, looks harder to follow to me.

apps/wallet-mobile/src/yoroi-wallets/utils/utils.ts
michalszorad: That's now updated

PR #2774, opened 2023-10-13
fix: Removes loading overlay

jorbuedo:
That's my easy solution to the problem of the overlay not being full screen


michalszorad (approved):

stackchain:
To disable the inputs it will need the form context done.

PR #2779, opened 2023-10-16
fix(swap): Fix collateral check

michalszorad:


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/OpenOrders.tsx
```diff
@@ -209,6 +213,11 @@ export const OpenOrders = () => {
 
   const openCancellationModal = async (order: MappedOpenOrder) => {
     if (order.owner === undefined || order.utxo === undefined) return
```
stackchain, line 213: if this is possible I'd say to disable the button, otherwise the user will tap on Cancel and nothing will happen.

jorbuedo (approved):

PR #2801, opened 2023-10-19
update(swap): add zendesk real link

SorinC6:
Added zendesk real link


apps/wallet-mobile/src/yoroi-wallets/cardano/constants/common.ts
```diff
@@ -8,6 +8,7 @@ export const ACCOUNT_INDEX = 0
 export const HARD_DERIVATION_START = 2147483648
 export const CHIMERIC_ACCOUNT = 2
 export const STAKING_KEY_INDEX = 0
+export const SWAP_ZENDESK_LINK = 'https://emurgohelpdesk.zendesk.com/hc/en-us/sections/8154256843407-Swap'
```
banklesss: I think this folder (`yoroi-wallets`) will go out at some point.

banklesss: I would add the link in the same file that is being used.

banklesss (approved):

PR #2808, opened 2023-10-20
refactor: swap frontend fee

stackchain:


michalszorad (approved):

packages/api/src/app/api/frontend-fees.ts
```diff
@@ -0,0 +1,55 @@
...
+      if (!parsedResponse)
+        return Promise.reject(new Error('Invalid frontend fee response'))
+
+      return Promise.resolve(parsedResponse)
+    })
+  }
+
+const SwapAggregatorSchema = z.union([
+  z.literal('muesliswap'),
+  z.literal('dexhunter'),
```
SorinC6: @stackchain, related to the `dexhunter` in which case will use it? as is the first time I have seen this mentioned. I thought we would be using only `muesliswap`

stackchain: hey @SorinC6 no worries, it's not for now.

SorinC6 (commented):
@stackchain shouldn't we add the frontend fee when creating the`createYoroiEntry` as well ?

michalszorad:
@SorinC6 yes, I'm working that piece 😄

stackchain:
> @stackchain shouldn't we add the frontend fee when creating the`createYoroiEntry` as well ?

If is gte 0 yap, @MichalSzorad is checking the lib side to see if an `Entry` with targets will work, we are also blocked by the wallet address from management to perform the deposits.

PR #2815, opened 2023-10-20
chore: Added tests

stackchain:


packages/swap/src/translators/reactjs/state/state.ts
```diff
@@ -376,16 +377,8 @@ const orderReducer = (
...
-          draft.orderData.selectedPoolCalculation.pool.tokenA.tokenId
-        ) {
-          draft.orderData.amounts.buy =
-            draft.orderData.selectedPoolCalculation.sides.buy
-        } else {
-          draft.orderData.amounts.sell =
-            draft.orderData.selectedPoolCalculation.sides.sell
-        }
+        draft.orderData.amounts.buy =
+          draft.orderData.selectedPoolCalculation.sides.buy
```
stackchain, line 381: `sides` are already on the right side, and sell is locked (doesn't get updated), only buy needs update

packages/swap/src/translators/reactjs/state/state.ts
```diff
@@ -433,10 +426,10 @@ const orderReducer = (
       // this can cause a pool if not enough supply to be selected
       // must be handled on the UI
       case SwapCreateOrderActionType.LimitPriceChanged:
-        draft.orderData.limitPrice = action.limitPrice
-
         if (state.orderData.type === 'market') break
 
+        draft.orderData.limitPrice = action.limitPrice
```
stackchain, line 431: branching earlier, price shouldn't be updated if type is not limit

packages/swap/src/translators/reactjs/state/state.ts
```diff
@@ -606,7 +600,7 @@ const orderReducer = (
         draft.orderData.calculations = makeOrderCalculations({
           orderType: state.orderData.type,
           amounts: state.orderData.amounts,
-          limitPrice: state.orderData.limitPrice,
+          limitPrice: undefined,
```
stackchain, line 603: if limit is passed it won't get updated, the if (shouldn't happen) the FFEE changes while creating an order, it must be reseted to market

PR #2819, opened 2023-10-23
fix(swap): Form state

jorbuedo:
@banklesss Steps on the one you're fixing

https://emurgo.atlassian.net/browse/YOMO-896
https://emurgo.atlassian.net/browse/YOMO-897


banklesss:
I added the reducer change here: https://github.com/Emurgo/yoroi/pull/2820

Also there is a refactor of the error management.

thanks!

PR #2820, opened 2023-10-23
fix(swap): balance errors

banklesss:
Related to:

- [YOMO-896](https://emurgo.atlassian.net/browse/YOMO-896)
- [YOMO-897](https://emurgo.atlassian.net/browse/YOMO-897)

[YOMO-896]: https://emurgo.atlassian.net/browse/YOMO-896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[YOMO-897]: https://emurgo.atlassian.net/browse/YOMO-897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/common/SwapFormProvider.tsx
+++ b/apps/wallet-mobile/src/features/Swap/common/SwapFormProvider.tsx
@@ -72,39 +72,6 @@ export const SwapFormProvider = ({
     ...initialState,
   })
 
-  const noPoolError =
-    orderData.selectedPoolCalculation === undefined && state.buyQuantity.isTouched && state.sellQuantity.isTouched
-      ? strings.noPool
-      : undefined
-
-  const buyError =
-    noPoolError !== undefined
-      ? noPoolError
-      : (!Quantities.isZero(buyQuantity) && !hasBuyTokenSupply) ||
-        (state.sellQuantity.isTouched && state.buyQuantity.isTouched && pool === undefined)
-      ? strings.notEnoughSupply
-      : undefined
-
-  const sellError =
-    noPoolError !== undefined
-      ? noPoolError
-      : !Quantities.isZero(sellQuantity) && !hasSellBalance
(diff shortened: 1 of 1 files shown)


apps/wallet-mobile/src/features/Swap/common/SwapFormProvider.tsx
```diff
@@ -270,9 +335,11 @@ const swapFormReducer = (state: SwapFormState, action: SwapFormAction) => {
         break
 
       case SwapFormActionType.ResetSwapForm:
-      case SwapFormActionType.ClearSwapForm:
         return defaultState
 
+      case SwapFormActionType.ClearSwapForm:
+        return state
```
michalszorad, line 341: Isn't this case redundant?

jorbuedo: It could be removed altogether, yes, but I guess it's there for completeness in case there's something to do at some point?

stackchain: I'd say that everything that updates the form for first should control the change in the state, so if the clear set the inputs and dispatch the reset quantities underlying the state action should be initiated/moved here.

stackchain: owww, it's inside the helpers not inside the reducer. it should be colocated.

stackchain: k, since we throw instead of defaulting to return the state, I think it's fine, I'd colocate that later.

stackchain (approved):

PR #2823, opened 2023-10-24
refactor: Tokens state

stackchain:
- [x] Add partial token info to the state
- [x] Update the actions on UI 
- [x] Apply the denomination on frontend fee calcs

--- 

Later the Portfolio manager will change it, and the amounts will carry all info required, for now adding extra data to state

Relates to YOMO-902

---

Prices pre-calc are incorrect it can't be used yet, when adding fees (fees are in Lovelace) it needs conversion when adding it counting as non pt. I.e how much 2 ADA would be as MIN. (apples + bananas)

### DIFF
--- a/packages/swap/src/translators/reactjs/state/state.ts
+++ b/packages/swap/src/translators/reactjs/state/state.ts
@@ -1,4 +1,4 @@
-import {App, Balance, Swap} from '@yoroi/types'
+import {App, Balance, RemoveUndefined, Swap} from '@yoroi/types'
 import {produce} from 'immer'
 
 import {Quantities} from '../../../utils/quantities'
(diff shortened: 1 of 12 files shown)


packages/swap/src/translators/reactjs/state/state.ts
```diff
@@ -356,6 +377,11 @@ const orderReducer = (
           sell: state.orderData.amounts.buy,
           buy: state.orderData.amounts.sell,
         }
+        draft.orderData.tokens = {
+          sellInfo: state.orderData.tokens.buyInfo,
+          buyInfo: state.orderData.tokens.sellInfo,
+          denomination: -state.orderData.tokens.denomination,
```
michalszorad: What does negative denomination mean?

jorbuedo:
:warning: This PR will surely conflict with https://github.com/Emurgo/yoroi/pull/2826

packages/swap/src/translators/reactjs/state/state.ts
stackchain: it just invert the direction of 0 based on decimals difference for price

banklesss (approved):

michalszorad (approved):

jorbuedo (approved):
I see some things to be corrected, but I'll fix them on another PR. Merging

PR #2826, opened 2023-10-24
fix: Adjust sell quantity when changing token

jorbuedo:
See comments in: https://emurgo.atlassian.net/browse/YOMO-899

### DIFF
--- a/packages/swap/src/translators/reactjs/state/state.ts
+++ b/packages/swap/src/translators/reactjs/state/state.ts
@@ -1,5 +1,6 @@
 import {App, Balance, RemoveUndefined, Swap} from '@yoroi/types'
 import {produce} from 'immer'
+import BigNumber from 'bignumber.js'
 
 import {Quantities} from '../../../utils/quantities'
 import {makeOrderCalculations} from '../../../helpers/orders/factories/makeOrderCalculations'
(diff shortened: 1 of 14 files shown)


packages/swap/src/translators/reactjs/state/state.ts
```diff
@@ -87,7 +88,10 @@ export type SwapCreateOrderActions = Readonly<{
   limitPriceChanged: (limitPrice: Balance.Quantity) => void
   sellQuantityChanged: (quantity: Balance.Quantity) => void
   buyQuantityChanged: (quantity: Balance.Quantity) => void
-  sellTokenIdChanged: (tokenId: Balance.TokenInfo['id']) => void
+  sellTokenIdChanged: (
```
michalszorad: Maybe `sellTokenChanged` to highlight the function does not take only id?

packages/swap/src/translators/reactjs/state/state.ts
```diff
@@ -87,7 +88,10 @@ export type SwapCreateOrderActions = Readonly<{
...
+    tokenId: Balance.TokenInfo['id'],
+    denominationDifference: number,
+  ) => void
   buyTokenIdChanged: (tokenId: Balance.TokenInfo['id']) => void
```
michalszorad: Do we need the same for `buyTokenIdChanged`?

jorbuedo: No, the buy side recalculates based on the sell side, so the quantity is not kept.

packages/swap/src/translators/reactjs/state/state.ts
jorbuedo: https://github.com/Emurgo/yoroi/pull/2823 will probably implement this different

michalszorad (approved):

PR #2834, opened 2023-10-25
chore: hide liq prov fee when limit order

stackchain:


apps/wallet-mobile/translations/messages/src/WalletNavigator.json
stackchain, line 1: missed in another pr

banklesss (approved):

PR #2851, opened 2023-10-30
fix: Amountcard input ref

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-917

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
+++ b/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
@@ -1,5 +1,5 @@
 import {Balance} from '@yoroi/types'
-import React, {useRef} from 'react'
+import React from 'react'
 import {defineMessages, useIntl} from 'react-intl'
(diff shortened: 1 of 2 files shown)


apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
```diff
@@ -39,8 +39,6 @@ export const AmountCard = ({
 }: Props) => {
   const strings = useStrings()
   const {quantity, tokenId} = amount
-  const amountInputRef = useRef<TextInput>(inputRef?.current ?? null)
```
stackchain, line 42: 👀

stackchain (approved):

apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
jorbuedo: I don't know, I think that worked before, passing the parent's ref to useRef to reuse it. Could be something to do with it coming from a context now. Or maybe never worked like that.

PR #2852, opened 2023-10-30
fix: Use system language as locale when different than app language

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-916

### DIFF
--- a/apps/wallet-mobile/src/i18n/LanguageProvider.tsx
+++ b/apps/wallet-mobile/src/i18n/LanguageProvider.tsx
@@ -30,7 +30,7 @@ export const LanguageProvider = ({children}: {children: React.ReactNode}) => {
   return (
     <LanguageContext.Provider
       value={{
-        numberLocale: numberLocales[languageCode],
+        numberLocale: numberLocales[defaultLanguageCode],
         languageCode,
         selectLanguageCode,
         supportedLanguages,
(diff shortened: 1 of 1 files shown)


apps/wallet-mobile/src/i18n/LanguageProvider.tsx
```diff
@@ -93,7 +93,7 @@ const useLanguageCode = ({onSuccess, ...options}: UseQueryOptions<LanguageCode>
       return languageCode ?? defaultLanguageCode
     },
     onSuccess: (languageCode) => {
-      updateLanguageSettings(languageCode)
+      updateLanguageSettings(defaultLanguageCode)
```
michalszorad, line 96: Does it still make sense to call these functions if we always use defaultLanguageCode?

SorinC6 (approved):

apps/wallet-mobile/src/i18n/LanguageProvider.tsx
jorbuedo: Mm... well, we have to call it some place. Could do it just once outside everything, next the the defaultLanguageCode declaration

jorbuedo: Not a huge deal though. This way it could react to the user changing the device language maybe

PR #2868, opened 2023-11-02
Fix: Use react-native-localize for language detection and decimal separator

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-920


apps/wallet-mobile/jestSetup.ts
```diff
@@ -31,6 +31,34 @@ jest.mock('react-native-reanimated', () => {
   Reanimated.default.call = () => {}
   return Reanimated
 })
+jest.mock('react-native-localize', () => ({
```
jorbuedo, line 34: Docs say to use
```tsx
import localizeMock from "react-native-localize/mock";

jest.mock("react-native-localize", () => localizeMock);
```

But our jest complains importing esm, says `unexpected export` stuff.
Tried with require, wouldn't work either.
So I just copied all the mocked functions here.

stackchain (approved):

PR #2870, opened 2023-11-03
fix(banxa): Fix modal content line height and modal height

michalszorad:
Resolves https://emurgo.atlassian.net/browse/YOMO-941

**After**
Screenshot 2023-11-03 at 15 15 26

**Before**
IMG-1361

### DIFF
--- a/apps/wallet-mobile/src/TxHistory/ActionsBanner.tsx
+++ b/apps/wallet-mobile/src/TxHistory/ActionsBanner.tsx
@@ -41,12 +41,13 @@ export const ActionsBanner = ({disabled = false}: {disabled: boolean}) => {
 
   const handleOnBuy = () => {
     track.walletPageExchangeBottomSheetClicked()
-    const modalHeight = 338
+    const modalHeight = 300
     openModal(
       strings.buyTitle,
       <View style={styles.buyModalContent}>
         <Text style={styles.buyInfo}>{strings.buyInfo}</Text>
 
+        <Spacer height={32} />
         <Button
           shelleyTheme
           title={strings.proceed}
@@ -69,6 +70,7 @@ export const ActionsBanner = ({disabled = false}: {disabled: boolean}) => {
             closeModal()
           }}
         />
+        <Spacer height={23} />
       </View>,
       modalHeight,
     )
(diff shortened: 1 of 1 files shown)


banklesss (approved):

apps/wallet-mobile/src/TxHistory/ActionsBanner.tsx
```diff
@@ -200,25 +206,34 @@ const styles = StyleSheet.create({
...
 })
 
 const useStrings = () => {
   const intl = useIntl()
-  const bold = {b: (text) => <Text bold>{text}</Text>}
+
+  const formattingOptions = {
+    b: (text: ReactNode) => <Text style={[styles.buyInfo, styles.bold]}>{text}</Text>,
+    textComponent: (text: ReactNode) => <Text style={styles.buyInfo}>{text}</Text>,
+  }
```
stackchain: is it part of strings?

michalszorad: `b` is part of strings and `textComponent` is the default one

stackchain: imho, we should add styles only at the component, not in the strings, it looks odd to have bold here in the translations, is there a way to leave this decision up the component rendering the text?

michalszorad: The initial bug was that we had the following snippet, but the `lineHeight` does not propagate properly to it's children and the children had different `lineHeight`. 

```tsx
<Text style={{lineHeight: 24}}>
{["Text", <Text bold>Benxa</Text>, "Other text"]}
</Text>
```

To answer your question, we could pass it as a prop. Eg the following
```tsx
// ...

 <Text>
  {strings.buyInfo({
    b: (text: ReactNode) => <Text style={[styles.buyInfo, styles.bold]}>{text}</Text>,
    textComponent: (text: ReactNode) => <Text style={styles.buyInfo}>{text}</Text>,
  })}
</Text>

 // ...
    buyInfo: (options) => intl.formatMessage(actionMessages.buyInfo, options),
// ...
```
What are your thoughts? @stackchain

michalszorad: That's now updated

PR #2876, opened 2023-11-07
fix: Go to select token when clicking disabled input

jorbuedo:
https://emurgo.atlassian.net/browse/YOMO-939

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
+++ b/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
@@ -60,7 +60,7 @@ export const AmountCard = ({
         {label != null && <Text style={[styles.label, !isEmptyString(error) && styles.labelError]}>{label}</Text>}
 
         <View style={styles.content}>
-          <Pressable style={styles.amountWrapper} onPress={focusInput}>
+          <Pressable style={styles.amountWrapper} onPress={focusInput} {...(noTokenSelected && {onPress: navigateTo})}>
             <TextInput
               keyboardType="numeric"
               autoComplete="off"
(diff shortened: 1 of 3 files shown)


banklesss (approved):

apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
```diff
@@ -60,7 +60,7 @@ export const AmountCard = ({
         {label != null && <Text style={[styles.label, !isEmptyString(error) && styles.labelError]}>{label}</Text>}
```
stackchain, line 61: maybe, (p.s dunno if the exactly same thing)
```
const handleOnPress = () => noTokenSelected ? navigateTo() : focusInput()
```

jorbuedo: Would do the same thing, just moves the condition to the render or the on press

stackchain: if it is the same thing I'd say to implement as handle.. become a little more readable.

apps/wallet-mobile/src/features/Swap/common/SwapFormProvider.tsx
```diff
@@ -325,12 +325,14 @@ const swapFormReducer = (state: SwapFormState, action: SwapFormAction) => {
     switch (action.type) {
       case SwapFormActionType.SellTouched:
         draft.sellQuantity.isTouched = true
+        draft.sellQuantity.displayValue = ''
```
stackchain, line 328: Did product request it? That's sad. We were recovering from any inputted number.

jorbuedo: This happens before the calculations, resets the field, but then it's calculated by the swap state and we get the value. It just removes an edge case that was causing some confusion.

stackchain (approved):

PR #2877, opened 2023-11-07
fix(swap): modal swip down gesture

banklesss:


SorinC6 (approved):

apps/wallet-mobile/src/components/Modal/ModalScreen.tsx
```diff
@@ -16,20 +17,27 @@ import {useModal} from './ModalContext'
...
+
+  const clearDirectionCount = () => {
     setDownDirectionCount(0)
   }
 
   const onResponderMove = ({nativeEvent}: GestureResponderEvent) => {
     if (swipeLocationY < nativeEvent.locationY) {
       const newState = downDirectionCount + 1
-      if (newState > 4) {
+      if (newState > 4 && isOpen) {
```
stackchain: what 4 and 1 means? can have constants? avoid magic numbers.

apps/wallet-mobile/src/components/Modal/ModalScreen.tsx
```diff
@@ -16,20 +17,27 @@ import {useModal} from './ModalContext'
 
 export const ModalScreen = () => {
   const {current} = useCardAnimation()
-  const {height, closeModal, content} = useModal()
+  const {height, closeModal, content, isOpen} = useModal()
```
stackchain, line 19: won't require changes in all screens that consume it?

banklesss: Not really because it is just a variable to know the state of the modal.

The rest of the places where the modal is used only use the open modal / close modal actions

banklesss: Another way to do it would have been to check if the context variable `content` is `undefined` or not, but I think this is more explicit.

PR #2878, opened 2023-11-07
fix: total value in summary

jorbuedo:
https://emurgo.atlassian.net/browse/YOMO-952

### DIFF
--- a/packages/swap/src/helpers/orders/factories/makeOrderCalculations.ts
+++ b/packages/swap/src/helpers/orders/factories/makeOrderCalculations.ts
@@ -154,6 +154,19 @@ export const makeOrderCalculations = ({
       ]),
     }
 
+    const ptTotalValueSpent: Balance.Amount | undefined = Quantities.isZero(
+      sellInPtTerms,
+    )
+      ? undefined
+      : {
+          tokenId: tokens.ptInfo.id,
+          quantity: Quantities.sum([
+            pool.batcherFee.quantity,
+            frontendFeeInfo.fee.quantity,
+            sellInPtTerms,
+          ]),
+        }
+
     const result: SwapOrderCalculation = {
       order: {
         side,
@@ -176,6 +189,7 @@ export const makeOrderCalculations = ({
       },
       buyAmountWithSlippage,
       hasSupply,
+      ptTotalValueSpent,
       prices: {
         base: priceBase,
         market: marketPrice,
--- a/apps/wallet-mobile/src/TxHistory/BalanceBanner.tsx
+++ b/apps/wallet-mobile/src/TxHistory/BalanceBanner.tsx
@@ -3,13 +3,13 @@ import {StyleSheet, Text, TouchableOpacity, View} from 'react-native'
 
 import {Boundary, ResetErrorRef, Spacer} from '../components'
 import {Icon} from '../components/Icon'
+import {PairedBalance} from '../components/PairedBalance/PairedBalance'
 import {usePrivacyMode} from '../features/Settings/PrivacyMode/PrivacyMode'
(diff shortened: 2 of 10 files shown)


banklesss (approved):

apps/wallet-mobile/src/TxHistory/PairedBalance.tsx
stackchain, line 1: since now it's shared, let's move it to `/components`

apps/wallet-mobile/src/TxHistory/PairedBalance.tsx
```diff
@@ -14,8 +14,9 @@ import {Quantities} from '../yoroi-wallets/utils'
 type Props = {
   amount: Balance.Amount
   isPrivacyOff?: boolean
+  style?: TextStyle
 }
-export const PairedBalance = React.forwardRef<ResetErrorRef, Props>(({isPrivacyOff, amount}, ref) => {
+export const PairedBalance = React.forwardRef<ResetErrorRef, Props>(({isPrivacyOff, amount, style}, ref) => {
```
stackchain: maybe to don't pass the same style to different components, or even many styles errorStyle, amountStyle, create a `variant`?

packages/swap/src/helpers/orders/factories/makeOrderCalculations.ts
```diff
@@ -154,6 +154,19 @@ export const makeOrderCalculations = ({
       ]),
     }
 
+    const ptTotalValueSpent: Balance.Amount | undefined = Quantities.isZero(
```
stackchain, line 157: the reasoning behind Amounts to don't be undefined is to don't sum up (undefined + undefined) = 0, since here is a derivation of total PT, I'd say it's ok, but if its ever needed consider adding it to a sum, then we should remove the undefined

stackchain: shouldn't it always add the fees? maybe is a UI issue not state. dunno

SorinC6 (approved):

packages/swap/src/helpers/orders/factories/makeOrderCalculations.ts
jorbuedo: Problem here is that the price in ada can be missing, so all of this field needs to be undefined. Either all the value is there, or it doesn't represent the same thing.

apps/wallet-mobile/src/TxHistory/PairedBalance.tsx
jorbuedo: They me be different components, but they're all Text in the end. And all of them sharing the same style already. What's the benefit of variants over style? I would have to make the variants here instead of on usage.

stackchain: by reading the component api `style` it looks to me it will be applied only at the root component, such as View/Text/Button from react-native, and not into child components. It's been applied for 3 different components, the BalanceError + Amount, so if you add some padding it will add padding for everything. If it's too specific to Swap, copy into Swap.

jorbuedo: But they are not different components, they are all the same one. It's just that the component is over-complicated with sub components when all it does is either render null or Text, and change the value inside the Text. But the styles of the Text are always the same, and are the ones I want to override from outside.
If I created a variant it would have to be applied to all the sub components and check on each what variant it is.

stackchain: k, still, this `style` is not applied to the `PairedBalance` it's applied to texts, by reading the component API this is not clear. I'd say whether go with a variant, or pass it as `textsStyle`

stackchain (approved):

PR #2879, opened 2023-11-07
feat(screen-share): Create screen-share setting

michalszorad:
Resolves https://emurgo.atlassian.net/browse/YOMO-960


apps/wallet-mobile/src/features/ScreenShare/index.ts
stackchain, line 1: if it's only accessible through settings, move into settings.

michalszorad: Catalyst QrCode component uses `useAllowScreenshots` imported from here, should we move it anyway?

stackchain: we need a place for the mobile hooks, and not the wallet hooks, I'd say this is a mobile hook, add to the utils for now.

stackchain (approved):
lgtm

michalszorad:
Still WIP, needs translations @stackchain

michalszorad:
Reay to be reviewed & merged

PR #2890, opened 2023-11-09
chore: swap/add error boundary for orders

SorinC6:
- add error new component for server error in orders: https://emurgo.atlassian.net/browse/YOMO-940
- add missing cursor style


banklesss (approved):

stackchain:
just a quick heads up, this error boundary will be capturing whatever error is thrown and displaying as service unavailable for now.

PR #2892, opened 2023-11-09
feat(swap): add export for the SwapContext variable

nistadev:
Export the context of the swap to use it with hooks on react: `const context = useContext(SwapContext)`


jorbuedo:
@nistadev Try using `useSwap`, it's a hook exporting the context already.

PR #2894, opened 2023-11-09
feature: Use cdn for NFT images

jorbuedo:
|||||
|-|-|-|-|
|image|image|image|image|

### DIFF
--- a/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
+++ b/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
@@ -1,95 +1,64 @@
 import {isString} from '@yoroi/common'
 import {Balance} from '@yoroi/types'
-import React, {useEffect, useState} from 'react'
-import {ErrorBoundary} from 'react-error-boundary'
-import {Image, ImageResizeMode, ImageStyle, StyleProp, View} from 'react-native'
+import React from 'react'
+import {Image, ImageStyle, StyleProp, View} from 'react-native'
 import SkeletonPlaceholder from 'react-native-skeleton-placeholder'
-import {SvgUri} from 'react-native-svg'
 
 import placeholder from '../../assets/img/nft-placeholder.png'
-import {getNftFilenameMediaType, getNftMainImageMediaType, isSvgMediaType} from '../../yoroi-wallets/cardano/nfts'
+import {useSelectedWallet} from '../../SelectedWallet'
+import {getNftMainImageMediaType} from '../../yoroi-wallets/cardano/nfts'
+import {useNativeAssetImage} from '../../yoroi-wallets/hooks'
+
+type NftPreviewProps = {
+  nft: Balance.TokenInfo
+  showPlaceholder?: boolean
+  style?: StyleProp<ImageStyle>
+  height: number
+  width: number
+  resizeMode?: 'cover' | 'contain'
+  blurRadius?: number
+  zoom?: 1 | 2 | 3
+}
 
 export const NftPreview = ({
   nft,
   showPlaceholder,
   style,
-  showThumbnail,
   height,
   width,
-  resizeMode,
+  resizeMode = 'cover',
   blurRadius,
-}: {
-  nft: Balance.TokenInfo
-  showPlaceholder?: boolean
-  style?: StyleProp<ImageStyle>
-  showThumbnail?: boolean
-  height: number
-  width: number
-  resizeMode?: ImageResizeMode
-  blurRadius?: number
-}) => {
-  const [error, setError] = useState(false)
-  const [loading, setLoading] = useState(true)
-  const uri = showThumbnail ? nft.icon : nft.image
-  const isUriSvg =
-    isString(uri) &&
-    (uri.toLowerCase().endsWith('.svg') ||
-      isSvgMediaType(getNftMainImageMediaType(nft)) ||
-      isSvgMediaType(getNftFilenameMediaType(nft, uri)))
+  zoom = 1,
+}: NftPreviewProps) => {
+  const wallet = useSelectedWallet()
+  const [policy, name] = nft.id.split('.')
+  const {uri, isLoading, isError} = useNativeAssetImage({
+    networkId: wallet.networkId,
+    policy,
+    name,
+    width: width * zoom,
+    height: height * zoom,
+    kind: 'metadata',
+    resizeMode,
+    mediaType: getNftMainImageMediaType(nft),
+  })
 
-  const shouldShowPlaceholder = !isString(uri) || showPlaceholder || (isUriSvg && blurRadius !== undefined) || error
-
-  useEffect(() => {
-    setLoading(true)
-  }, [uri])
-
-  if (shouldShowPlaceholder) {
-    // Since SvgUri does not support blur radius, we show a placeholder
-    return <PlaceholderImage height={height} style={style} width={width} resizeMode={resizeMode} />
-  }
+  const shouldShowPlaceholder = !isString(uri) || showPlaceholder || isError
 
   return (
-    <ErrorBoundary fallback={<PlaceholderImage height={height} style={style} width={width} resizeMode={resizeMode} />}>
-      <View style={{width, height, overflow: 'hidden'}}>
-        {loading ? (
-          <SkeletonPlaceholder enabled={true}>
-            <View style={{width, height}} />
-          </SkeletonPlaceholder>
-        ) : null}
-
-        {isUriSvg ? (
-          <SvgUri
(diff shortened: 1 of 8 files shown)


jorbuedo:
Waiting for resize method to be added to the cdn

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -960,3 +961,39 @@ export function useHideBottomTabBar() {
...
+}
+export const useNativeAssetImage = ({
+  networkId,
+  fingerprint,
+  width,
+  height,
+  kind = 'metadata',
+}: NativeAssetImageRequest) => {
+  const network = networkId === 300 ? 'preprod' : 'mainnet'
+  const env = Config.BUILD_VARIANT === 'PROD' ? 'prod' : 'dev'
```
michalszorad: Maybe worth reusing `isProduction()` utility

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -960,3 +961,39 @@ export function useHideBottomTabBar() {
...
+  const query = useQuery({
+    staleTime: Infinity,
+    queryKey: ['native-asset-img', fingerprint, `${width}x${height}`],
+    queryFn: async () => {
+      const response = await fetch(
+        `https://${network}.cardano-nativeassets-${env}.emurgornd.com/${fingerprint}?width=${width}&height=${height}&kind=${kind}`,
```
michalszorad: We have a package `api`, so wondering whether it would be a better place for the fetching logic

stackchain: I'd say for this update it is ok, since the API still need a lot of work.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: Actually according to Pedro we will only use `prod`, regardless of the app's build. But only once the cdn is stable. So the env line will go away.

jorbuedo:
Latest commit untested (preprod db is down)
Will finish Monday

apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
```diff
@@ -1,95 +1,61 @@
...
+}: NftPreviewProps) => {
+  const wallet = useSelectedWallet()
+  const {uri, isLoading, isError} = useNativeAssetImage({
+    networkId: wallet.networkId,
+    fingerprint: nft.fingerprint,
+    width: width * zoom,
+    height: height * zoom,
+    kind: 'metadata',
+    resizeMode,
+    mediaType: (nft.metadatas.mintNft as {mediaType?: string})?.mediaType,
```
michalszorad: Maybe `getNftMainImageMediaType` can be reused

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -960,3 +960,46 @@ export function useHideBottomTabBar() {
     return () => navigation.getParent()?.setOptions({tabBarStyle: true, tabBarVisible: undefined})
   }, [navigation])
 }
+
+const supportedTypes = ['image/png', 'image/jpeg', 'image/gif', 'image/webp', 'image/svg+xml', 'image/tiff']
```
michalszorad: These are supported by backend? What do we show if NFT has unsupported media type?

jorbuedo: Yes! That's the backend list. This is here to not do unnecessary calls to the backend when we know it won't support the image.
If we ask for an unsupported one we get an error. If `enabled` is false (because it's not in this list) react query returns initialData, and since none is provided, it's undefined. 

Both for errors and undefined data, preview shows the placeholder image.

apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
jorbuedo: Sure, didn't know there was a function to access that property

michalszorad (approved):

jorbuedo:
Backend in process of updating, may need to change things.

stackchain (approved):

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -960,3 +960,51 @@ export function useHideBottomTabBar() {
...
+  kind = 'metadata',
+}: NativeAssetImageRequest) => {
+  const network = networkId === 300 ? 'preprod' : 'mainnet'
+  const isMediaTypeSupported = supportedTypes.includes(mediaType.toLocaleLowerCase())
+  const query = useQuery({
+    staleTime: Infinity,
+    queryKey: ['native-asset-img', policy, name, `${width}x${height}`, resizeMode],
+    queryFn: async () => {
+      const response = await fetch(
+        `https://${network}.cardano-nativeassets-prod.emurgornd.com/${policy}/${name}?width=${width}&height=${height}&kind=${kind}&fit=${resizeMode}`,
```
stackchain, line 1002: TODO:
- [ ] Move to the wallet api interface `.getTokenImage`

jorbuedo:
@stackchain Ready to merge from my side

stackchain:
> @stackchain Ready to merge from my side

I think on FE is done, still some images are not going through the CDN, holding for a bit while BE figures it out.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -960,3 +960,86 @@ export function useHideBottomTabBar() {
...
+
+type NativeAssetInvalidationRequest = {
+  networkId: NetworkId
+  policy: string
+  name: string
+}
+export const useNativeAssetInvalidation = ({networkId, policy, name}: NativeAssetInvalidationRequest) => {
+  const network = networkId === 300 ? 'preprod' : 'mainnet'
+  const mutation = useMutationWithInvalidations({
+    invalidateQueries: [['native-asset-img', policy, name]],
```
stackchain, line 1029: @jorbuedo I don't think `useMutationWithInvalidations` will work here, it was design to work the same way `react-query` does, it requires a match `1:1` to the `queryKey`, and in the key besides the current info we also have the `dimensions` and the `resizeMode`. I believe you will have to implement something like `const queryKeys = queryClient.getQueryCache().findAll().filter(..)` I'm not aware if they have implemented wildcards `['key', '*']` I don't think so, which would be fantastic.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -960,3 +960,86 @@ export function useHideBottomTabBar() {
     return () => navigation.getParent()?.setOptions({tabBarStyle: true, tabBarVisible: undefined})
   }, [navigation])
 }
+
+const supportedTypes = [
+  'img/png', // Yeah, someone minted that
```
stackchain, line 965: 👍🏻

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: Query invalidation doesn't need to be exact
https://tanstack.com/query/v4/docs/react/guides/query-invalidation

We use `['native-asset-img', policy, name, `${width}x${height}`, resizeMode]` as query key.
Using `['native-asset-img', policy, name]` will invalidate all the size and resizeMode for the image. 

useMutationWithInvalidations loops an array and sends each content to the react-query invalidation, that's why I send it with double array `[['native-asset-img', policy, name]]` 
Seems to work fine, after invalidation the images get re-fetched immediately.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: I don't love to have to list all the supported types here, but it's better than trying to fetch videos, audios, and any other thing we know won't work.
Unfortunately it means wrongly typed things like `img/png` need to be here or won't load.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: @stackchain

stackchain: nice! ty @jorbuedo

PR #2896, opened 2023-11-09
fix(swap): apply proper svg to swap order lists

banklesss:
Related to [YOMO-958](https://emurgo.atlassian.net/browse/YOMO-958)

[YOMO-958]: https://emurgo.atlassian.net/browse/YOMO-958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


michalszorad (approved):

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/EmptyOpenOrdersLogo.tsx
stackchain, line 1: hey @banklesss since the useCase show empty complete order is covered inside the completed orders, I'd like to suggest few things, are nitpicks so if you want to skip is fine. 

- [x] Move the LOGOs -> to a `../common/illustrations` (rename them to `<whatever>Ilustration.tsx`) I think it denominates more accurately the itention, even though they are not shared, I think by collocating all Illustration in the same place it will make simpler further updates, it has plenty already (for the connection error/transaction success/etc...) it's scattered lets collocate them. Or maybe we can create a convention to locally `./illustrations` and add it there, up to you, and keep it collocated inside the component since is not shared. 
- [x] Be mindful when updating the components to update the styles along with it, the `NoOrdersYet` styles keys are with the old names `image` `imageContainter` it should be renamed to match new components.

stackchain (approved):
lgtm

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/EmptyOpenOrdersLogo.tsx
banklesss: done

PR #2901, opened 2023-11-14
fix: swap UAT CR

SorinC6:
task:
- https://emurgo.atlassian.net/browse/YOMO-973
- https://emurgo.atlassian.net/browse/YOMO-971
- https://emurgo.atlassian.net/browse/YOMO-970 - for the cards box shadow not sure how to create the exact shadow as in the design-> from what I see there will always be a small difference between IOS/ANDROID on this matter 
- https://emurgo.atlassian.net/browse/YOMO-969
- https://emurgo.atlassian.net/browse/YOMO-976
- https://emurgo.atlassian.net/browse/YOMO-905 - change precision to 10
- https://emurgo.atlassian.net/browse/YOMO-977 - force/add gray color when input is 0

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
+++ b/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
@@ -158,6 +158,7 @@ const styles = StyleSheet.create({
   },
   borderError: {
     borderColor: COLORS.ALERT_TEXT_COLOR,
+    borderWidth: 2,
   },
   active: {
     borderWidth: 2,
--- a/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
+++ b/apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
@@ -70,7 +70,7 @@ export const AmountCard = ({
               onChangeText={onChange}
               allowFontScaling
               selectionColor={isFocused ? '#242838' : COLORS.TRANSPARENT_BLACK}
-              style={styles.amountInput}
+              style={[styles.amountInput, value === '0' && styles.grayText]}
               underlineColorAndroid="transparent"
               ref={inputRef}
               editable={inputEditable && touched}
@@ -214,5 +214,8 @@ const styles = StyleSheet.create({
   errorText: {
     color: COLORS.ALERT_TEXT_COLOR,
     fontSize: 12,
+  },
+  grayText: {
+    color: '#6B7384',
   },
 })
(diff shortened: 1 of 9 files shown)


apps/wallet-mobile/src/features/Swap/common/helpers.ts
```diff
@@ -99,3 +99,11 @@ export const sortTokensByName = (a: Balance.TokenInfo, b: Balance.TokenInfo, wal
 
   return nameA.localeCompare(nameB, undefined, {sensitivity: 'base'})
 }
+
+export const switchDayAndMonth = (dateString: string): string => {
```
stackchain: hey @SorinC6 is there a ticket for this one? cuz this should be based on the `Intl` I don't think we should update it.

stackchain: please ping Product.

SorinC6: @stackchain  the fix is for this task: https://emurgo.atlassian.net/browse/YOMO-976 . Did not find a way to made it as it the design only if switching them

jorbuedo: If we can tell the formatter to use 2 digits for years, that may be done. But changing the order of the months and days or whatever... that's locale based, not design based.

apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
```diff
@@ -70,7 +70,7 @@ export const AmountCard = ({
               onChangeText={onChange}
               allowFontScaling
               selectionColor={isFocused ? '#242838' : COLORS.TRANSPARENT_BLACK}
-              style={styles.amountInput}
+              style={[styles.amountInput, value === '0' && styles.grayText]}
```
jorbuedo, line 73: I don't think this one's what they want.

SorinC6: have gone back and forward regarding it - they want the grey color even when the input is 0 .. at least this is what I understand... can discuss more related to it here https://emurgo.atlassian.net/browse/YOMO-977 @jorbuedo

apps/wallet-mobile/src/features/Swap/common/helpers.ts
SorinC6: @jorbuedo not sure how to achieve it without switching the order 🤔

jorbuedo: We shouldn't be deciding any order. There's date formatters that take into account the locale. I'm pretty sure javascript comes with one by default. But we also have one installed 
`@formatjs/intl-datetimeformat` So whatever that thing says goes.

stackchain: can you check it @SorinC6 ?

stackchain (approved):

PR #2907, opened 2023-11-20
Release/test ci version bump

crypto-titan:


banklesss (approved):

stackchain (approved):

apps/wallet-mobile/.bundle/config
```diff
@@ -1,2 +1,5 @@
-BUNDLE_PATH: "vendor/bundle"
-BUNDLE_FORCE_RUBY_PLATFORM: 1
+---
+BUNDLE_PATH: "/Users/runner/work/yoroi/yoroi/apps/wallet-mobile/vendor/bundle"
```
michalszorad, line 2: Is the path supposed to be hardcoded?

crypto-titan: it was automatically generated by the CI

crypto-titan: not really sure if we need it though, but on the next bump it will probably come back.

michalszorad (approved):

PR #2911, opened 2023-11-20
feature(links): new links package

stackchain:
### Abstract
Links package is responsible to create and parse crypto URLs.

It adds support to the legacy payment request of `CIP13`, which was missing the `amount`, `memo` and `message` proper validation. The legacy payment request of CIP13 requires some denormalization of the URL to build and parse it properly, since it uses the _authority_ to hold the `address`.
It also adds support for the `CIP99` links proposal, the `claim` _authority_ to allow airdrops, PoO (Proof of Onboarding) or any kind of token distribution.

#### Missing
`/stake` _authority_, Yoroi for now supports only 1 stake pool delegation, multi delegation is not implemented on the lib side and it is part of `CIP13`.
`/transfer` _authority_ with MAT (multi-asset transfer) to support the new `CIPXXXX` (1399) proposal (only internal for now)

#### Extra
It adds support for the Receive funnel to create payment request links with `amount`, `memo` and `message`, before only the `address` was presented to another user when requesting a payment.


michalszorad:
Can we add the description from this PR to a README in this package?

stackchain:
> Can we add the description from this PR to a README in this package?

`#00FF00` done.

michalszorad (approved):

PR #2913, opened 2023-11-21
fix(swap): Prevent page viewed event from accidentally firing

michalszorad:
Resolves [YOMO-998](https://emurgo.atlassian.net/browse/YOMO-998)

The issue is that when opening a modal, we call `navigation.navigate('modal')` which then emits `blur` event used by `useFocusEffect`. The internal `focus` state inside the `useFocusEffect` is then reset. Once we close the modal, the original screen is focused again. That's why every time someone closes a modal, the `useFocusEffect` executes it's callback.

To fix the issue I've created a ref, that holds latest navigation state. The state does not change when opening/closing modal and that's why the check `if (currentState === previousState) return` resolves the issue.

[YOMO-998]: https://emurgo.atlassian.net/browse/YOMO-998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/OpenOrders.tsx
+++ b/apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/OpenOrders.tsx
@@ -86,6 +86,8 @@ export const OpenOrders = () => {
   const navigation = useNavigation()
 
   const trackSwapConfirmPageViewed = React.useCallback(() => {
+    // Closing a modal triggers this callback.
+    // https://github.com/Emurgo/yoroi/pull/2913
     const currentState = navigation.getState()
     const previousState = navigationRef.current
     if (currentState === previousState) return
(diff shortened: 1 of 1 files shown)


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/OpenOrders.tsx
```diff
@@ -81,11 +83,19 @@ export const OpenOrders = () => {
...
   const {track} = useMetrics()
 
-  useFocusEffect(
-    React.useCallback(() => {
-      track.swapConfirmedPageViewed({swap_tab: 'Open Orders'})
-    }, [track]),
-  )
+  const navigation = useNavigation()
+
+  const trackSwapConfirmPageViewed = React.useCallback(() => {
```
stackchain, line 88: again, funnels on modals make me sick.

stackchain (approved):

PR #2919, opened 2023-11-23
chore: dropped sync on focus

stackchain:
#### Summary

Some of the funnels didn't work properly in the past and the app crashes if for example a token vanished from the wallet while the user was inside the funnel
Most of these crashes was handled and nothing new is showing up on Sentry
Community constantly complains about the spinning spinning while landing on tx history

#### To do
- [ ] Enable buttons while syncing 
- [ ] Lock Screen when the first sync is happening (UI rebuild, restore/create wallet)
- [x] Remove refresh on focus since it happens every 20s and the user can pull to refresh


stackchain:
will be refactored along with WalletManager module.

PR #2920, opened 2023-11-24
feature(claim): cip99

stackchain:
Adds support to Cip99 Claims


apps/wallet-mobile/src/components/Modal/ModalContext.tsx
```diff
@@ -6,10 +6,13 @@ type ModalState = {
   isOpen: boolean
   title: string
   content: React.ReactNode
+  isLoading: boolean
```
stackchain, line 9: since the modal is out of the context tree added the loading to block interactions to the state

apps/wallet-mobile/src/components/ModalScreenWrapper/ModalScreenWrapper.tsx
stackchain, line 1: the current modal implementation is out of the context scope of the wallet, this is a derived version of the current modal but to be implemented within the navigator, it will required to wrap all navigators within a new navigator to overlap the current rendered one and to update all navigations routes. It was working on this PR, but some edge cases were delaying the PR, created a task to update it later.

packages/common/src/api/fetchData.ts
```diff
@@ -0,0 +1,106 @@
...
+ *   method: 'post',
+ *   data: { /* some data *\/ }
+ * }).then(response => {
+ *   // Handle response
+ * }).catch(error => {
+ *   // Handle error
+ * })
+ * ```
+ */
+export const fetchData: FetchData = <T, D = any>(
```
stackchain, line 66: legacy `fetch` was designed to throw errors, which is not great, so old `fetch` now is deprecated and whether the API errors should be throw or ignored will be a implementation decision.

packages/common/src/api/fetcher.ts
```diff
@@ -1,6 +1,10 @@
 import axios, {AxiosRequestConfig, AxiosResponse} from 'axios'
+
 import {ApiError, NetworkError} from '../errors/errors'
 
+/**
+ * @deprecated This function is deprecated and will be removed in a future release. Use `fetchData` instead.
```
stackchain, line 6: marked as deprecated.

packages/common/src/api/handleApiError.ts
```diff
@@ -0,0 +1,32 @@
...
+    case 409:
+      throw new Api.Errors.Conflict(error.message)
+    case 410:
+      throw new Api.Errors.Gone(error.message)
+    case 425:
+      throw new Api.Errors.TooEarly(error.message)
+    case 429:
+      throw new Api.Errors.TooManyRequests(error.message)
+    default:
+      throw new Api.Errors.Unknown(error.message)
```
stackchain, line 30: any particular http status code that needs to be handled will require specific error otherwise it will fallback to unknown.

apps/wallet-mobile/src/TxHistory/TxHistoryNavigator.tsx
```diff
@@ -80,216 +83,232 @@ export const TxHistoryNavigator = () => {
...
-              options={{
-                title: strings.listAmountsToSendTitle,
-                ...sendOptions,
+        <ClaimProvider key={wallet.id} claimApi={claimApi}>
+          <SwapFormProvider>
+            <Stack.Navigator
+              screenListeners={{}}
+              screenOptions={{
+                ...defaultStackNavigationOptions,
+                detachPreviousScreen: true /* https://github.com/react-navigation/react-navigation/issues/9883 */,
```
michalszorad, line 106: The workaround in that issue mentions it needs to be set to `false`. Does the flashing not appear anymore? If not, we can remove the comment as well I guess

apps/wallet-mobile/src/components/Modal/ModalContext.tsx
michalszorad: Could `openModal(title, <Provider manager={manager}><ModalContent /></Provider>)` help? This way we won't have to add new props to the modal context

packages/common/src/api/fetcher.ts
banklesss: i am using it in the resolver package 👍🏻

apps/wallet-mobile/src/features/Claim/useCases/AskConfirmation.tsx
```diff
@@ -0,0 +1,122 @@
...
+  },
+  monospace: {
+    ...Platform.select({
+      ios: {fontFamily: 'Menlo'},
+      android: {fontFamily: 'monospace'},
+    }),
+  },
+})
+
+function getDomain(url) {
```
michalszorad: The `url` may be implicit any

banklesss (approved):

jorbuedo (approved):

apps/wallet-mobile/src/components/Modal/ModalContext.tsx
stackchain: I was tempted of doing that, if we were using redux to manage the requests I’d have gone for it, but the way it’s structured now by leveraging transitions states with react query it was not looking right, cuz it will reflect the query state only, basically 1:1 to query, also to block the screen the LoadingOverlay will have to sit on Modal container anyway. How are we preventing further interactions with the screen in other funnels that are leveraging modal ? (This will be fixed when we migrate to the modal wrapper component, but for that we will need to refactor the navigation)

packages/common/src/api/fetcher.ts
stackchain: @banklesss definitely not a issue for now, the APIError that is throw by it is ignored at the root level of the APP. I think we wont display any error to the user if the resolver request can’t go through (fail silently)

apps/wallet-mobile/src/TxHistory/TxHistoryNavigator.tsx
stackchain: aaaa my bad, I changed when testing the new modal within the context, I forgot to change it back, nice catch

stackchain: aaaa my bad, I changed when testing the new modal within the context, I forgot to change it back, nice catch

apps/wallet-mobile/src/features/Claim/useCases/AskConfirmation.tsx
stackchain: Nice catch, fixing it

stackchain: Why TS didn’t complain about it 🤔

michalszorad: `"noImplicitAny": false,` in tsconfig

stackchain: wow boy, I did it while bumping RN :/

stackchain: created a task on Jira, now there are thousand places to fix it.

michalszorad (approved):

PR #2961, opened 2024-01-03
refactor(nfts): Use new supply request to fetch asset supplies

michalszorad:
Resolves [YOMO-1013](https://emurgo.atlassian.net/jira/software/c/projects/YOMO/issues/YOMO-1013)

[YOMO-1013]: https://emurgo.atlassian.net/browse/YOMO-1013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
+++ b/apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
@@ -8,7 +8,7 @@ import {fetchTokensSupplies} from './assetSuply'
 import fetchDefault from './fetch'
 import {toAssetNameHex, toPolicyId} from './utils'
 
-export const getNFTs = async (ids: `${string}.${string}`[], config: BackendConfig): Promise<Balance.TokenInfo[]> => {
+export const getNFTs = async (ids: string[], config: BackendConfig): Promise<Balance.TokenInfo[]> => {
   if (ids.length === 0) {
     return []
   }
@@ -22,14 +22,14 @@ export const getNFTs = async (ids: `${string}.${string}`[], config: BackendConfi
 
   const [assetMetadatas, assetSupplies] = await Promise.all([
     fetchDefault<unknown>('multiAsset/metadata', payload, config),
-    fetchTokensSupplies(ids, config),
+    fetchTokensSupplies(ids as Array<`${string}.${string}`>, config),
   ])
 
   const possibleNfts = parseNFTs(assetMetadatas, config.NFT_STORAGE_URL)
   return possibleNfts.filter((nft) => assetSupplies[nft.id] === '1')
 }
 
-export const getNFT = async (id: `${string}.${string}`, config: BackendConfig): Promise<Balance.TokenInfo | null> => {
+export const getNFT = async (id: string, config: BackendConfig): Promise<Balance.TokenInfo | null> => {
   const [nft] = await getNFTs([id], config)
   return nft || null
 }
--- a/apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
+++ b/apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
@@ -22,7 +22,7 @@ export const getNFTs = async (ids: string[], config: BackendConfig): Promise<Bal
 
   const [assetMetadatas, assetSupplies] = await Promise.all([
     fetchDefault<unknown>('multiAsset/metadata', payload, config),
-    fetchTokensSupplies(ids as Array<`${string}.${string}`>, config),
+    fetchTokensSupplies(ids, config),
   ])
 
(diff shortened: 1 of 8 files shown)


apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
```diff
@@ -22,11 +22,11 @@ export const getNFTs = async (ids: string[], config: BackendConfig): Promise<Bal
 
   const [assetMetadatas, assetSupplies] = await Promise.all([
     fetchDefault<unknown>('multiAsset/metadata', payload, config),
-    fetchTokensSupplies(ids, config),
+    fetchTokensSupplies(ids as Array<`${string}.${string}`>, config),
```
michalszorad: Is there a better way?

packages/api/src/cardano/api/token-supply.ts
```diff
@@ -26,7 +26,7 @@ export const getTokenSupply =
     const payload = {assets: Array.from(assetsMap.values())}
 
     return request<Api.Cardano.TokenSupplyResponse>({
-      url: `${baseUrl}/multiAsset/supply`,
+      url: `${baseUrl}/multiAsset/supply?numberFormat=string`,
```
michalszorad, line 29: We use string because large supplies in number type can cause precision loss

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
stackchain: there is a loop that maps from ids (Balance) to assets {k,v}, I think you can use the same logic to transform from Balance Token Id, to fetchToeknsSupplies id with a type guard, --> `filter(isTypeguard)` by filtering out if is not

jorbuedo: @MichalSzorad  The Api introduced the type change to include the dot. Either `getNFTs` uses that type as input as well, or it gets casted like you did here, or inside `fetchTokensSupplies`. Not very type safe, and not actually handling anything, but it silences the compiler.

If we want to move the type up to the input, then I'd alias it and force it everywhere.
```
type TokenId = `${string}.${string}`
```

Problem is, forcing this pattern from strings on usage will get tedious fast.

The other "safe" alternative is to add, either here or inside `fetchTokensSupplies`, an actual runtime check to make sure the string has the dot, or throw otherwise, then handle the throw possibility.

jorbuedo (approved):
Lgtm. 
Handling possible errors for ids not following the pattern would be going an extra mile, but I'm guessing the backend will return an error anyway if that happens.

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
jorbuedo: The `filter(isTypeguard)` would silently ignore wrong ones, could lead to confusion maybe. Not sure how critical this is.

jorbuedo: I see there's utils around this, you could map the ids with `toTokenId`.
It doesn't really handle incorrect strings other than getting a wrong result, but it will cast the types.
Not sure it's worth it considering there's no error handling either, it would just add more runtime without much benefit.

michalszorad: Thanks for the suggestions, having a look

stackchain: casting is flaky, it wouldn't prevent the API of returning error if the payload is safely parsed on the other side (if it has validators) so for 50 to 500 assets that you can consume at once, if you provide a single one (i.e primary token) as input the entire request would fail. By filtering, and not casting, we are saying that we don't mind about those that are not matching the expected payload type of the API, cuz the API doesn't support it. I believe it boils down to, shall we throw before even firing the request and client deal with it, or as we know what is supported by the API, we get the data that we can? I'd prefer the 2nd one.

stackchain (changes requested):
@MichalSzorad can you also check the nameless tokens, how the app is receiving them from the UTXO endpoint, with or without the dot separator, it needs to check how we are storing and retrieving, dunno if nameless tokens are retrieve as `policyId.` or `policyId`.

michalszorad:
@jorbuedo @stackchain I've introduced `YoroiTokenId` type and `asTokenId` helper, not to cause any backward compatibility issues. The supply endpoint does not work for primary tokens, but this use-case is handled in ShelleyWallet and ByronWallet in the `fetchTokenInfo` method, where we check for `''` and `'ADA'` and return hardcoded values for these tokens.

stackchain (approved):

PR #2966, opened 2024-01-04
fix: lifting action section

banklesss:
Related to [YOMO-913](https://emurgo.atlassian.net/browse/YOMO-913)

[YOMO-913]: https://emurgo.atlassian.net/browse/YOMO-913?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/Staking/DelegationConfirmation/DelegationConfirmation.tsx
+++ b/apps/wallet-mobile/src/Staking/DelegationConfirmation/DelegationConfirmation.tsx
@@ -2,9 +2,9 @@
 import {Balance} from '@yoroi/types'
 import React, {useEffect, useState} from 'react'
 import {defineMessages, useIntl} from 'react-intl'
-import {ScrollView, StyleSheet, View, ViewProps} from 'react-native'
+import {Platform, ScrollView, StyleSheet, View, ViewProps} from 'react-native'
 
-import {KeyboardAvoidingView, Text, ValidatedTextInput} from '../../components'
+import {KeyboardSpacer, Text, ValidatedTextInput} from '../../components'
 import {ConfirmTx} from '../../components/ConfirmTx'
 import {useStakePoolInfoAndHistory} from '../../Dashboard/StakePoolInfo'
 import {debugWalletInfo, features} from '../../features'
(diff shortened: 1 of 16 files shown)


jorbuedo (approved):

SorinC6 (approved):

apps/wallet-mobile/src/Staking/DelegationConfirmation/DelegationConfirmation.tsx
```diff
@@ -54,73 +54,69 @@ export const DelegationConfirmation = () => {
 
   return (
     <View style={styles.container}>
```
stackchain, line 56: is it a screen? if it is it should be wrapped within SafeAreaView.

apps/wallet-mobile/src/Staking/DelegationConfirmation/DelegationConfirmation.tsx
```diff
@@ -54,73 +54,69 @@ export const DelegationConfirmation = () => {
...
+          onSuccess={onSuccess}
+          setUseUSB={setUseUSB}
+          useUSB={useUSB}
+          yoroiUnsignedTx={yoroiUnsignedTx}
+          chooseTransportOnConfirmation
+        />
+      </Actions>
+
+      {/* hack to fix weird KeyboardAvoidingView bug in THIS SCREEN */}
+      {Platform.OS === 'ios' && <KeyboardSpacer />}
```
stackchain, line 119: maybe we should start using View.onLayout  (thinking out loud, ignore it for now)

apps/wallet-mobile/src/Staking/DelegationConfirmation/DelegationConfirmation.tsx
stackchain: Maybe this is the issue?

banklesss: just checking and unfortunately the issue still persists

banklesss: also I tested using paddings instead of margins

banklesss: The problem is that the button stays behind the keyboard until you click to edit the input. When you start editing, the button goes into place.

I tried to force the focus but it didn't work either.

PR #2974, opened 2024-01-08
help: Create new app only for storybook

jorbuedo:
I've created a new app for storybook, trying to run it with at least 1 story.
Eslint and tsc are passing, but not sure how reliably.

Right now running it complains about `build.gradle`. I tried copying it from the other app but no joy.


stackchain:
- Bumping storybook after RN, it will work differently.

PR #2978, opened 2024-01-09
fix(resolver): display domains as NFT

banklesss:
Related to [YOMO-1103](https://emurgo.atlassian.net/browse/YOMO-1103)

[YOMO-1103]: https://emurgo.atlassian.net/browse/YOMO-1103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


michalszorad (commented):
One example of an asset that causes this issue is here 
 https://cexplorer.io/asset/asset1snevgde6wtnsehdjjlm7rrl9faphdkaxwcjlsg

I wonder what encoding is there used to decode/encode the hex name

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
```diff
@@ -62,7 +62,7 @@ export const toPolicyId = (tokenIdentifier: string) => {
   return tokenSubject.slice(0, 56)
 }
 export const toAssetName = (tokenIdentifier: string) => {
-  return hexToUtf8(toAssetNameHex(tokenIdentifier))
+  return hexToUtf8(toAssetNameHex(tokenIdentifier).replace('000de140', ''))
```
michalszorad: Let's maybe add a comment that this refers to CIP68 or a function `transformCIP68AssetName`

michalszorad (approved):

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
```diff
@@ -62,7 +62,7 @@ export const toPolicyId = (tokenIdentifier: string) => {
   return tokenSubject.slice(0, 56)
 }
 export const toAssetName = (tokenIdentifier: string) => {
```
stackchain: is this function only used for the UI? otherwise it can affect for example the API requests, would mind checking?

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
```diff
@@ -125,3 +125,17 @@ export const toTokenFingerprint = ({
...
+
+export const transformCIP0068AssetNameHex = (assetNameHex) => {
+  const refprefix = '000643b0'
+  const usrprefix = '000de140'
+  const stringsToRemove = [refprefix, usrprefix]
+
+  let transformedAssetNameHex = assetNameHex
+
+  stringsToRemove.forEach((str) => {
+    transformedAssetNameHex = transformedAssetNameHex.replace(str, '')
```
jorbuedo: Not sure I understood completely the issue with CIP68.
It sounds like asset names come with a prefix to indicate something?
With this approach, it'll remove both strings, regardless of whether they are the prefix or not.

I would recommend a replace with regex, and prefix it with ` ^` so that it only matches at the beginning.

jorbuedo (dismissed):

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
stackchain: can we rename it `toDisplayAssetName`, cuz the name usually comes from metadata, yet, we fallback to the `assetName` if the metadata is not following any CIP specs.

stackchain:
> One example of an asset that causes this issue is here https://cexplorer.io/asset/asset1snevgde6wtnsehdjjlm7rrl9faphdkaxwcjlsg
> 
> I wonder what encoding is there used to decode/encode the hex name

According to the specs `assetName` must be `ASCII` (we are unpacking as `utf-8`) it needs fixing as well.

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
banklesss: finally a yoroi-lib tool has been used

banklesss:
Blocked because an investigation is ongoing

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
```diff
@@ -61,8 +62,8 @@ export const toPolicyId = (tokenIdentifier: string) => {
   const tokenSubject = toTokenSubject(tokenIdentifier)
   return tokenSubject.slice(0, 56)
 }
-export const toAssetName = (tokenIdentifier: string) => {
-  return hexToUtf8(toAssetNameHex(tokenIdentifier))
+export const toDisplayAssetName = (tokenIdentifier: string) => {
+  return hexToUtf8(toUntaggedAssetNameHex(tokenIdentifier))
```
stackchain: it should drop the `hexToUtf8` cuz it's not following the protocol specs, it needs to consume --> `.asciiName ?? .hexName` directly from the `AssetNameUtils.`

stackchain (changes requested):

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
```diff
@@ -61,8 +64,15 @@ export const toPolicyId = (tokenIdentifier: string) => {
...
+export const toDisplayAssetName = (tokenIdentifier: string) => {
+  return toUntaggedAssetName(tokenIdentifier)
+}
+
+export const toUntaggedAssetName = (tokenIdentifier: string) => {
+  const hexName = toAssetNameHex(tokenIdentifier)
+  const properties = AssetNameUtils.resolveProperties(hexName)
+  const untaggedName = properties.asciiName ?? properties.hexName
+  return untaggedName
 }
```
stackchain: untagged + ascii when possible is the displayed asset name, I think we can collocate.
```suggestion
export const toDisplayAssetName = (tokenIdentifier: string) => {
  const hexName = toAssetNameHex(tokenIdentifier)
  const properties = AssetNameUtils.resolveProperties(hexName)
  const untaggedName = properties.asciiName ?? properties.hexName
  return untaggedName
}

```

banklesss: thanks to your review i just noticed that if we can't get the ascii we need to parse the ` properties.hexName` to utf8

what do you think?

packages/resolver/src/adapters/handle/api.ts
```diff
@@ -104,3 +105,13 @@ export const handleHandleApiError = (error: unknown): never => {
 
   throw error
 }
+
+export const adaHandleDomainNormalizer: Record<
```
stackchain: I dare to say that this is not part of the `api` it is a `translators/handle/` to the UI, move into the `translators/` folder as `adaHandleDomainNormalizer` move the test to `.test.`

packages/resolver/src/index.ts
```diff
@@ -13,6 +13,7 @@ export * from './translators/reactjs/provider/ResolverProvider'
 export * from './utils/isResolvableDomain'
 export * from './utils/isDomain'
 export * from './utils/isNameServer'
+export * from './utils/domainNormalizer'
```
stackchain: `-->` `translators/` root

packages/resolver/src/utils/domainNormalizer.ts
```diff
@@ -0,0 +1,11 @@
+import {adaHandleDomainNormalizer} from '../adapters/handle/api'
+
+export const domainNormalizer = (policyId: string, domain: string) => {
```
stackchain: the facade can be on the root of the `translators` --> UI thing, client thing.

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
stackchain: why we can't get the ascii ?

packages/resolver/src/adapters/handle/api.ts
```diff
@@ -92,6 +92,7 @@ export const isAdaHandleDomain = (value: string) =>
 export const handleApiConfig = {
   mainnet: {
     getCryptoAddress: 'https://api.handle.me/handles/',
+    policyId: 'f0ff48bbb7bbe9d59a40f1ce90e9e9d0ff5002ec48f232b49ca0fb9a',
```
michalszorad, line 96: What is this policy id referring to?

banklesss: $handles assets

do you have any suggestion?

stackchain: @banklesss I think there is more than 1.

stackchain: maybe it should split into `isAdaHandle` -> look into the `policyIds` and if it is -> normalize the name. :shrug:

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
```diff
@@ -8,60 +8,46 @@ import {fetchTokensSupplies} from './assetSuply'
...
 
-    const [policyId, shortName] = id.split('.')
-    const metadata = nftAsset.metadata?.[policyId]?.[shortName]
-    return convertNft({metadata, storageUrl, policyId, shortName: shortName})
-  })
+  const metadata = nftAsset.metadata?.[policyId]?.[nameHex]
+  const nft = convertNft({metadata, storageUrl: config.NFT_STORAGE_URL, policyId, nameHex})
 
-  return tokens.filter(isNonNullable)
+  return assetSupplies[nft.id] === '1' || assetSupplies[nft.id] === '0' ? nft : null
```
michalszorad: The requirement for an asset to be an NFT is that it needs to have supply exactly 1. Why do we allow here supply to be `'0'`?

banklesss: The supplies of $handle is 0 🥲

banklesss: Let me check again because i got that supply before all these changes

apps/wallet-mobile/src/yoroi-wallets/cardano/api/utils.ts
banklesss: I meant, when ascii property is undefined so it fallbacks to hexName

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
banklesss: Maybe the bug was influencing the request

michalszorad:
```
curl --location 'https://api.yoroiwallet.com/api/multiAsset/supply?numberFormat=string' \
--header 'Content-Type: application/json' \
--data '{
    "assets": [
        {
            "nameHex": "000de1406a6176696275656e6f2e616461",
            "policy": "f0ff48bbb7bbe9d59a40f1ce90e9e9d0ff5002ec48f232b49ca0fb9a"
        }
    ]
}'

```
seems to respond with 1

packages/resolver/src/adapters/handle/api.ts
banklesss: apparently there is only one for mainnet https://docs.adahandle.com/official-policy-ids

banklesss: > maybe it should split into `isAdaHandle` -> look into the `policyIds` and if it is -> normalize the name. 🤷

the problem is when is not a normalized ada handle domain. in this case `isAdaHandle` is false

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
banklesss: We need to support CIP68 here too https://github.com/Emurgo/yoroi/blob/develop/packages/api/src/cardano/translators/helpers/getTokenIdentity.ts#L24

Thats the reason the supply was 0

apps/wallet-mobile/src/yoroi-wallets/cardano/api/assetSuply.ts
banklesss, line 1: Removed until CIP68 is applied to the API

packages/resolver/src/adapters/handle/api.ts
banklesss: please let me know if you have any suggestion

apps/wallet-mobile/src/yoroi-wallets/cardano/api/assetSuply.ts
michalszorad: Maybe it would work if we used `nameHex` instead of `name` in the payload for `getTokenSupply` in `packages/api/src/cardano/api/token-supply.ts`

banklesss: Yes you are right. I will check it.

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
stackchain: is it the last update pending for this PR?

apps/wallet-mobile/src/yoroi-wallets/cardano/api/assetSuply.ts
banklesss: this can be addressed in this ticket https://emurgo.atlassian.net/browse/YOMO-1109

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
banklesss: changing it . Pushing in a few minutes

stackchain (approved):
updates in the api package will be done in another ticket

PR #2980, opened 2024-01-10
fix: Remove screen transition top-to-bottom for buy/sell screen

jorbuedo:
realted: https://emurgo.atlassian.net/browse/YOMO-1045

> it was an experimental transition we agreed on to apply and test 
> ...
> Considering the fact that the rest transitions are right-to-left, I’d advice to re-use the right-to-left one

I'd like to point out that the default transition is not right-to-left for android. 

In any case, I saved up this top-to-bottom one in themes.

### DIFF
--- a/apps/wallet-mobile/src/theme/transitions/screenOptions.ts
+++ b/apps/wallet-mobile/src/theme/transitions/screenOptions.ts
@@ -1,0 +1,55 @@
+import {Easing} from 'react-native-reanimated'
+
+/**
+ * Options to be added in Stack.Screen to change transition
+ * see more: https://reactnavigation.org/docs/stack-navigator/#animations
+ */
+
+const topToBottomTransitionSpec = {
+  open: {
+    animation: 'spring',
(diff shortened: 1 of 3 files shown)


SorinC6 (approved):

apps/wallet-mobile/src/theme/transitions/screenOptions.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+   * Usage:
+   *    <Stack.Screen
+            options={{
+                ...topToBottom
+            }}
+        />
+   */
+export const topToBottom = {
```
michalszorad, line 52: Where is this used?

stackchain: It was RFU.

stackchain: in the comments ⬆️

PR #2983, opened 2024-01-10
feat(theme): Add package structure

SorinC6:
task: https://emurgo.atlassian.net/browse/YOMO-972


apps/wallet-mobile/src/features/RampOnOff/common/AmountCard/AmountCard.tsx
```diff
@@ -193,6 +190,9 @@ const getStyles = (props: {theme: Theme}) => {
       color: theme.color.magenta[500],
       fontSize: 12,
     },
+    placeholderColor: theme.color.gray[600],
+    textFocused: theme.color.gray[900],
+    textBlurr: theme.color['black-static'],
```
stackchain:
```suggestion
```

apps/wallet-mobile/src/features/RampOnOff/common/AmountCard/AmountCard.tsx
```diff
@@ -193,6 +190,9 @@ const getStyles = (props: {theme: Theme}) => {
...
   })
   return styles
```
stackchain: @SorinC6 we can't touch the `styles` created by the `StyleSheet`, if we add non-style properties, later when "babel/process/css" into a browser style it is doom to fail, so non-style properties need to be returned a part from the styles, follow my suggestion.
```suggestion
  })
  const colors = {
    placeholder: theme.color.gray[600],
    focused: theme.color.gray[900],
    blur: theme.color['black-static'],
  }
  return {styles, colors} as const
```

stackchain: an alternative would be to use the text style `placeholderText: { color: ...}` but the lint for unused styles is dumb.

apps/wallet-mobile/src/features/RampOnOff/common/AmountCard/AmountCard.tsx
```diff
@@ -50,9 +49,7 @@ export const AmountCard: React.FC<AmountCardProps> = ({
 
   const fallback = React.useCallback(() => <TokenIconPlaceholder />, [])
 
-  const {theme} = useTheme()
-
-  const styles = React.useMemo(() => getStyles({theme: theme}), [theme])
+  const styles = useStyles()
```
stackchain:
```suggestion
  const {styles, colors} = useStyles()
```

apps/wallet-mobile/src/features/RampOnOff/common/AmountCard/AmountCard.tsx
```diff
@@ -68,10 +65,10 @@ export const AmountCard: React.FC<AmountCardProps> = ({
               autoComplete="off"
               value={value}
               placeholder="0"
-              placeholderTextColor={theme.color.gray[600]}
+              placeholderTextColor={styles.placeholderColor}
               onChangeText={onChange}
               allowFontScaling
-              selectionColor={isFocused ? theme.color.gray[900] : theme.color['black-static']}
+              selectionColor={isFocused ? styles.textFocused : styles.textBlurr}
```
stackchain:
```suggestion
              selectionColor={isFocused ? colors.focused : colors.blurr}
```

apps/wallet-mobile/src/features/RampOnOff/common/AmountCard/AmountCard.tsx
```diff
@@ -68,10 +65,10 @@ export const AmountCard: React.FC<AmountCardProps> = ({
               autoComplete="off"
               value={value}
               placeholder="0"
-              placeholderTextColor={theme.color.gray[600]}
+              placeholderTextColor={styles.placeholderColor}
```
stackchain:
```suggestion
              placeholderTextColor={colors.placeholder}
```

apps/wallet-mobile/src/features/RampOnOff/common/ShowBuyBanner/BuyBannerBig.tsx
```diff
@@ -34,7 +32,7 @@ export const BuyBannerBig = () => {
         style={styles.gradient}
         start={{x: 1, y: 1}}
         end={{x: 1, y: 1}}
-        colors={theme.color.gradients['green']}
+        colors={styles.gradientColor as (string | number)[]}
```
stackchain: @SorinC6 we need to type the colors in the style with the StyleSheet color type, so we can avoid this casting.

apps/wallet-mobile/src/features/RampOnOff/common/ShowBuyBanner/BuyBannerBig.tsx
```diff
@@ -99,6 +97,7 @@ const getStyles = (props: {theme: Theme}) => {
     spaceButton: {
       paddingHorizontal: 16,
     },
+    gradientColor: theme.color.gradients['green'] as ViewStyle,
```
stackchain: to avoid this as well, which is incorrect. {'gradientColor': number} is not a StyleSheet, that's why its need to strip off only colors, or fontNames, if we are passing it down in a non-styled way.

apps/wallet-mobile/src/features/RampOnOff/common/ShowBuyBanner/BuyBannerSmall.tsx
```diff
@@ -29,7 +26,7 @@ export const BuyBannerSmall = ({onClose}: SmallBannerProps) => {
         style={styles.gradient}
         start={{x: 1, y: 1}}
         end={{x: 0, y: 0}}
-        colors={theme.color.gradients['blue-green']}
+        colors={styles.gradientColor as (string | number)[]}
```
michalszorad: Ideally we would like to avoid type casting here and other places

apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
```diff
@@ -144,78 +145,88 @@ const useStrings = () => {
...
-    fontSize: 12,
-  },
-  grayText: {
-    color: '#6B7384',
-  },
-})
+const useStyles = () => {
+  const {theme} = useTheme()
+  const {color} = theme
+  const styles = StyleSheet.create({
```
michalszorad, line 151: Can we expect any significant performance degradation by creating the styles on every re-render compared to having the styles created only once?

packages/theme/src/Typography.stories.tsx
```diff
@@ -23,5 +27,7 @@ const Typography = () => {
 }
 
 const Row = ({children}: {children: React.ReactNode}) => {
-  return <View style={{alignItems: 'center', flexDirection: 'row'}}>{children}</View>
+  return (
+    <View style={{alignItems: 'center', flexDirection: 'row'}}>{children}</View>
```
michalszorad, line 31: Has there been any new lint config change, or why is this change here?

apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
SorinC6: Hi Michael, don't think that there will be issues because is a hook and react will automatically handle the change on dependency change 🤔

packages/theme/src/Typography.stories.tsx
SorinC6: @MichalSzorad just tried this and when I save the component is automatically formats like this... let me check my settings locally as well 🤔

apps/wallet-mobile/src/features/Swap/common/AmountCard/AmountCard.tsx
banklesss: Its possible to use variables there:

```
const styles = StyleSheet.create({
  container: {
    ...
  },
  foo: bar => ({
    backgroundColor: bar,  
  })
})
```

`<FancyComponet style={styles.foo('fancy-color')}  />`

jorbuedo: I wish I shared such an optimism on hooks having good performance. They tend to run way too many times.
In this case, `useStyles` uses `useTheme` inside, which is a context hook. But `useStyles` is not limited in any way on how many times it runs. `StyleSheet.create` will be called on every render.

I don't think we're tracking performance, or even that this in particular will be noticeable. But definitely it'll be more costly than the previous way.
I assume moving `theme` from being static to being inside hooks is because we want them reactive, to change to dark mode on the fly or something?

`StyleSheet.create` could be done inside a memo to reduce rerenders, but that also has some extra setup and memory cost. Hard to tell what's better without profiling it.

apps/wallet-mobile/src/features/Swap/common/ButtonGroup/ButtonGroup.tsx
```diff
@@ -28,26 +29,32 @@ export const ButtonGroup: React.FC<ButtonGroupProps> = ({labels, onSelect, selec
...
+    },
+    button: {
+      paddingHorizontal: 8,
+      paddingVertical: 8,
+      borderRadius: 8,
+    },
+    selected: {
+      backgroundColor: color.gray[200],
+    },
+    label: {
```
stackchain, line 50: @SorinC6 should we start applying the typography as well? I think this would be `body-1-medium`, also in the typography the weight applied now has only the font, `Rubik-Medium` and `Rubik` (normal and strong), I'd say we should add to these properties the `fontWeight` as well, cuz eventually we can switch from away `Rubik`'s limitation

packages/theme/src/ThemeProvider.tsx
```diff
@@ -14,12 +19,18 @@ export const ThemeProvider = ({children}: {children: React.ReactNode}) => {
 
   const selectColorScheme = useSaveColorScheme()
   const colorScheme = savedColorScheme ?? systemColorScheme
-  const theme = themes[isColorScheme(colorScheme) ? colorScheme : systemColorScheme]
+  const theme =
+    themes[isColorScheme(colorScheme) ? colorScheme : systemColorScheme]
 
-  return <ThemeContext.Provider value={{theme, colorScheme, selectColorScheme}}>{children}</ThemeContext.Provider>
+  return (
+    <ThemeContext.Provider value={{theme, colorScheme, selectColorScheme}}>
```
stackchain: @SorinC6 I'd say we should add 2 more attributes, `isLight` `isDark`, there are a lot of references to it, by returning directly from the context, will spare some logic in the components.

packages/theme/src/types.ts
```diff
@@ -0,0 +1,121 @@
+import {TextStyle} from 'react-native'
+
+export type Theme = {
+  color: Palette
+  typography: Typography
+  spacing: Spacing
+}
+
+export type HexColor = `#${string}`
```
stackchain, line 9: replace to `ColorValue` from `react-native`

packages/theme/src/types.ts
```diff
@@ -0,0 +1,121 @@
...
+  'yellow': {
+    500: HexColor
+    100: HexColor
+  }
+  'gradients': {
+    'blue-green': Gradient
+    'green': Gradient
+    'blue': Gradient
+  }
+  'overlay-extension': {hex: HexColor; opacity: number}
```
stackchain, line 72: does it require to be in `hex` ?

packages/theme/src/Typography.stories.tsx
stackchain: @MichalSzorad `/packages/**` are using a slightly modified lint/prettier config.

apps/wallet-mobile/src/features/Swap/common/LiquidityPool/LiquidityPool.tsx
```diff
@@ -25,20 +28,26 @@ export const LiquidityPool = ({
...
+const useStyles = () => {
+  const {theme} = useTheme()
+  const {color} = theme
+  const styles = StyleSheet.create({
+    liquidityPoolLink: {
+      alignItems: 'center',
+      justifyContent: 'center',
+    },
+    liquidityPoolText: {
+      color: color.primary[500],
```
stackchain, line 40: let's later check all `Text`s components, it should be matching figma, therefore any `Text` must have a typography attached to it, not a blocker to merge it.

apps/wallet-mobile/src/features/Swap/common/ButtonGroup/ButtonGroup.tsx
SorinC6: yes, I will add it but adding it now will increase the PR a lot, I think we can add this in the next PR that follows this `add dark/light/typography theme support for ramp on/of feature`

packages/theme/src/types.ts
SorinC6: hmnm, let me double-check this later as I think we can remove some of it.. was not sure that we will use it

apps/wallet-mobile/src/features/Settings/ApplicationSettings/ApplicationSettingsScreen.tsx
```diff
@@ -91,6 +94,15 @@ export const ApplicationSettingsScreen = () => {
             label={strings.analytics}
             onNavigate={navigateTo.analytics}
           />
+
+          {!isProduction() && (
+            <SettingsItem
+              icon={<Icon.EyeOff {...iconProps} />} // TODO
```
michalszorad, line 100: Is the todo still relevant?

apps/wallet-mobile/src/features/Settings/ApplicationSettings/ApplicationSettingsScreen.tsx
```diff
@@ -91,6 +94,15 @@ export const ApplicationSettingsScreen = () => {
...
+              label={`${capitalize(colorScheme)} Theme`} // TODO
```
michalszorad, line 101: This may need a translation

apps/wallet-mobile/src/features/Settings/ApplicationSettings/ApplicationSettingsScreen.tsx
SorinC6: @MichalSzorad  yes! we don't have yet the design for light/dark theme switcher, I have just added something using that flag until the design is done

stackchain (approved):

stackchain:
it's looking solid, bringing in

- [ ] update types to use ColorValue
- [ ] start updating Text to correlate to its design typography
- [ ] check how hex colors in gradients are define / used

PR #2986, opened 2024-01-11
fix: Show loading on multiple Send attempts

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-1082


SorinC6 (approved):

banklesss (approved):

apps/wallet-mobile/src/features/Send/useCases/ListAmountsToSend/ListAmountsToSendScreen.tsx
```diff
@@ -81,6 +82,8 @@ export const ListAmountsToSendScreen = () => {
   }
   const onNext = () => {
     track.sendSelectAssetSelected(assetsToSendProperties({tokens, amounts}))
+    // Trigger suspense on refetch
+    queryClient.resetQueries([wallet.id, 'send-tx'])
```
stackchain: hey @jorbuedo I think the issue here is that `useSendTx`, which is a wrong name, cuz it just create the unsigned transaction and doesn't actually send it, is using a query, cuz the create unsigned tx is supposed to be an idempotent call, which is "correct", yet it mutates data on the wallet balance, (if it was a web2 synchronous operations for example). I'd say the query should be replaced to a mutation, cuz it does cause side effects, and shoudln't be considered idempotent. (The name of the hook should also match with the operation I think it worth a refactor)

stackchain: along with the query key.

jorbuedo: Makes sense. I didn't know creating an unsigned tx was changing balance on the wallet.
I just went through the rabbit hole of why react-query only did suspense the first time and fixed that.

I'll try changing it from query to a mutation and see if that keeps the suspense and doesn't need a reset.

stackchain: it won't directly, not for now, but since we will be chaining transactions, or going for optimistic updates, then it will change the utxo state.

stackchain: it's legacy though, no worries.

jorbuedo: @stackchain looking into it, it doesn't seem like react-query mutations support suspense, only errorBoundry.
I would have to manually show the fallback spinner with a classic isLoading, so no boundary on the parent.
Is this preferable or should we keep the current solution on the PR?

apps/wallet-mobile/src/features/Settings/ManageCollateral/ManageCollateralScreen.tsx
```diff
@@ -49,15 +49,11 @@ export const ManageCollateralScreen = () => {
     tokenSelectedChanged,
     yoroiUnsignedTxChanged,
   } = useSend()
-  const {refetch: createUnsignedTx, isFetching: isLoadingTx} = useSendTx(
```
michalszorad, line 52: There will be some conflicts https://github.com/Emurgo/yoroi/pull/2989 😬

michalszorad (approved):

apps/wallet-mobile/src/features/Settings/ManageCollateral/ManageCollateralScreen.tsx
jorbuedo: Mm it's the same file but not same lines, maybe it merges it fine

stackchain (approved):

PR #2989, opened 2024-01-12
feat(collateral): Swap collateral text and flow changes

michalszorad:
Resolves [YOMO-1023](https://emurgo.atlassian.net/browse/YOMO-1023)

- [x] general alert for when  Collateral is NOT found, NOT valid, or NOT synchronized
- [x] settings icon
- [x] changed text for pending tx

[YOMO-1023]: https://emurgo.atlassian.net/browse/YOMO-1023?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/wallet-mobile/src/features/Swap/common/navigation.ts
```diff
@@ -29,9 +29,16 @@ export const useNavigateTo = () => {
                   state: {
                     routes: [
                       {
-                        name: 'swap-start-swap',
+                        name: 'history',
```
stackchain, line 32: 👀 working previously on navigator fallback maybe?

michalszorad: It redirected the person to tx history (which was intended behaviour), however the function name was `navigateTo.swapOpenOrders`. I've fixed the redirect inside this fn, so it does what the fn name says and also added `  onPress={() => walletNavigate.navigateToTxHistory()}` to `ShowSubmittedTxScreen` so it keeps the original functionality

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/OpenOrders.tsx
```diff
@@ -176,12 +188,20 @@ export const OpenOrders = () => {
     )
   }
 
+  const handleCollateralError = () => {
+    if (modalOpenRef.current) {
+      closeModal()
```
michalszorad, line 193: Need to use ref here, because when `handleCollateralError` is called asynchronously, it can happen the `isModalOpen` is still set to false

stackchain (approved):
lgtm

PR #2995, opened 2024-01-15
chore: update yarn.lock

banklesss:


stackchain:
monorepo local packages don't require yarn updates

PR #3000, opened 2024-01-15
feature(cdn): Request the raw image and replace the image component

jorbuedo:
realted: 
- https://emurgo.atlassian.net/browse/YOMO-1095
- https://emurgo.atlassian.net/browse/YOMO-1092
- https://emurgo.atlassian.net/browse/YOMO-1096

Notes:
- ~Uses `expo-image` which adds caching (memory and/or disk)~
- Asks for raw images (option for base64 still available, but unused)
- Asks for gif on ios
- Removes zoom x3 for now, image is too big and crashes expo-image. @pedromtcosta Backend is not limiting the max size to the original image size
- ~Loading with Image instead of react-query makes it harder to use the animated skeleton, it shows the placeholder image instead while loading. Could try with loading callbacks and component state to add an absolutely positioned skeleton~ **Skeletons showing fine now**
- Cache invalidation ~doesn't work~  **works**

### DIFF
--- a/apps/wallet-mobile/package.json
+++ b/apps/wallet-mobile/package.json
@@ -144,6 +144,7 @@
     "expo-barcode-scanner": "~12.3.2",
     "expo-camera": "^13.2.1",
     "expo-device": "^5.4.0",
+    "expo-image": "~1.0.2",
     "expo-status-bar": "~1.4.4",
     "jsc-android": "241213.1.0",
     "lodash": "^4.17.21",
--- a/apps/wallet-mobile/package.json
+++ b/apps/wallet-mobile/package.json
@@ -144,7 +144,6 @@
     "expo-barcode-scanner": "~12.3.2",
     "expo-camera": "^13.2.1",
     "expo-device": "^5.4.0",
-    "expo-image": "~1.0.2",
     "expo-status-bar": "~1.4.4",
     "jsc-android": "241213.1.0",
     "lodash": "^4.17.21",
--- a/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
+++ b/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
@@ -62,7 +62,7 @@ export const NftPreview = ({
         </SkeletonPlaceholder>
       ) : (
         <Image
-          cachePolicy={shouldShowPlaceholder ? 'none' : cachePolicy}
+          cachePolicy={failed ? 'none' : cachePolicy}
           source={shouldShowPlaceholder ? placeholder : {uri, headers}}
           placeholder={placeholder}
           placeholderContentFit="contain"
--- a/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
+++ b/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
@@ -1,7 +1,8 @@
 import {isString} from '@yoroi/common'
 import {Balance} from '@yoroi/types'
-import React from 'react'
-import {Image, ImageStyle, StyleProp, View} from 'react-native'
+import {Image, ImageStyle} from 'expo-image'
+import React, {useEffect, useState} from 'react'
+import {View} from 'react-native'
 import SkeletonPlaceholder from 'react-native-skeleton-placeholder'
 
 import placeholder from '../../assets/img/nft-placeholder.png'
@@ -12,38 +13,46 @@ import {useNativeAssetImage} from '../../yoroi-wallets/hooks'
 type NftPreviewProps = {
   nft: Balance.TokenInfo
   showPlaceholder?: boolean
-  style?: StyleProp<ImageStyle>
+  style?: ImageStyle
   height: number
   width: number
-  resizeMode?: 'cover' | 'contain'
+  contentFit?: 'cover' | 'contain'
   blurRadius?: number
-  zoom?: 1 | 2 | 3
+  zoom?: number
+  cachePolicy?: 'none' | 'memory' | 'disk' | 'memory-disk'
 }
 
 export const NftPreview = ({
   nft,
   showPlaceholder,
-  style,
+  style = {},
   height,
   width,
-  resizeMode = 'cover',
+  contentFit = 'cover',
   blurRadius,
   zoom = 1,
+  cachePolicy,
 }: NftPreviewProps) => {
   const wallet = useSelectedWallet()
   const [policy, name] = nft.id.split('.')
-  const {uri, isLoading, isError} = useNativeAssetImage({
+  const {uri, headers, isLoading, isError} = useNativeAssetImage({
     networkId: wallet.networkId,
     policy,
     name,
     width: width * zoom,
     height: height * zoom,
     kind: 'metadata',
-    resizeMode,
+    contentFit,
     mediaType: getNftMainImageMediaType(nft),
   })
 
-  const shouldShowPlaceholder = !isString(uri) || showPlaceholder || isError
+  const [failed, setFailed] = useState(false)
+
+  useEffect(() => {
+    setFailed(false)
+  }, [uri])
+
+  const shouldShowPlaceholder = !isString(uri) || showPlaceholder || isError || failed
 
   return (
     <View style={{width, height, overflow: 'hidden'}}>
@@ -53,9 +62,13 @@ export const NftPreview = ({
         </SkeletonPlaceholder>
       ) : (
         <Image
-          source={shouldShowPlaceholder ? placeholder : {uri}}
+          cachePolicy={shouldShowPlaceholder ? 'none' : cachePolicy}
+          source={shouldShowPlaceholder ? placeholder : {uri, headers}}
+          placeholder={placeholder}
+          placeholderContentFit="contain"
+          onError={() => setFailed(true)}
           style={[style, {width, height}]}
-          resizeMode={resizeMode}
+          contentFit={contentFit}
           blurRadius={blurRadius}
         />
       )}
--- a/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
+++ b/apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
@@ -1,8 +1,7 @@
 import {isString} from '@yoroi/common'
 import {Balance} from '@yoroi/types'
-import {Image, ImageStyle} from 'expo-image'
-import React, {useEffect, useState} from 'react'
-import {View} from 'react-native'
+import React from 'react'
+import {Image, ImageStyle, View} from 'react-native'
 import SkeletonPlaceholder from 'react-native-skeleton-placeholder'
 
 import placeholder from '../../assets/img/nft-placeholder.png'
@@ -19,7 +18,6 @@ type NftPreviewProps = {
   contentFit?: 'cover' | 'contain'
   blurRadius?: number
   zoom?: number
-  cachePolicy?: 'none' | 'memory' | 'disk' | 'memory-disk'
 }
 
 export const NftPreview = ({
@@ -31,11 +29,10 @@ export const NftPreview = ({
   contentFit = 'cover',
   blurRadius,
   zoom = 1,
-  cachePolicy,
 }: NftPreviewProps) => {
   const wallet = useSelectedWallet()
   const [policy, name] = nft.id.split('.')
-  const {uri, headers, isLoading, isError} = useNativeAssetImage({
+  const {uri, headers, isLoading, isError, onError, onLoad} = useNativeAssetImage({
     networkId: wallet.networkId,
     policy,
     name,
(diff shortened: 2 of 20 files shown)


apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -985,34 +986,51 @@ export const useNativeAssetImage = ({
...
   kind = 'metadata',
+  responseType = 'binary',
 }: NativeAssetImageRequest) => {
   const network = networkId === 300 ? 'preprod' : 'mainnet'
   const pWidth = PixelRatio.getPixelSizeForLayoutSize(Number(width))
   const pHeight = PixelRatio.getPixelSizeForLayoutSize(Number(height))
-  const isMediaTypeSupported = supportedTypes.includes(mediaType.toLocaleLowerCase())
+  const lcMediaType = mediaType.toLocaleLowerCase()
+  const isMediaTypeSupported = supportedTypes.includes(lcMediaType)
+  const needsGif = lcMediaType === 'image/gif' && Platform.OS === 'ios'
```
michalszorad, line 1006: Why don't we support gif for android?

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -985,34 +986,51 @@ export const useNativeAssetImage = ({
...
+      'X-Encoded-Mimetype': mimeType,
+    }),
+    [mimeType, responseType],
+  )
+
   const query = useQuery({
+    enabled: isMediaTypeSupported,
     staleTime: Infinity,
-    queryKey: ['native-asset-img', policy, name, `${width}x${height}`, resizeMode],
+    queryKey: ['native-asset-img', policy, name, `${pWidth}x${pHeight}`, contentFit],
```
michalszorad: `responseType` may need to be here as well

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -985,34 +986,51 @@ export const useNativeAssetImage = ({
...
     queryFn: async () => {
-      const response = await fetch(
-        `https://${network}.cardano-nativeassets-prod.emurgornd.com/${policy}/${name}?width=${pWidth}&height=${pHeight}&kind=${kind}&fit=${resizeMode}`,
-      )
+      const requestUrl = `https://${network}.cardano-nativeassets-prod.emurgornd.com/${policy}/${name}?width=${pWidth}&height=${pHeight}&kind=${kind}&fit=${contentFit}`
```
michalszorad: Is it worth having the URL somewhere in a config file?

apps/wallet-mobile/package.json
```diff
@@ -144,6 +144,7 @@
     "expo-barcode-scanner": "~12.3.2",
     "expo-camera": "^13.2.1",
     "expo-device": "^5.4.0",
+    "expo-image": "~1.0.2",
```
michalszorad: The latest version is 1.5.1 why are we using older one? Is it a current RN version limitation? Does the older version support all images (eg SVGs) as the latest one?

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: Android uses webp, which is better
It's ios the buggy one that needs gif

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: Technically true, but we don't mix responseType, so it doesn't really change.
We're either using base64 or raw, not both

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: I don't know, it's composed with a lot of variables. 
I guess `cardano-nativeassets-prod.emurgornd.com` could be on another variable, but sounds more confusing to me

apps/wallet-mobile/package.json
jorbuedo: The install command `npx expo install expo-image` picks the version depending on the project, so I guess it's limited by some other dependency we have

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
michalszorad: Does it make sense in that case to remove the `responseType` param?

jorbuedo: If we remove the param then all of the support for base64 needs to be removed as well. Which very well could happen, it just needs to be decided first. Once the raw has been fully tested we can do that. There's still multiple tickets with cdn improvements.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
stackchain: @jorbuedo check with @crypto-titan, maybe we have a CNAME already for it, I think emurgornd, usually is closed through our VPN only, so it needs to go over Cloudfare.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
stackchain: remember that `Infinity` here means, user session, since we are not persisting the react-query cache.

stackchain: yet, not sure about `expo-image`

michalszorad: While we do have support for both, I'd recommend adding the `responseType` as a dependency, just not to introduce issues when we actually use both response types.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
michalszorad: I thought just about the domain name broken down by network type as we have here `packages/openswap/src/config.ts`

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: I added `expo-image` to improve on the base images, but I had it working without. Just with the same cache issue, because the image just doesn't reload for the same url.

jorbuedo: > remember that Infinity here means, user session, since we are not persisting the react-query cache.

My issue is not with the react-query cache. 
It's with the Image component. Neither Image from react-native nor Image from expo-image redownload an image with the same url

apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
```diff
@@ -12,38 +13,46 @@ import {useNativeAssetImage} from '../../yoroi-wallets/hooks'
...
+    contentFit,
     mediaType: getNftMainImageMediaType(nft),
   })
 
-  const shouldShowPlaceholder = !isString(uri) || showPlaceholder || isError
+  const [failed, setFailed] = useState(false)
+
+  useEffect(() => {
+    setFailed(false)
+  }, [uri])
```
banklesss: Could `onSuccess` or `onSettle` from `useNativeAssetImage` be useful?

stackchain:
waiting infra.

apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
banklesss: well callbacks are deprecated in new versions so i guess is the best option

stackchain: I think it's ok cuz react-query is dropping on the next version all callbacks.

jorbuedo: `useNativeAssetImage` doesn't have those. Would have to create them and pass them to useQuery. But since newer versions of useQuery don't support that, seems like adding something that will need to be removed whenever react-query gets updated

banklesss: Yeah that was my initial idea but i remembered this

https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
stackchain: we need `ETags` on next implementation

jorbuedo:
Fixed issues. From my side it's fine. Waiting for infra about cdn url change and routing through cloudflare.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -985,34 +994,83 @@ export const useNativeAssetImage = ({
...
   })
 
+  const timerRef = React.useRef<any>(0)
+  React.useEffect(() => () => clearTimeout(timerRef.current), [])
+
+  const onError = useCallback(() => {
+    setError(true)
+    const count = queryClient.getQueryState(queryKey)?.dataUpdateCount
+    if (count && count < 10) {
+      timerRef.current = setTimeout(query.refetch, count * 300)
```
michalszorad, line 1057: I wonder if `refetchInterval` param coult be used instead. Eg 

```ts
useQuery<string, Error>({
// ...
refetchInterval: (data, query) => {
      const count = query.state.dataUpdateCount
      if (query.state.error &&  count < 10) {
        return count * 300
      }
      return false;
    },
});
```

jorbuedo: Mmm... could be, I only want it refetching when there's an error though. And it is possible to have an image, no errors, and then invalidate it and get an error.
Not sure how useQuery  would react to refetchInterval being modified. Or if it calls it again after having returned false before...

jorbuedo: Also, `query.state.error` wouldn't work, because the query didn't fail. The query generates the url successfully. 
It's the `<Image/>` component the one doing the image fetching and possibly failing. That's why I have to use the onError callback to see if it failed.
On the base64 version, queryFn can fail and react-query will automatically retry a few times before actually showing the error. That's what I'm imitating for the binary version with the timeouts.

michalszorad (approved):

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -985,34 +994,83 @@ export const useNativeAssetImage = ({
...
+        throw new Error(`NativeAsset CDN: Processing - ${requestUrl}`)
       }
-      return `data:image/webp;base64,${await response.text()}`
+      return `data:${mimeType};base64,${await response.text()}`
     },
-    enabled: isMediaTypeSupported,
   })
 
+  const timerRef = React.useRef<any>(0)
+  React.useEffect(() => () => clearTimeout(timerRef.current), [])
```
stackchain: it supposed to work.
```suggestion
  const timerRef = React.useRef<number | null>(null)
  React.useEffect(() => () => clearTimeout(timerRef.current), [])
```

jorbuedo: It complains because `node.js` types are interfering with react-native's https://dev.to/retyui/react-native-fixing-typescript-issue-type-timeout-is-not-assignable-to-type-number-10hg

`number|null` doesn't work. I can try to find a fully compliant solution if `any` is too bad

jorbuedo: This one seems to silence the error
```
  const timerRef = React.useRef<ReturnType<typeof setTimeout>>()
```

stackchain: 👀  https://github.com/facebook/react-native/blob/cd6755f2720dd6787545834ffd83905fd2d7517b/packages/react-native/types/modules/globals.d.ts#L27

stackchain: I'm guessing is the `null` issue.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
jorbuedo: Alright, we have `processed-media.yoroiwallet.com` now

jorbuedo: There's no `config` file on the app to put this on, and the idea is to refactor the whole thing once is finished.
So I'd say this is resolved for now.

PR #3003, opened 2024-01-16
fix(swap): removed price impact on limit orders

SorinC6:
task: https://emurgo.atlassian.net/browse/YOMO-1085

### DIFF
--- a/apps/wallet-mobile/src/components/AmountItem/AmountItem.tsx
+++ b/apps/wallet-mobile/src/components/AmountItem/AmountItem.tsx
@@ -1,4 +1,5 @@
 import {Balance} from '@yoroi/types'
+import {SwapOrderType} from '@yoroi/types/lib/swap/order'
 import * as React from 'react'
 import {StyleSheet, View, ViewProps} from 'react-native'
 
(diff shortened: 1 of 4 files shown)


apps/wallet-mobile/src/components/AmountItem/AmountItem.tsx
```diff
@@ -43,6 +46,7 @@ export const AmountItem = ({
   const formattedQuantity = Quantities.format(quantity, tokenInfo.decimals ?? 0)
   const showSwapDetails = !isPrimary && variant === 'swap'
   const priceImpactRiskTheme = usePriceImpactRiskTheme(priceImpactRisk ?? 'none')
+  const priceImpactRiskTextColor = orderType === 'market' ? priceImpactRiskTheme.text : styles.text.color
```
stackchain, line 49: hey @SorinC6 just to confirm, so it won't hide, it will just display without emphasis in the color?

SorinC6: I had a task before for when is a limit order to not display the Price Impact on the Confirm Transaction(summary) screen. As I see it if there is a limit order everything will be as usual and not show the price impact colors and warnings 🤔

PR #3014, opened 2024-01-17
fix(nft): metadata display

banklesss:


apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
```diff
@@ -40,7 +39,12 @@ export const parseNFT = (
     return null
   }
 
-  const metadata = nftAsset.metadata?.[policyId]?.[nameHex]
+  const policyId = toPolicyId(tokenId)
+  const metadataPolicyId = nftAsset.metadata?.[policyId]
+
+  const nameHex = toAssetNameHex(tokenId)
+  const metadata = metadataPolicyId?.[nameHex] ?? metadataPolicyId?.[toDisplayAssetName(tokenId)]
```
michalszorad, line 46: Why is the metadata not accessible by the nameHex key?

banklesss: example for adadomains

```
"fc411f546d01e88a822200243769bbc1e1fbdde8fa0f6c5179934edb.6a6176696275656e6f": [
    {
      "key": "721",
      "metadata": {
        "fc411f546d01e88a822200243769bbc1e1fbdde8fa0f6c5179934edb": {
          "javibueno": {
            "image": "ipfs://QmNbZQLBoqNdsmmVJ42Yip5wcbRiqTbwgFAi3ChWQX74nm",
            "name": "javibueno.ada",
            "project": "adadomains"
          }
        }
      }
    }
  ]
```

banklesss: let me know if you see anything wrong please

stackchain: @MichalSzorad name key (`metadata.policyId.name`), is based on version, if the version is 1, is utf-8 (which eventually fails, and goes against the cip26 specs to be ascii only), on v2 it was changed to be in hex, so if it is v1, (not present version is equal v1) the key-name is utf8, v2 -> hex (raw bytes).

stackchain (approved):

PR #3018, opened 2024-01-17
fix(cdn): Remove base64 support

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-1119

Backend is gonna remove base64 support. Just having the accept header will make it break. So this change needs to go in along with backend update @pedromtcosta

### DIFF
--- a/apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
+++ b/apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
@@ -1012,7 +1012,7 @@ export const useNativeAssetImage = ({
   const mimeType = needsGif ? 'image/gif' : 'image/webp'
   const headers = useMemo(
     () => ({
-      'X-Encoded-Mimetype': mimeType,
+      Accept: mimeType,
     }),
     [mimeType],
   )
--- a/apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
+++ b/apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
@@ -992,7 +992,6 @@ type NativeAssetImageRequest = {
   mediaType?: string
   contentFit?: 'contain' | 'cover' | 'fill' | 'inside' | 'outside'
   kind?: 'logo' | 'metadata'
-  responseType?: 'binary' | 'base64'
 }
 export const useNativeAssetImage = ({
   networkId,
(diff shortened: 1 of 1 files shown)


apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -1014,42 +1012,28 @@ export const useNativeAssetImage = ({
   const mimeType = needsGif ? 'image/gif' : 'image/webp'
   const headers = useMemo(
     () => ({
-      Accept: responseType === 'binary' ? mimeType : 'text/plain',
       'X-Encoded-Mimetype': mimeType,
```
stackchain: in this case, should it replace the `Accept` to use directly the `mimeType` expected, and drop the custom header?

jorbuedo: IIii1iI don't know, not how @pedromtcosta did it. We can change it I guess? Right now it should be failing though, since he already deployed it like this.

banklesss:
@jorbuedo 

That random error from PR check will be fixed in this PR https://github.com/Emurgo/yoroi/pull/3019

stackchain (approved):

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -1014,42 +1012,28 @@ export const useNativeAssetImage = ({
...
 
-  const queryKey = ['native-asset-img', policy, name, responseType, `${width}x${height}`, contentFit]
+  const queryKey = ['native-asset-img', policy, name, `${width}x${height}`, contentFit]
 
   const query = useQuery({
     enabled: isMediaTypeSupported,
     staleTime: Infinity,
     queryKey,
-    queryFn: async (context) => {
+    queryFn: (context) => {
```
michalszorad, line 1030: Does it still need to be in `useQuery` now that it only returns a string?

stackchain: it is used for the caching only, later it will be replaced by the portfolio manager.

jorbuedo: Makes it easier to do the invalidation, be disabled for unsupported, etc.

PR #3019, opened 2024-01-18
fix(resolver): cns error handling

banklesss:


michalszorad (approved):

stackchain:
my guess on the pipeline error is the 2 renders of react when in strict mode. I'd say to update the test to check if it was called only.

packages/resolver/src/adapters/api.test.ts
```diff
@@ -16,10 +16,12 @@ const mockError = new Error('Test Error')
 describe('resolverApiMaker', () => {
   const domain = 'example.domain'
 
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
   describe('getCardanoAddresses', () => {
     describe('strategy "all"', () => {
-      beforeEach(jest.clearAllMocks)
```
stackchain: 🤦🏻‍♂️

PR #3026, opened 2024-01-19
fix(resolver): update unstoppable supported tlds and manage errors

banklesss:


jorbuedo (commented):
I don't love all the `data` used here.
If the API is returning an error message in the error response, why not use that as `message`? The `statusText` is superfluous anyway when you know the status code.

banklesss:
> I don't love all the `data` used here.
> 
> If the API is returning an error message in the error response, why not use that as `message`? The `statusText` is superfluous anyway when you know the status code.

Message field in this case is undefined. The data is coming in the field data.

packages/common/src/api/fetchData.ts
```diff
@@ -90,19 +90,25 @@ export const fetchData: FetchData = <T, D = any>(
       if (error.response) {
         const status = error.response.status
         const message = error.response.statusText
+        const data = error.response.data
```
jorbuedo: I know it means adding logic in the fetcher that could potentially for some endpoint not work like that. But I'd say it's pretty safe for most cases to have:

```ts
        const data = error.response.data as {message?: string; error?: string}
        const message =
          data?.message ?? data?.error ?? error.response.statusText
   ```

Comment (jorbuedo) [reply packages/common/src/api/fetchData.ts]: > Message field in this case is undefined. The data is coming in the field data.

`data` is just where axios puts the response body. Basically the server is responding with:
```
{message: "Unsupported TLD"}
```

Which is a common format. Errors usually come with  `message`, `error`, `details`, `help`... So having the fetcher function try finding a message in the body and using that as `message`, seems better to me than just passing `data` along and having to deal with a `data` field (which was made up by axios) everywhere.

stackchain: Yap, I was bugging about the same thing, still there are some errors that might contains a response data detailing the problem, therefore if we can't merge (message + data), @banklesss I'd like at least to have the `data` renamed to `responseData`. So at least the semantic is better, and the client has a clear way to inspect the response body in case of error. Like 400, Bad Request (missing fields, or invalid fields will come in the body)

banklesss:
Created this task to also change the field `data` from the `fetchData` success response after the current release.

https://emurgo.atlassian.net/browse/YOMO-1124

stackchain:
~~hey @banklesss I think you can apply it now, this only affects `.com` from UD, will only include it if another change is requested by Product after the UAT.~~
I mix up the comments

stackchain:
> Created this task to also change the field `data` from the `fetchData` success response after the current release.
> 
> https://emurgo.atlassian.net/browse/YOMO-1124

> Created this task to also change the field `data` from the `fetchData` success response after the current release.
> 
> https://emurgo.atlassian.net/browse/YOMO-1124

I think this is returning the axios object directly, if it is no need to refactor for now.

stackchain:
> > Created this task to also change the field `data` from the `fetchData` success response after the current release.
> > https://emurgo.atlassian.net/browse/YOMO-1124
> 
> > Created this task to also change the field `data` from the `fetchData` success response after the current release.
> > https://emurgo.atlassian.net/browse/YOMO-1124
> 
> I think this is returning the axios object directly, if it is no need to refactor for now.

yap, it's remapping, keeping the task.

stackchain (approved):

PR #3034, opened 2024-01-23
fix(resolver): show error when the domain has not a cardano address

banklesss:
Related to [YOMO-1122](https://emurgo.atlassian.net/browse/YOMO-1122)

[YOMO-1122]: https://emurgo.atlassian.net/browse/YOMO-1122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/i18n/locales/en-US.json
+++ b/apps/wallet-mobile/src/i18n/locales/en-US.json
@@ -770,6 +770,7 @@
   "scan.linksErrorUnsupportedVersion.title": "Unsupported version",
   "scan.title": "Scan the QR code",
   "send.helper.addressError.invalid": "Please enter a valid receiver address, ADA Handle or domain",
+  "send.helper.addressError.invalidBlockchain": "You are trying to resolve a domain on a different network. Please double-check and try again",
   "send.helper.addressError.wrongNetwork": "Please enter an address for the current wallet network",
   "send.helper.resolver.resolvedAddress": "Resolved address",
   "send.helper.resolverError.domainNotFound": "Receiver address, ADA Handle or domain you entered doesn't exist. Please double-check it and try again",
--- a/apps/wallet-mobile/src/i18n/locales/en-US.json
+++ b/apps/wallet-mobile/src/i18n/locales/en-US.json
@@ -770,7 +770,7 @@
   "scan.linksErrorUnsupportedVersion.title": "Unsupported version",
   "scan.title": "Scan the QR code",
   "send.helper.addressError.invalid": "Please enter a valid receiver address, ADA Handle or domain",
-  "send.helper.addressError.invalidBlockchain": "You are trying to resolve a domain on a different network. Please double-check and try again",
+  "send.helper.addressError.wrongBlockchain": "You are trying to resolve a domain on a different blockchain. Please double-check and try again",
   "send.helper.addressError.wrongNetwork": "Please enter an address for the current wallet network",
   "send.helper.resolver.resolvedAddress": "Resolved address",
   "send.helper.resolverError.domainNotFound": "Receiver address, ADA Handle or domain you entered doesn't exist. Please double-check it and try again",
--- a/packages/resolver/src/adapters/unstoppable/api.ts
+++ b/packages/resolver/src/adapters/unstoppable/api.ts
@@ -60,8 +60,7 @@ export const unstoppableApiGetCryptoAddress = (
         if (hasOtherBlockchainAddress)
           throw new Resolver.Errors.WrongBlockchain()
 
-        if (hasNotAnyAddress)
-          throw new Resolver.Errors.NotFound()
+        if (hasNotAnyAddress) throw new Resolver.Errors.NotFound()
 
         throw new Resolver.Errors.InvalidResponse()
       }
--- a/packages/resolver/src/adapters/unstoppable/api.ts
+++ b/packages/resolver/src/adapters/unstoppable/api.ts
@@ -61,9 +61,8 @@ export const unstoppableApiGetCryptoAddress = (
         if (hasOtherBlockchainAddress)
           throw new Resolver.Errors.WrongBlockchain()
 
-        if (hasNotAnyAddress) {
+        if (hasNotAnyAddress)
           throw new Resolver.Errors.NotFound()
-        }
 
         throw new Resolver.Errors.InvalidResponse()
       }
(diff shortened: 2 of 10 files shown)


jorbuedo (approved):

packages/resolver/src/adapters/unstoppable/api.ts
```diff
@@ -42,13 +40,25 @@ export const unstoppableApiGetCryptoAddress = (
...
+        const safeParsedGeneralResponse =
+          UnstoppableApiGeneralResponseSchema.safeParse(response.value.data)
+
+        if (safeParsedAdaResponse.success) {
+          const result = response.value.data.records['crypto.ADA.address']
+          return result
+        }
+
+        if (safeParsedGeneralResponse.success) {
+          if (Object.keys(response.value.data.records).length > 0)
```
stackchain: Shouldn't check only if contains ADA? I had 3 different addresses on my UD in the past, I think it will respond always with all you got. It should look for ADA and ignore the rest IMHO.

packages/types/src/index.ts
```diff
@@ -59,6 +59,7 @@ import {
 } from './api/errors'
 import {ResolverNameServer} from './resolver/name-server'
 import {
+  ResolverErrorInvalidBlockchain,
```
stackchain: I think it'd be more precise WrongBlockchain, in case a TLD resolves only for other chain. For all others NotFound.

stackchain (commented):
⬆️

apps/wallet-mobile/src/i18n/locales/en-US.json
```diff
@@ -770,6 +770,7 @@
   "scan.linksErrorUnsupportedVersion.title": "Unsupported version",
   "scan.title": "Scan the QR code",
   "send.helper.addressError.invalid": "Please enter a valid receiver address, ADA Handle or domain",
+  "send.helper.addressError.invalidBlockchain": "You are trying to resolve a domain on a different network. Please double-check and try again",
```
stackchain: shouldn't it be something like "This domain resolves only for other blockchains", cuz it's mentioning only "network" here. Please ping Product about the copy.

stackchain: Let's also rename all to WrongBlockchain, to keep convention.

packages/resolver/src/adapters/unstoppable/api.ts
banklesss: Hi @stackchain !

The purpose of this piece of code is to capture the situation when the user does not have an address associated with cardano but does have one for other blockchains.

What I am doing is first checking if there are any records associated with cardano, if not, then I check that the number of records is greater than 0. If it is greater, the "wrong blockchain" error is thrown.

I have refactored the code to make it more readable.

Please comment if you have any suggestions

Thanks!

banklesss: I think it's an interesting step forward towards multichain!

packages/resolver/src/adapters/unstoppable/api.ts
```diff
@@ -42,13 +40,33 @@ export const unstoppableApiGetCryptoAddress = (
...
+        const hasOtherBlockchainAddress =
+          !hasCardanoAddress &&
+          safeParsedGeneralResponse.success &&
+          Object.keys(response.value.data.records).length > 0
+        const hasNotAnyAddress =
+          !hasOtherBlockchainAddress && safeParsedGeneralResponse.success
+
+        if (hasCardanoAddress) {
+          const result = response.value.data.records['crypto.ADA.address']
+          return result
```
stackchain: it was used before just to throw, now it is throwing earlier
```suggestion
          return response.value.data.records['crypto.ADA.address']
```

packages/resolver/src/adapters/unstoppable/api.ts
stackchain: yap, really good.

packages/types/src/index.ts
stackchain:
```suggestion
  ResolverErrorWrongBlockchain,
```

packages/types/src/index.ts
```diff
@@ -248,6 +249,7 @@ export namespace Resolver {
     export class NotFound extends ResolverErrorNotFound {}
     export class UnsupportedTld extends ResolverErrorUnsupportedTld {}
     export class Expired extends ResolverErrorUnsupportedTld {}
+    export class WrongBlockchain extends ResolverErrorInvalidBlockchain {}
```
stackchain:
```suggestion
    export class WrongBlockchain extends ResolverErrorWrongBlockchain {}
```

packages/types/src/resolver/errors.ts
```diff
@@ -3,3 +3,4 @@ export class ResolverErrorInvalidDomain extends Error {}
 export class ResolverErrorNotFound extends Error {}
 export class ResolverErrorUnsupportedTld extends Error {}
 export class ResolverErrorExpiredDomain extends Error {}
+export class ResolverErrorInvalidBlockchain extends Error {}
```
stackchain:
```suggestion
export class ResolverErrorWrongBlockchain extends Error {}
```

packages/resolver/src/adapters/unstoppable/api.ts
```diff
@@ -42,13 +40,32 @@ export const unstoppableApiGetCryptoAddress = (
...
+
+        if (hasOtherBlockchainAddress)
+          throw new Resolver.Errors.WrongBlockchain()
 
-        const result = parsedResponse.records['crypto.ADA.address']
-        if (!result) throw new Resolver.Errors.NotFound()
-        return result
+        if (hasNotAnyAddress) {
+          throw new Resolver.Errors.NotFound()
+        }
```
stackchain:
```suggestion
        if (hasNotAnyAddress)
          throw new Resolver.Errors.NotFound()
```

packages/resolver/src/adapters/unstoppable/api.ts
```diff
@@ -42,13 +40,31 @@ export const unstoppableApiGetCryptoAddress = (
...
+        const hasOtherBlockchainAddress =
+          !hasCardanoAddress &&
+          safeParsedGeneralResponse.success &&
+          Object.keys(response.value.data.records).length > 0
+        const hasNotAnyAddress =
+          !hasOtherBlockchainAddress && safeParsedGeneralResponse.success
+
+        if (hasCardanoAddress) {
+          return response.value.data.records['crypto.ADA.address']
+        }
```
stackchain:
```suggestion
        if (hasCardanoAddress)
          return response.value.data.records['crypto.ADA.address']
```

banklesss:
thanks @stackchain 🙏🏻

PR #3036, opened 2024-01-24
feature(workspace): relative path suggestion

banklesss:
Now VSC will suggest `../path/to/file`

insetad of `src/path/to/file`


stackchain:
🥇

stackchain (approved):

PR #3037, opened 2024-01-24
fix(resolver): wrong blockchain error

banklesss:
Related to [YOMO-1122](https://emurgo.atlassian.net/browse/YOMO-1122)

[YOMO-1122]: https://emurgo.atlassian.net/browse/YOMO-1122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/resolver/src/adapters/unstoppable/api.ts
```diff
@@ -104,9 +108,19 @@ const UnstoppableApiAdaResponseSchema = z.object({
...
+    blockchain: z.string(),
+  }),
   records: z.record(z.string(), z.string()),
 })
 
+const UnstoppableApiOtherBlockchainResponseSchema =
+  UnstoppableApiGeneralResponseSchema.refine(
+    ({records, meta}) =>
+      Object.keys(records).length > 0 ||
+      !['CARDANO', 'ADA'].includes(meta.blockchain),
```
stackchain: @banklesss I believe this is where the record of the resolver lives, i.e for Cardano domains, they are hosted on MATIC blockchain. Maybe the check is, if is not 404, but doesn't contain ...ADA... and there is a response 200, is for other blockchain.

stackchain:
```
curl -i -X GET \\n  'https://api.unstoppabledomains.com/resolve/domains/stackchain.blockchain' \\n  -H 'Authorization: Bearer ${ENV-TOKEN_UD}'
```

banklesss: i will take a look thanks

banklesss: yeah @stackchain i think we are refering to the same logic. I reflected that (if i am not wrong 😅)

please let me know if you have any suggestion

banklesss:
```
Object.keys(records).length > 0 ||  !['CARDANO', 'ADA'].includes(meta.blockchain),
```

if the records are not empty or "blockchain" is not empty and is not "ADA" or "CARDANO", the domain has assigned an address on another blockchain other than cardano.

stackchain: my point is `meta.blockchain` will never contain 'CARDANO' or 'ADA' cuz Cardano records are kept on MATIC network.

banklesss: ah! got it

banklesss: so @stackchain 

this response should be considered has "has not addresses in any blockchain" because doesn't have records right?

```
{
  meta: {
    domain: 'john.go',
    tokenId: '68339273160565029666262671917772397379555826087941337618091301383115593492574',
    namehash: '0x9716a86b68212d3124300dbeb22cbeb49f5131a8539bbdb747e3e4a2320c285e',
    blockchain: 'MATIC',
    networkId: 137,
    owner: '0x0c8ccc39309ddca7b803662231b780cf4131b6b1',
    resolver: '0xa9a6a3626993d487d2dbda3173cf58ca1a9d9e9f',
    registry: '0xa9a6a3626993d487d2dbda3173cf58ca1a9d9e9f',
    reverse: false,
    type: 'Uns'
  },
  records: {}
}
```

banklesss: if its right we can close the PR

banklesss: according to the official documentation `records` field will contain the linked addresses

stackchain: I dunno, cuz it looks like the `records` property was added later, cuz for `john.go` I'm guessing the `owner` is the resolved addressed, and the resolver is the contract responsible to hold the data, needs checking.
Then the test for the `.length > 0` can be dropped, and the rule for returning the `WrongBlockchain` is -> the response is 200, and there is no ..ADA.. in the records.

banklesss: i will take a look

banklesss: mixing this and updating the tlds makes everything work as expected

packages/resolver/src/adapters/unstoppable/api.ts
```diff
@@ -40,28 +40,23 @@ export const unstoppableApiGetCryptoAddress = (
...
 
         if (hasCardanoAddress)
           return response.value.data.records['crypto.ADA.address']
 
         if (hasOtherBlockchainAddress)
           throw new Resolver.Errors.WrongBlockchain()
 
-        if (hasNotAnyAddress) throw new Resolver.Errors.NotFound()
-
         throw new Resolver.Errors.InvalidResponse()
```
banklesss, line 60: keeping this to cover the case when a wrong result is resolved

i.e.

`{ '...ADA...': undefined }`

packages/resolver/src/adapters/unstoppable/api.ts
```diff
@@ -40,28 +40,23 @@ export const unstoppableApiGetCryptoAddress = (
...
 
+        // checking
         const hasCardanoAddress = safeParsedAdaResponse.success
-        const hasOtherBlockchainAddress =
-          !hasCardanoAddress &&
-          safeParsedGeneralResponse.success &&
-          Object.keys(response.value.data.records).length > 0
-        const hasNotAnyAddress =
-          !hasOtherBlockchainAddress && safeParsedGeneralResponse.success
+        const hasOtherBlockchainAddress = safeParsedGeneralResponse.success
```
banklesss, line 52:
```
const UnstoppableApiGeneralResponseSchema = z.object({
  records: z.record(z.string(), z.string()),
})
```

stackchain (approved):
lgtm

PR #3042, opened 2024-01-25
fix: remove unused keyboard spacer

banklesss:


apps/wallet-mobile/src/features/Settings/ManageCollateral/ConfirmTx/ConfirmTxScreen.tsx
```diff
@@ -83,8 +83,6 @@ export const ConfirmTxScreen = () => {
               testID="spendingPasswordInput"
             />
           )}
-
-          <KeyboardSpacer />
```
michalszorad, line 87: What issues was this causing?

banklesss: with `KeyboardAvoidingView` is unnecessary

in fact it is the cause of the problem

michalszorad: I see, the screen got bigger when the keyboard showed up

michalszorad (approved):

jorbuedo (approved):

PR #3044, opened 2024-01-25
feature(receive): Receive new design funnel [delivery]

euvini:


stackchain:
develop was out of sync before opening it.

PR #3050, opened 2024-01-26
refactor(api): Use hex format to check for asset supply

michalszorad:
Resolves [YOMO-1008](https://emurgo.atlassian.net/browse/YOMO-1008)

[YOMO-1008]: https://emurgo.atlassian.net/browse/YOMO-1008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
```diff
@@ -43,12 +43,16 @@ export const parseNFT = (
   const nameHex = toAssetNameHex(tokenId)
   const displayAssetName = toDisplayAssetName(tokenId)
 
+  const hexSupplyKey = `${policyId}.${nameHex}`
+  const nameSupplyKey = `${policyId}.${displayAssetName}`
+  const assetSupply = assetSupplies[hexSupplyKey] ?? assetSupplies[nameSupplyKey]
+
+  if (assetSupply !== '1') return null
```
michalszorad, line 49: skips unnecessary parsing if supply is not 1

jorbuedo (approved):

banklesss (approved):

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
```diff
@@ -43,12 +43,16 @@ export const parseNFT = (
   const nameHex = toAssetNameHex(tokenId)
   const displayAssetName = toDisplayAssetName(tokenId)
```
stackchain, line 44: @MichalSzorad `toDisplayAssetName` will remove the `tag` if CIP67, I think the API returns the name as it is (with tags). I think it needs to use `toAssetName`

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
```diff
@@ -43,12 +43,16 @@ export const parseNFT = (
...
 
+  const hexSupplyKey = `${policyId}.${nameHex}`
```
stackchain: maybe: withHexName, withUtf8Name

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
michalszorad: Having a look

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
stackchain: can you apply on `@yoroi/api` -> src/cardano as well ?

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
banklesss: the tag is in hex format, but ascii name doesn't have tag when used in ids if i am not wrong

michalszorad: I've removed the ascii check in supply, as it's actually not needed

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
michalszorad: Updated

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
michalszorad: Looks like that function is not used. Is it worth removing the function from metadata.ts and using only the one from `@yoroi/api` ?

stackchain: please update it just for sanity, later with the portfolio manager it will be cleaned up.

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.test.ts
```diff
@@ -27,82 +21,42 @@ describe('parseNFT', () => {
     const assetMetadatas = [nft]
     const assetSupplies = {[nftId]: '1'}
 
-    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-    // @ts-ignore
-    convertNft.mockImplementation(() => parsedNft)
-
-    expect(parseNFT(assetMetadatas, assetSupplies, policyId, nameHex, config)).toEqual(parsedNft)
```
michalszorad, line 34: Removed mocks and used actual `parseNFT` as the old test had wrong arguments

apps/wallet-mobile/src/yoroi-wallets/cardano/api/metadata.ts
michalszorad: That's now updated

PR #3083, opened 2024-02-15
fix(wallet-mobile): rn init twice + restore state crash

stackchain:
- Fixed react-native-screens double [initialization](https://github.com/software-mansion/react-native-screens#android)
- Fixed for react-native-screens activity when restoring state (no other state should be mixed into)
- Enabled react-freeze (memo at rn level)


apps/wallet-mobile/ios/Podfile.lock
```diff
@@ -959,4 +959,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 4045625f1556b4d232075df8dbd22524a7010c3f
 
-COCOAPODS: 1.14.3
```
banklesss: this change is really needed?

stackchain: - leftovers from RN 73, definitely not needed now, reverting it.

michalszorad (approved):

jorbuedo (approved):

banklesss (approved):

PR #3091, opened 2024-02-26
feature(wallet-mobile): add theme support for settings & collateral funnels

SorinC6:
task: https://emurgo.atlassian.net/browse/YOMO-1177


apps/wallet-mobile/src/features/Settings/EasyConfirmation/EnableEasyConfirmationScreen.tsx
```diff
@@ -115,27 +119,33 @@ const messages = defineMessages({
...
+  const styles = StyleSheet.create({
+    container: {
+      flex: 1,
+      backgroundColor: color.gray.min,
+    },
+    contentContainer: {
+      padding: 16,
+    },
+    heading: {
+      fontSize: 16,
```
stackchain: can it consume any typography?

apps/wallet-mobile/src/features/Settings/EasyConfirmation/EnableEasyConfirmationScreen.tsx
```diff
@@ -115,27 +119,33 @@ const messages = defineMessages({
...
+      lineHeight: 24,
+      paddingBottom: 20,
+    },
+    warning: {
+      color: color.magenta[500],
+      fontSize: 14,
```
stackchain: :arrow_up:

apps/wallet-mobile/src/features/Settings/EasyConfirmation/EnableEasyConfirmationScreen.tsx
SorinC6: yes!l let me check all the `fontSize` again

stackchain (approved):

jorbuedo:
Every time I see the `useStyles` functions I worry a bit about react performance.

Inside, useTheme is a useContext, so it should only trigger renders when the provider actually changes the objects inside (theme, colorScheme, etc.). Which is fine.

But other causes of rerenders in the parent component will basically run useStyles again, which calls StyleSheet.create with a possibly big object full of spreads inside. 

I'm not sure if the cost of all these calls vs the cost of adding memos everywhere would be worth it though.

jorbuedo (approved):

PR #3094, opened 2024-02-27
fix(wallet-mobile): Design issues

SorinC6:
tasks: 
- https://emurgo.atlassian.net/browse/YOMO-743
- https://emurgo.atlassian.net/browse/YOMO-744
- https://emurgo.atlassian.net/browse/YOMO-745

⚠️ Many sections of the application are still using legacy styles and `bold` font. This will only be solved when all the application will use the typography/colors from the theme in the correct way, this revamp is on the way. The current PR is solving only the issues presented in the tasks


apps/wallet-mobile/src/NftDetails/NftDetails.tsx
```diff
@@ -126,72 +130,66 @@ const MetadataRow = ({title, copyText, children}: {title: string; children: Reac
...
-            <Spacer width={2} />
-
-            <Text style={styles.linkText}>Cexplorer</Text>
-          </View>
-        </TouchableOpacity>
+        <View
+          style={{
+            display: 'flex',
+            flexDirection: 'row',
+          }}
```
michalszorad, line 175: Inline styles

michalszorad (approved):

banklesss (approved):

PR #3104, opened 2024-03-04
fix(wallet-mobile): yoroi zero mainnet url

jorbuedo:
Ticket: https://emurgo.atlassian.net/browse/YOMO-1224


stackchain (approved):

jorbuedo:
Seems to be unstable right now. Clarifying with infra.

stackchain:
> Seems to be unstable right now. Clarifying with infra.

My guess is that is a slow query that is taking longer than timeout on gateway, so it's returning the `504` on proxy, yet the query is still running, need to check with BE as well.

PR #3113, opened 2024-03-06
chore: Update packages

jorbuedo:
Need to double check the value locked, not sure I got that right.


apps/wallet-mobile/src/yoroi-wallets/cardano/constants/common.ts
```diff
@@ -58,4 +58,5 @@ export const HISTORY_REFRESH_TIME = 35000
 
 export const COINS_PER_UTXO_WORD = '34482'
 
+// TODO: This should come from protocolParams endpoint
```
stackchain, line 61: @pedromtcosta any estimate for BE zero to be updated?

banklesss:
@jorbuedo byron wallet needs update too

banklesss (approved):

stackchain (approved):

apps/wallet-mobile/src/yoroi-wallets/cardano/constants/common.ts
jorbuedo: We do have the endpoint, not sure if there's another update is necessary on BE. Maybe for the caching so ogmios doesn't struggle so much.
I changed useLockedAmount to use the data from the endpoint, but I think there's other places in the app that still use this (and maybe other?) constant and should all be changed.

apps/wallet-mobile/src/yoroi-wallets/cardano/assetUtils/assetUtils.ts
```diff
@@ -1,19 +1,24 @@
...
 
-export async function calcLockedDeposit(utxos: RawUtxo[], networkId: NetworkId) {
-  const networkConfig = getCardanoNetworkConfigById(networkId)
-  const coinsPerUtxoWord = await CardanoMobile.BigNum.fromStr(networkConfig.COINS_PER_UTXO_WORD)
+export async function calcLockedDeposit(utxos: RawUtxo[], address: Address, coinsPerUtxoByteStr: string) {
   const utxosWithAssets = utxos.filter((u) => u.assets.length > 0)
 
+  const coinsPerUtxoByte = await CardanoMobile.BigNum.fromStr(coinsPerUtxoByteStr ?? COINS_PER_UTXO_BYTE)
+  const dataCost = await CardanoMobile.DataCost.newCoinsPerByte(coinsPerUtxoByte)
+  const normalizedAddress = await normalizeToAddress(address)
```
lisicky, line 14: @jorbuedo is it real address of utxo ?

apps/wallet-mobile/src/yoroi-wallets/cardano/assetUtils/assetUtils.ts
```diff
@@ -1,19 +1,24 @@
...
+  if (normalizedAddress === undefined) throw new Error('calcLockedDeposit::Error not a valid address')
+
   const promises = utxosWithAssets.map((u) => {
     return cardanoValueFromRemoteFormat(u)
-      .then((v) => CardanoMobile.minAdaRequired(v, false, coinsPerUtxoWord))
-      .then((v) => v.toStr())
+      .then((v) => CardanoMobile.TransactionOutput.new(normalizedAddress, v))
```
lisicky, line 19: Do we 100% sure that utxo doesn't  have an inlined datum/script ?

stackchain: I think the API doesn't return it now. cc: @pedromtcosta

lisicky: @stackchain  It's not case for casual wallet user, but anyway from blockchain perspective it is possible. If it worked before it will work for now,  but this edge case still will not be covered

apps/wallet-mobile/src/yoroi-wallets/cardano/assetUtils/assetUtils.ts
jorbuedo: It's a `wallet.receiveAddresses[0]`. From what I understood, I needed a wallet's address for the next operations?

lisicky: @jorbuedo it depends where data come from, because if it is change utxo ( I assume `wallet.receiveAddresses[0] ` is change address ) it is okay. But if you are trying to calc min ada for utxo that you receives from API, you need to use address of utxo not just wallet.receiveAddresses[0].
But there is also trick, if you are sure that all utxos have a BaseAddress , the `wallet.receiveAddresses[0]`  will work. 
I guess if it non standard wallet the `wallet.receiveAddresses[0]`  will work but keep in mind that there is this edge case.

lisicky (approved):
LGTM, except edge cases that I mentioned

apps/wallet-mobile/src/yoroi-wallets/cardano/assetUtils/assetUtils.ts
stackchain: actually this is a limitation on mobile side, it only works with BaseAddress, so if we sync from another wallet a pointer or a enterprise, it doesn't add up to the balance, and definitely will consider more min ada than needed.

PR #3129, opened 2024-03-13
fix(wallet-mobile): Removes tx history animate down

jorbuedo:
Looks like I was wrong with the cause of this. I have no idea why it breaks. It just does.
This removes the listeners. TxHistory Animates up, once, when you select a wallet. And stays up.
I've run it 10 times the QR camera in and out. Doesn't break anymore. 

realted: https://emurgo.atlassian.net/browse/YOMO-1222?focusedCommentId=21955


stackchain:
closing in favour of #3135 by Product decision.

PR #3132, opened 2024-03-14
fix(receive): Add last used date to receive address

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-1230

image

### DIFF
--- a/apps/wallet-mobile/src/features/Receive/common/ShareDetailsCard/ShareDetailsCard.tsx
+++ b/apps/wallet-mobile/src/features/Receive/common/ShareDetailsCard/ShareDetailsCard.tsx
@@ -1,10 +1,12 @@
 import {useTheme} from '@yoroi/theme'
-import React from 'react'
+import React, {useMemo} from 'react'
+import {useIntl} from 'react-intl'
 import {StyleSheet, useWindowDimensions, View} from 'react-native'
 import LinearGradient from 'react-native-linear-gradient'
 
 import {CopyButton, Text} from '../../../../components'
 import {useMetrics} from '../../../../metrics/metricsManager'
+import {useSelectedWallet} from '../../../../SelectedWallet'
 import {isEmptyString} from '../../../../utils/utils'
 import {useStrings} from '../useStrings'
 
@@ -16,8 +18,22 @@ type AddressDetailsProps = {
 
 export const ShareDetailsCard = ({address, spendingHash, stakingHash}: AddressDetailsProps) => {
   const strings = useStrings()
+  const intl = useIntl()
   const {styles, colors} = useStyles()
   const {track} = useMetrics()
+  const wallet = useSelectedWallet()
+  const lastUsed = useMemo(
+    () =>
+      Object.values(wallet.transactions).reduce((currentLast, tx) => {
+        const {inputs, outputs} = tx
+        const isRelevant = inputs.some((v) => address === v.address) || outputs.some((v) => address === v.address)
+        if (!isRelevant) return currentLast
+        const lastUpdatedAt = new Date(tx.lastUpdatedAt).getTime()
+        console.log('>> ', tx.lastUpdatedAt)
+        return Math.max(currentLast, lastUpdatedAt)
+      }, 0),
+    [address, wallet.transactions],
+  )
 
   const hasStakingHash = !isEmptyString(stakingHash)
   const hasSpendingHash = !isEmptyString(spendingHash)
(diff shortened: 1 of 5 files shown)


apps/wallet-mobile/src/features/Receive/common/ShareDetailsCard/ShareDetailsCard.tsx
```diff
@@ -93,12 +125,12 @@ const useStyles = () => {
       alignSelf: 'center',
       overflow: 'hidden',
       paddingHorizontal: 16,
+      paddingVertical: 32,
```
SorinC6:
```suggestion
      ...padding['y-xxl],
```

SorinC6 (approved):

apps/wallet-mobile/src/features/Receive/common/ShareDetailsCard/ShareDetailsCard.tsx
```diff
@@ -16,8 +18,22 @@ type AddressDetailsProps = {
...
+  const lastUsed = useMemo(
+    () =>
+      Object.values(wallet.transactions).reduce((currentLast, tx) => {
+        const {inputs, outputs} = tx
+        const isRelevant = inputs.some((v) => address === v.address) || outputs.some((v) => address === v.address)
+        if (!isRelevant) return currentLast
+        const lastUpdatedAt = new Date(tx.lastUpdatedAt).getTime()
+
+        return Math.max(currentLast, lastUpdatedAt)
+      }, 0),
```
michalszorad: Can this be extracted into some helper function or hook?

apps/wallet-mobile/src/features/Receive/common/ShareDetailsCard/ShareDetailsCard.tsx
```diff
@@ -70,6 +86,22 @@ export const ShareDetailsCard = ({address, spendingHash, stakingHash}: AddressDe
           </View>
         </View>
       )}
+
+      {Boolean(lastUsed) && (
```
michalszorad, line 76: maybe just `!!lastUsed`

jorbuedo: I was being nice because people on the internet consider !! unclear, I couldn't care less 😄

apps/wallet-mobile/src/features/Receive/common/ShareDetailsCard/ShareDetailsCard.tsx
jorbuedo: Tbh if this was some info that were needed often, it should be provided by the wallet, transactionManager and all of that. Which is were we are currently getting the list of transaction ids an address has used.

Considering it's a one off requirement for this card and that no other place needs to know when was the last time an address was used, I would say here is the best place.

michalszorad: At least for readability let's move it to a hook in this file

michalszorad (approved):

apps/wallet-mobile/src/features/Receive/common/ShareDetailsCard/useLastDateAddressUsed.ts
```diff
@@ -0,0 +1,28 @@
...
+      }, 0),
+    [address, wallet.transactions],
+  )
+  return lastUsed
+    ? intl.formatDate(new Date(lastUsed), {
+        dateStyle: 'short',
+        timeStyle: 'short',
+        hour12: false,
+      })
+    : null
```
stackchain:
```suggestion
  return React.useMemo(
    () =>
      Object.values(wallet.transactions).reduce((currentLast, tx) => {
        const {inputs, outputs} = tx
        const isRelevant = inputs.some((v) => address === v.address) || outputs.some((v) => address === v.address)
        if (!isRelevant) return currentLast
        const lastUpdatedAt = new Date(tx.lastUpdatedAt).getTime()

        const lastUsed = Math.max(currentLast, lastUpdatedAt)
        const formattedLastUsed = lastUsed
    ? intl.formatDate(new Date(lastUsed), {
        dateStyle: 'short',
        timeStyle: 'short',
        hour12: false,
      })
    : null
      }, 0),
    [address, wallet.transactions],
  )
```

jorbuedo: I don't think that would work, the reducer uses a number as accumulator, not a formatted date.

jorbuedo: If the intention is to just memo the number to formatted as well, I can move that inside the memo, but not inside the reducer.

stackchain: it was a suggestion, to collocate the logic, and don't kick the hook entirely since there will be only hooks in the body, it needs to adjust the memo function `{..yayada, return..}`

stackchain:
imho it'd be much better if it includes the last tx, instead of the last date, so as a user I could check the last tx 🤔

stackchain (approved):

PR #3135, opened 2024-03-14
fix(wallet-mobile): tx history animation flick

stackchain:
- The Animated.View was conflicting with the react-native-navigation default animation, therefore both animations were fighting to complete
- If its needed to translate X/Y it needs to happen on the navigation animation (route)
- Fixed the StatusBar on qr-code (it was hidden it), yet it should be always dark as the QR code screen, otherwise it shows the background of it.

Relates to #3219


stackchain:
@jorbuedo can you test it?

apps/wallet-mobile/src/TxHistory/TxHistory.tsx
```diff
@@ -57,69 +53,67 @@ export const TxHistory = () => {
 
   return (
     <View style={styles.scrollView}>
-      <Animated.View style={[styles.container, translateStyles]}>
```
michalszorad, line 60: So we've decided not to use the custom animation?

michalszorad (approved):

jorbuedo (approved):

apps/wallet-mobile/src/TxHistory/TxHistory.tsx
stackchain: yap Product just approved, it will be done later - in the navigation.

PR #3137, opened 2024-03-15
fix(wallet-mobile): modal double close

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-1244

### DIFF
--- a/apps/wallet-mobile/src/components/Modal/ModalContext.tsx
+++ b/apps/wallet-mobile/src/components/Modal/ModalContext.tsx
@@ -36,8 +36,10 @@ export const ModalProvider = ({
   const navigation = useNavigation()
   const actions = React.useRef<ModalActions>({
     closeModal: () => {
-      dispatch({type: 'close'})
-      navigation.goBack()
+      if (navigation.getState().routes.slice(-1)[0].name === 'modal') {
+        dispatch({type: 'close'})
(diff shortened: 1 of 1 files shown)


banklesss (approved):

banklesss:
👍🏻👌🏻👏🏻

apps/wallet-mobile/src/components/Modal/ModalContext.tsx
```diff
@@ -36,8 +36,10 @@ export const ModalProvider = ({
   const navigation = useNavigation()
   const actions = React.useRef<ModalActions>({
     closeModal: () => {
-      dispatch({type: 'close'})
-      navigation.goBack()
+      if (navigation.getState().routes.slice(-1)[0].name === 'modal') {
```
stackchain: 2 points here:
```const lastRouteName = navigation.getState().routes.slice(-1)[0].name``` for readability, and if I'm not mistaken react-navigation eventually adds to the name `key-<random>-<name>` but I don't recall it precisely, if it still does, then I'd say to check with something like "endsWith", print the navigation.getState(), and give it a check please, maybe I'm just delusional.

jorbuedo: `key-<random>-<name>` is the key, the name is just the name

PR #3144, opened 2024-03-18
fix(exchange): access token management

banklesss:
Related to [YOMO-1252](https://emurgo.atlassian.net/browse/YOMO-1252)

[YOMO-1252]: https://emurgo.atlassian.net/browse/YOMO-1252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/exchange/src/adapters/api.test.ts
```diff
@@ -49,39 +53,101 @@ describe('exchangeApiMaker', () => {
...
+            })),
+          },
+        }
+
+        const api = exchangeApiMaker({isProduction, partner}, deps)
+
+        try {
+          await api.getBaseUrl(providerId)
+        } catch (e: any) {
+          expect(e).toBeInstanceOf(Exchange.Errors.ProviderNotFound)
```
michalszorad: The test may pass if the `api.getBaseUrl` does not throw. Maybe additional test that expects the fn to throw?

stackchain: await expect(() => api.getBase...).toThrow...

packages/exchange/src/adapters/api.test.ts
```diff
@@ -49,39 +53,101 @@ describe('exchangeApiMaker', () => {
...
+          banxaApi: {
+            getBaseUrl: jest
+              .fn()
+              .mockReturnValue(jest.fn().mockResolvedValue('banxa-url')),
+          },
+          encryptusApi: {
+            getBaseUrl: jest
+              .fn()
+              .mockReturnValue(jest.fn().mockResolvedValue('encryptus-url')),
+            extractParamsFromBaseUrl: jest.fn(() => ({
```
stackchain: I don't think it needs to be part of the API, cuz the expected return of the manager.referralLink -> is to return the referral link complete, this is implementation, for me this should be extracted as a helper, and be part of the create referral link, anytime the baseUrl contains an access_token, it adds in the referral link returned.

stackchain: like doesn't even matter the provider, if there is a `access_token` it will include it.

packages/exchange/src/adapters/encryptus/api.test.ts
```diff
@@ -130,3 +131,22 @@ describe('getEncryptusBaseUrl', () => {
     expect(getCryptoAddress).toBeDefined()
   })
 })
+
+describe('encryptusExtractParamsFromBaseUrl', () => {
+  it('extracts the params from the base url', () => {
+    const baseUrl = `https://sandbox.encryptus.co/pw?access_token=FAKE_TOKEN`
+    const params = encryptusExtractParamsFromBaseUrl(baseUrl.toString())
+    expect(params).toEqual({access_token: 'FAKE_TOKEN'})
```
stackchain: this validation should be included in the API return -> after parsing the schema, if the schema is correct, then check if the data is correct.

banklesss: ok i will simplify this

stackchain: it can throw Api.Error.MalformatedResponse

michalszorad (approved):

packages/exchange/src/manager.ts
```diff
@@ -49,15 +49,31 @@ export const exchangeManagerMaker = ({
...
+              accessToken !== null
+                ? {...queries, access_token: accessToken}
+                : queries
+
+            const validatedQueries =
+              urlReferralQueryStringParamsSchema.parse(allQueries)
+
             const params = new URLSearchParams()
             for (const [key, value] of Object.entries(validatedQueries)) {
               params.append(key, value.toString())
```
stackchain:
```suggestion
            const parsedQueries =
              urlReferralQueryStringParamsSchema.parse(allQueries)

            const params = new URLSearchParams()
            for (const [key, value] of Object.entries(parsedQueries)) {
              params.append(key, value.toString())
```

stackchain (approved):

PR #3145, opened 2024-03-18
feature(swap): Use new completed orders endpoint

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-1253


apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/CompletedOrders.tsx
```diff
@@ -190,7 +155,7 @@ export const ExpandableOrder = ({order, tokenInfos}: {order: MappedRawOrder; tok
         sellLabel={sellLabel}
         buyLabel={buyLabel}
         tokenAmount={`${buyQuantity} ${buyLabel}`}
-        txTimeCreated={intl.formatDate(new Date(order.date), {
+        txTimeCreated={intl.formatDate(new Date(order.placedAt), {
```
stackchain: it looks they are not returning unix format

stackchain:
I've marked as blocked, it looks like it is listing only some LPs, but not all of them, of my 25 orders, only 4 were listed  : (

jorbuedo:
> I've marked as blocked, it looks like it is listing only some LPs, but not all of them, of my 25 orders, only 4 were listed : (

Really? I have 5 of 5. Could you check directly in muesliswap and see if they're there?
https://muesliswap.com/wallet/orders-history?period=all_time&contract=all&type=all&status=all

apps/wallet-mobile/src/features/Swap/useCases/StartSwapScreen/ListOrders/CompletedOrders.tsx
jorbuedo: Right, I saw a number and just assumed. Needs to be multiplied by 1000.

jorbuedo:
Even if it's not working 100%, I'm gonna remove the changes in mobile but still update the swap package. This function was unused in mobile anyway because the previous endpoint didn't work either. This new one may be useful for extension while there's no custom backend for this.

stackchain:
> Even if it's not working 100%, I'm gonna remove the changes in mobile but still update the swap package. This function was unused in mobile anyway because the previous endpoint didn't work either. This new one may be useful for extension while there's no custom backend for this.

That's a good idea. also since you didn't forced push we can pick up later ;)

stackchain:
@jorbuedo can you bump the versions swap + anything importing it. So I can publish for @vsubhuman

stackchain (approved):

PR #3149, opened 2024-03-20
fix(exchange): manager link generation

banklesss:


michalszorad (approved):

packages/exchange/src/adapters/base-url-adapter.ts
```diff
@@ -0,0 +1,8 @@
+export const baseUrlAdapter = (baseUrl: string, providerId: string) => {
+  switch (providerId) {
+    case 'encryptus':
```
stackchain, line 3: just for history: Encryptus has this limitation on their end where the path MUST end with `/` which is not part of URIs definition.

stackchain (approved):

PR #3179, opened 2024-04-04
fix(receive): Adjust QR card spacing

jorbuedo:


michalszorad (approved):

rahulnr7 (approved):
Confirmed with Larissa also.

PR #3183, opened 2024-04-08
fix(wallet-mobile): Add copy tool tip on address details

rahulnr7:
Fixes [YOMO-1285](https://emurgo.atlassian.net/browse/YOMO-1285)

After  fix:

Image

[YOMO-1285]: https://emurgo.atlassian.net/browse/YOMO-1285?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/components/CopyButton.tsx
+++ b/apps/wallet-mobile/src/components/CopyButton.tsx
@@ -4,6 +4,7 @@ import {StyleProp, StyleSheet, TouchableOpacity, View, ViewStyle} from 'react-na
 import Animated, {FadeInDown, FadeOutDown, Layout} from 'react-native-reanimated'
 
 import {Text} from '../../../wallet-mobile/src/components'
+import {isEmptyString} from '../../../wallet-mobile/src/utils/utils'
 import {Icon} from '../components/Icon'
 import {useCopy} from '../legacy/useCopy'
 
@@ -46,7 +47,7 @@ const AnimatedCopyButton = ({
     <TouchableOpacity onPress={onCopy} disabled={isCopying} testID="copyButton" style={style}>
       {isCopying ? (
         <View style={styles.rowContainer}>
-          {message != null ? (
+          {!isEmptyString(message) ? (
             <Animated.View layout={Layout} entering={FadeInDown} exiting={FadeOutDown} style={styles.isCopying}>
               <Text style={styles.copiedText}>{message}</Text>
             </Animated.View>
--- a/apps/wallet-mobile/src/components/CopyButton.tsx
+++ b/apps/wallet-mobile/src/components/CopyButton.tsx
@@ -1,8 +1,11 @@
 import {useTheme} from '@yoroi/theme'
 import React from 'react'
-import {StyleProp, TouchableOpacity, ViewStyle} from 'react-native'
+import {StyleProp, StyleSheet, TouchableOpacity, View, ViewStyle} from 'react-native'
+import Animated, {FadeInDown, FadeOutDown, Layout} from 'react-native-reanimated'
 
+import {Text} from '../../../wallet-mobile/src/components'
 import {Icon} from '../components/Icon'
+import {useStrings} from '../features/Receive/common/useStrings'
 import {useCopy} from '../legacy/useCopy'
 
 export type CopyButtonProps = {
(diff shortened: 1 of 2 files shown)


apps/wallet-mobile/src/components/CopyButton.tsx
```diff
@@ -29,17 +33,32 @@ export const CopyButton = ({value, onCopy, children, style}: CopyButtonProps) =>
...
+}: Omit<CopyButtonProps, 'value'> & {isCopying: boolean; message?: string | null | undefined}) => {
+  const {styles, colors} = useStyles()
 
   return (
     <TouchableOpacity onPress={onCopy} disabled={isCopying} testID="copyButton" style={style}>
-      {isCopying ? <Icon.CopySuccess size={26} color={colors.gray} /> : <Icon.Copy size={26} color={colors.gray} />}
+      {isCopying ? (
+        <View style={styles.rowContainer}>
+          {message != null ? (
+            <Animated.View layout={Layout} entering={FadeInDown} exiting={FadeOutDown} style={styles.isCopying}>
```
michalszorad, line 51: Didn't we use any library to display the tooltips?

apps/wallet-mobile/src/components/CopyButton.tsx
```diff
@@ -29,17 +33,32 @@ export const CopyButton = ({value, onCopy, children, style}: CopyButtonProps) =>
...
-}: Omit<CopyButtonProps, 'value'> & {isCopying: boolean}) => {
-  const {colors} = useStyles()
+  message,
+}: Omit<CopyButtonProps, 'value'> & {isCopying: boolean; message?: string | null | undefined}) => {
+  const {styles, colors} = useStyles()
 
   return (
     <TouchableOpacity onPress={onCopy} disabled={isCopying} testID="copyButton" style={style}>
-      {isCopying ? <Icon.CopySuccess size={26} color={colors.gray} /> : <Icon.Copy size={26} color={colors.gray} />}
+      {isCopying ? (
```
michalszorad, line 48: Could be simplified to `isCopying && message !== null && <View>...`

apps/wallet-mobile/src/components/CopyButton.tsx
```diff
@@ -29,17 +33,32 @@ export const CopyButton = ({value, onCopy, children, style}: CopyButtonProps) =>
   )
 }
 
-const AnimatedCopyButton = ({
+export const AnimatedCopyButton = ({
```
michalszorad: Is this export needed? I don't see it imported in this PR

apps/wallet-mobile/src/components/CopyButton.tsx
```diff
@@ -29,17 +33,32 @@ export const CopyButton = ({value, onCopy, children, style}: CopyButtonProps) =>
...
   onCopy,
   children,
   style,
   isCopying,
-}: Omit<CopyButtonProps, 'value'> & {isCopying: boolean}) => {
-  const {colors} = useStyles()
+  message,
+}: Omit<CopyButtonProps, 'value'> & {isCopying: boolean; message?: string | null | undefined}) => {
```
michalszorad: `message?: string` should be enough

rahulnr7: Updated, thanks!

apps/wallet-mobile/src/components/CopyButton.tsx
rahulnr7: Updated, thanks!

apps/wallet-mobile/src/components/CopyButton.tsx
rahulnr7: @michaeljscript Thanks for the review.
The component is used in few places where it is not passing any string message(as it was earlier), so if we combine the two conditions, this line <Icon.CopySuccess size={26} color={colors.gray} /> will not work for calls thats have an empty message. 
I am not sure if this is the best way. Please advise if it could be made simpler.

apps/wallet-mobile/src/components/CopyButton.tsx
```diff
@@ -29,17 +33,32 @@ export const CopyButton = ({value, onCopy, children, style}: CopyButtonProps) =>
...
+  const {styles, colors} = useStyles()
 
   return (
     <TouchableOpacity onPress={onCopy} disabled={isCopying} testID="copyButton" style={style}>
-      {isCopying ? <Icon.CopySuccess size={26} color={colors.gray} /> : <Icon.Copy size={26} color={colors.gray} />}
+      {isCopying ? (
+        <View style={styles.rowContainer}>
+          {message != null ? (
```
stackchain: otherwise empty strings i.e '' will end up show an empty tooltip, I'd advise to rename the prop from `message` to `withAnimatedSuccessMessage`, message in this case is too generic.
```suggestion
          {!isEmptyString(message) ? (
```

apps/wallet-mobile/src/components/CopyButton.tsx
rahulnr7: No, could not find one. Other tool tips are also implemented the same way. for "Address Copied' message.

apps/wallet-mobile/src/components/CopyButton.tsx
rahulnr7: Thanks @stackchain . Will include this along with the other changes.

stackchain (approved):
later let's find a library, the anchor now is at bottom, dunno if all cases can rely on it.

PR #3184, opened 2024-04-08
feature(pool-transition): Adds Pool transition modal and warnings

jorbuedo:
related: 
- https://emurgo.atlassian.net/browse/YOMO-1272
- https://www.figma.com/file/KeBVo6rWNYFLXwTphz9lHi/Mobile-App?type=design&node-id=16592-16113&mode=design&t=dExxblZVpV7zpAvv-0

Testing wallets are the production wallets, listed on https://github.com/Emurgo/yoroi-lib/blob/main/src/pools.ts#L42

|TxHistory|Settings|Staking Center|Failed|
|-|-|-|-|
|image|image|image|image|

Missing:

- yoroi-lib update with PoolTransition and getPoolTransition


apps/wallet-mobile/src/Dashboard/Dashboard.tsx
```diff
@@ -197,6 +209,37 @@ const useCurrentTime = () => {
   return currentTime
 }
 
+const PoolTransitionNotice = () => {
```
michalszorad: Maybe worth moving this component to a new file

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,396 @@
...
+import {Quantities} from '../../yoroi-wallets/utils'
+
+type PoolTransition = {
+  currentPool: PoolInfo
+  suggestedPool: PoolInfo
+  deadlineMilliseconds: number
+}
+
+const getPoolTransition = async (currentPoolId: string): Promise<PoolTransition | null> => {
+  return new Promise((resolve) => {
```
michalszorad: Since the function is `async`, there is no need to return `Promise`

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,396 @@
...
+    if (currentPoolId === 'd248ded3c18e0e80d07a46f00a2d808075b989ccb1a0e40a76e5cee1') {
+      resolve({
+        currentPool: {name: 'emurgo old', roa: '0', share: '3.5'} as unknown as PoolInfo,
+        suggestedPool: {
+          id: 'df1750df9b2df285fcfb50f4740657a18ee3af42727d410c37b86207',
+          name: 'emurgo new',
+          roa: '5.1',
+          share: '3.2',
+        } as unknown as PoolInfo,
+        deadlineMilliseconds: 2999777000,
```
michalszorad: Should these values be hardcoded? If yes, maybe let's move them into some config file

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,396 @@
...
+      resolve(null)
+    }
+  })
+}
+
+const MS_IN_SEC = 1000,
+  SEC_IN_DAY = 86400,
+  SEC_IN_HOUR = 3600,
+  SEC_IN_MIN = 60
+export const formatTimeSpan = (ms: number) => {
```
michalszorad: Lets move this to utils

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
jorbuedo: Indeed, I was just following along with how the rest of sub-components were done in that file. I may move them to the same folder as the hooks

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Are you sure? It complains without it. It's a mock anyway, this needs to com from lib.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: This is a mock, getPoolTransition needs to be implemented in yoroi-lib

jorbuedo:
@stackchain  Let me know if we want to have it on develop with the mock or block the PR till yoroi-lib is ready.

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
```diff
@@ -114,16 +125,16 @@ export const Dashboard = () => {
...
           {stakingInfo?.status === 'staked' && (
-            <Row>
+            <View style={styles.row}>
               <StakePoolInfos />
-            </Row>
+            </View>
           )}
         </ScrollView>
 
-        <Actions>
```
stackchain: favour readability

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+import {formatTimeSpan, Quantities} from '../../yoroi-wallets/utils'
+
+// TODO: PoolTransition and getPoolTransition will come from yoroi-lib
+type PoolTransition = {
+  currentPool: PoolInfo & {tax: number}
+  suggestedPool: PoolInfo & {tax: number}
+  deadlineMilliseconds: number
+}
+
+const getPoolTransition = async (currentPoolId: string): Promise<PoolTransition | null> => {
```
stackchain: I believe this can be called getPoolInfo, the deadlineInMs, should happen always when the pool is set to retire, also a great candidate to become part of yoroi/staking

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+  const wallet = useSelectedWallet()
+  const {stakingInfo} = useStakingInfo(wallet)
+
+  const isStaked = stakingInfo?.status === 'staked'
+  const currentPoolId = isStaked ? stakingInfo?.poolId : ''
+
+  const poolTransitionQuery = useQuery({
+    enabled: isStaked,
+    retry: false,
+    staleTime: 5 * 60 * 1000,
```
stackchain: 5m is default 
```suggestion
```

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+    queryKey: [wallet.id, 'poolTransition', currentPoolId],
+    queryFn: () => getPoolTransition(currentPoolId),
+  })
+
+  const poolTransition = poolTransitionQuery.data ?? null
+  const poolId = poolTransition?.suggestedPool.id ?? ''
+
+  const stakingTxQuery = useQuery({
```
stackchain:
```suggestion
  const createDelegationTx = useQuery({
```

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+    enabled: poolId !== '',
+    retry: false,
+    staleTime: 5 * 60 * 1000,
```
stackchain: it can cause the utxos to become stale since every block 20s

stackchain: since the user needs to confirm, maybe it should wait until the last moment to create the tx

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+    queryKey: [wallet.id, 'stakingTx', poolId],
+    queryFn: async () => {
+      const accountStates = await wallet.fetchAccountState()
+      const accountState = accountStates[wallet.rewardAddressHex]
+      // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
+      if (!accountState) throw new Error('Account state not found')
+
+      const stakingUtxos = await wallet.getAllUtxosForKey()
+      const amountToDelegate = Quantities.sum([
+        ...stakingUtxos.map((utxo) => utxo.amount as Balance.Quantity),
```
stackchain:
```suggestion
        ...stakingUtxos.map((utxo) => asQuantity(utxo.amount)),
```

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+
+        <View>
+          <Text style={[styles.regular, styles.red]}>
+            {isActive ? strings.poolWillStopRewards : strings.poolNoRewards}
+          </Text>
+
+          {isActive && (
+            <Text style={[styles.red, styles.bold]}>{formatTimeSpan(poolTransition.deadlineMilliseconds)}</Text>
+          )}
+        </View>
```
stackchain:
```suggestion
          <Text style={[styles.regular, styles.red]}>
            {isActive ? strings.poolWillStopRewards : strings.poolNoRewards}
          </Text>

          {isActive && (
            <Text style={[styles.red, styles.bold]}>{formatTimeSpan(poolTransition.deadlineMilliseconds)}</Text>
          )}
```

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+
+        <View style={styles.row}>
+          <Text style={[styles.regular, styles.gray]}>{strings.fee}</Text>
+
+          <Text style={[styles.bold, styles.max]}>{poolTransition.suggestedPool.tax} %</Text>
+        </View>
+
+        <View>
+          <Text style={[styles.regular, styles.max]}>{strings.poolGeneratesRewards}</Text>
+        </View>
```
stackchain:
```suggestion
          <Text style={[styles.regular, styles.max]}>{strings.poolGeneratesRewards}</Text>
```

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+            {isActive ? strings.poolWillStopRewards : strings.poolNoRewards}
+          </Text>
+
+          {isActive && (
+            <Text style={[styles.red, styles.bold]}>{formatTimeSpan(poolTransition.deadlineMilliseconds)}</Text>
+          )}
+        </View>
+      </View>
+
+      <Spacer height={8} />
```
stackchain: theming PR was merged, it can leverage the tokens for size

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+    },
+    max: {
+      color: theme.color.gray['max'],
+    },
+    red: {
+      color: theme.color.magenta['500'],
+    },
+    primary: {
+      color: theme.color.primary[600],
+    },
```
stackchain: into -> colors

apps/wallet-mobile/src/TxHistory/TxHistory.tsx
```diff
@@ -36,7 +37,10 @@ export const TxHistory = () => {
     setActiveTab(tab)
   }
 
-  const {sync, isLoading} = useSync(wallet)
+  const {sync, isLoading: isLoadingWallet} = useSync(wallet)
+  const {isLoading: isLoadingPoolTransition} = usePoolTransitionModal()
```
stackchain, line 41: since is requested while opening the wallet, it should swallow errors, otherwise a failed request won't allow the user to open the wallet, it will happen again when opening so it looks safe.

stackchain: it can also leverage the error boundary with a retry

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+
+  const poolTransitionQuery = useQuery({
+    enabled: isStaked,
+    retry: false,
+    staleTime: 5 * 60 * 1000,
+    queryKey: [wallet.id, 'poolTransition', currentPoolId],
+    queryFn: () => getPoolTransition(currentPoolId),
+  })
+
+  const poolTransition = poolTransitionQuery.data ?? null
```
stackchain, line 45: shouldTransition? if pool is retiring?

apps/wallet-mobile/src/features/Menu/Menu.tsx
```diff
@@ -39,9 +39,10 @@ export const MenuNavigator = () => {
 
 export const Menu = () => {
   const strings = useStrings()
-  const styles = useStyles()
+  const {styles, color} = useStyles()
   const navigateTo = useNavigateTo()
   const wallet = useSelectedWallet()
+  const {poolTransition} = usePoolTransition()
```
stackchain: that's why the tx creation should be on demand, if the pool is retiring, at this point the tx was created, and eventually it can even fail.

stackchain: isPoolRetiring

apps/wallet-mobile/src/features/Menu/Menu.tsx
```diff
@@ -198,6 +205,7 @@ const useNavigateTo = () => {
         },
       })
     },
+    staking: () => navigateToStakingDashboard(),
```
stackchain: if dashboard is a screen it should be called DashboardScreen and be wrapped in a SafeAreView (when it is not happening in the navigator)

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
```diff
@@ -239,45 +248,50 @@ const EpochInfo = () => {
...
+    },
+    actions: {
+      flexDirection: 'row',
+      backgroundColor: color.gray.min,
+      padding: 16,
+      borderTopLeftRadius: 8,
+      borderTopRightRadius: 8,
+      elevation: 1,
+      shadowOpacity: 0.06,
+      shadowColor: 'black',
```
stackchain: black static

apps/wallet-mobile/src/Staking/PoolTransition/PoolTransitionNotice.tsx
```diff
@@ -0,0 +1,67 @@
...
+import {StyleSheet, Text, View} from 'react-native'
+
+import {Button, Icon, Spacer} from '../../components'
+import {formatTimeSpan} from '../../yoroi-wallets/utils'
+import {usePoolTransition, useStrings} from './usePoolTransition'
+
+export const PoolTransitionNotice = () => {
+  const {styles, color} = useStyles()
+  const strings = useStrings()
+  const {poolTransition, navigateToUpdate, canTransitionPool} = usePoolTransition()
```
stackchain: tx created, move to on-demand

apps/wallet-mobile/src/Staking/PoolTransition/PoolTransitionNotice.tsx
stackchain, line 1: stories, with the states

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+      ])
+
+      return wallet.createDelegationTx(poolId, new BigNumber(amountToDelegate))
+    },
+  })
+
+  const yoroiUnsignedTx = stakingTxQuery.data
+
+  const navigateToUpdate = () =>
+    !yoroiUnsignedTx
```
stackchain: this is a UI responsability to disable it, the api should navigate.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: The function was defined by @vsubhuman and is to be implemented in yoroi-lib

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -0,0 +1,426 @@
...
+                poolId,
+                yoroiUnsignedTx,
+              },
+            },
+          },
+        })
+
+  return {
+    poolTransition,
+    canTransitionPool: !!yoroiUnsignedTx,
```
stackchain: did Product provide a "not enough balance" to create the TX, or the state for that is just, "not showing the modal" or "disabling the re-delegation" ?

stackchain (dismissed):
few points, also it would be nice to have tests for the in hooks functions

apps/wallet-mobile/src/yoroi-wallets/utils/timeUtils.ts
```diff
@@ -298,3 +298,18 @@ export function genToRealTime(
 }
 
 export const delay = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms))
+
+const MS_IN_SEC = 1000,
```
stackchain: it doesn't need to be capital imho

stackchain: can we have tests?

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
jorbuedo: I'd like to disagree that creating a component whose only purpose is to add a style is somehow more readable.
Specially considering the new component needs to call the useStyles hook to get that style.

So it's either

```
        <Actions>
          ...
        </Actions>
        
        const Actions = (props: ViewProps) => {
          const { styles } = useStyles()
          return <View {...props} style={styles.actions} />
        }
```

Or 
```
        <View style={styles.actions}>
          ...
        </View>
```

I believe the second one is more readable since:

- There's no `Actions` component that you have to figure out what it does (which is actually nothing)
- You don't have to create a nothing component that adds more lines to the file
- It's clear that this is a View with an `actions` style

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Also, this doesn't get the PoolInfo, it gets the PoolTransition, which indicate current PoolInfo, suggested PoolInfo and deadline till current pool is expired.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Ok, that will require to create the tx in the navigateToUpdate function. Only problem with that is we go back to needing to handle the error when this fails.

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
stackchain: - almost everything is View, how fast can you read ```<Head><Body><Content>```, and how fast can you read ```<View styles={[styles.background, styles.head, styles.padding]}> <View styles={[styles.padding, styles.container, styler...MAYBE IS THE BODY?]}>```
- with tiny specialization, in just a single read you can figure out what is on screen, View is implementation, and implementation "doesn't" (quotes ofc) matter. 
- this tiny specialization are so important that are even libraries designed for that, such as styled-components. (which I think we should consider later)
- imho, at least for me, I can read and understand 10x faster tiny specialization, over a lot of views. - that's why prioritize readability

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Are you just removing the View? That would increase the gap between these 2 components.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: it doesn't make sense to me to have 2 different endpoints for that, whenever the pool info is requested it should return if the pool is retiring or not cc: @vsubhuman

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: `shouldTransition` sounds like a boolean, and isn't very specific as to what needs transitioning, which is pool.

`poolTransition` existing or being null already signifies the need to transition or not  🤔

apps/wallet-mobile/src/features/Menu/Menu.tsx
jorbuedo: Sure, we can add a boolean to indicate whether the poolTransition is not null. But in this case, doesn't affect the tx creation.

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
jorbuedo: Yeah, missed that when converting the raw styles to useStyles.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Navigate where? If the unsignedTx failed you can't navigate there. The UI is already checking canNavigate, which make sure this condition doesn't happen. 

We're discussing when to create the tx in other threads anyway, so resolving this one.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: I asked for their input, and a tentative new screen for that was suggested: https://www.figma.com/file/KeBVo6rWNYFLXwTphz9lHi/Mobile-App?type=design&node-id=20568-39794&mode=design&t=P59sb3i9dFWiSh4I-0

I suggested the course of not showing the modal if the tx failed to be created, until the portfolio changes affected the staking center.

apps/wallet-mobile/src/features/Menu/Menu.tsx
jorbuedo: I just copied what the footer link had and it worked. 
I'll rename it and look into the SafeAreaView.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Added `isPoolRetiring` as a boolean checking it `poolTransition` exists or not.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Not sure I understand this one. They're colors, but I want to use them as styles for text. If a move them to the colors object, I'd still need them as styles for the text.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: 5m is default for cacheTime. Default staleTime is 0.

By default, every query gets executed again every time. The only thing the cache does is return early the previous result while computing the new one, and then update. It's a terrible default in my opinion, designed for non critical network fetches and not for computed values.
To not reexecute the queries, staleTime needs to be specified.

apps/wallet-mobile/src/TxHistory/TxHistory.tsx
jorbuedo: Not sure I understand. If usePoolTransitionModal got an error from usePoolTransition, isLoading should still go to false.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Following that in slack with them

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: why ?

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: for pools ok, for create tx, terrible.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Discussing the error screen in slack, but it will be an on demand function once that's clear.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: so why giving a text style a color name? should it be warningText, errorText?

apps/wallet-mobile/src/TxHistory/TxHistory.tsx
stackchain: what will happen to user if getPoolTrasitino fails?

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: yap, my point is the API, if a function is navigateToUpdate, when it's called it should navigate, it should be pure, it is relying on external context to decided whether is can/or/cannot navigate, this is the caller responsibility to check before calling it

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
jorbuedo: Well that depends on what I care about at any particular time.
If I want a high overview of the parent component with a list of child components I don't care about, then naming them is fine.
But if I'm writing, editing or trying to find those components based on the style they have, I'd prefer having them there instead of double hopping to find what they actually do.

I get wanting to establish a philosophy and sticking to it. But in this particular case it's single View (not reused) with a single style. If we're gonna Name every View in the app, that sounds a nightmare to navigate tbh, never knowing if those components do something or are just flavor. 

And that's ignoring the performance overhead on every usage of having to instantiate a component, destructure its props, call the styles hook, merge the props with the specific style.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Even if the PoolInfo returns the retiring ms, you still need an endpoint that tells you the new suggested one.
Unless you want every PoolInfo to also have a pointer to a suggested new PoolInfo.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Sure, already established in other threads that creating the tx should be on demand.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Because the parent component uses flex and gap to separate it's children. This is a single sentence divided in 2 Texts with different styles and one of them being optional. With a wrapper, both texts stay together. Without it, they'll be treated as separate children of the flex parent and be affected by the gap.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: I mean, other than being longer and having to figure out a name for them, no reason.
The design has in a single portion of the screen 7 different combinations of color and boldness. So just trying to be simple with the names.
I can totally append `Text` to all of them.

apps/wallet-mobile/src/TxHistory/TxHistory.tsx
jorbuedo: It's inside useQuery, so you'll get an error returned there. Since it's not used, then nothing out of the ordinary happens.

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
stackchain: Performance in these cases are just irrelevant. (there are libraries that do 10x more overhead to accomplish the same) I'm more concern about our sync issues over the renderings.
The first paragraph I didn't get it. 
About the philosophy, we are not restricting anything, but if a good soul already abstracted the Views aways, favouring readability, there is absolutely no reason to throw that away.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: This would be ideal, separation of concerns, and not a big soup.

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
jorbuedo: The reason I changed  is that this file was not converted for themes and was using `styles` directly. So this was the component:

```
const Actions = (props: ViewProps) => <View {...props} style={styles.actions} />
```

Since `styles` can't be accessed now without useStyles, that's why I removed the abstraction.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: it would be simpler to read if the Texts were contained by a Text, not by a View, to indicate a single line, if it was grabbing a ready text node from somewhere else it would make sense, but here it can be accomplished by just moving Text - inside Text.

apps/wallet-mobile/src/Dashboard/Dashboard.tsx
stackchain: yap, I got it, but I think we shouldn't drop it, my suggestion is too consume the styles here and keep as it was before `<Row..` and `<Actions`, we don't need to code everything this way, but since it's there, it would be nice to don't drop it.
later with styled-components it will be simpler or at least more "beautiful" to accomplish it.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: The new pool suggestion can be done in the same way of FFEE, S3 with cloud front on top, it will fill the needs

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: I don't think this will be useful at all, is like 10*5/10, what are they representing?

apps/wallet-mobile/src/TxHistory/TxHistory.tsx
stackchain: got it, so that's fine as long as it doesn't become a suspense.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: I assumed yoroi-lib getting the deadlines and suggested pool would be via an S3 file or similar, yes.

But right now yoroi-lib gets PoolInfo from `cexplorer.io`.

You're saying when we transform PoolInfo from the api, we add there the deadline and suggested pool from the S3 file there as well?

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: Sure, the wrapper can be another Text. I was thinking of html `<p>` which can't be nested.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
jorbuedo: image

Well they represent `text`, `textInsideCard`, `rowLabel`, `poolName`, `rowCurrentValue`, `rowSuggestedValue`, `warningText`, `warningTextTime`. I disagree on that being better, but I can change them all.

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: it looks like there is an API for the pool info coming from the aggregator, need to check with @pedromtcosta

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
stackchain: these are the places where the colors are implemented, my question is more, is there a word that represents what it is visually trying to indicate? maybe it's too many places then I agree, let's keep it, but if it's possible to identify what is representing red -> warning, gray -> labels, but if it's spread no worries

apps/wallet-mobile/src/features/Menu/Menu.tsx
jorbuedo: This was changed in another branch and now applied from develop

apps/wallet-mobile/src/Staking/PoolTransition/usePoolTransition.tsx
pedromtcosta: @stackchain not sure I understand what this is about, the aggregator is not ready yet 🤔

stackchain: @pedromtcosta can we get the retirement info from the current getPoolInfo ?

pedromtcosta: @stackchain we can return the epoch of the Pool retirement certificate, if that's what you are referring to. Should be an easy change. For reference: https://github.com/IntersectMBO/cardano-ledger/blob/8884d921c8c3c6e216a659fca46caf729282058b/eras/babbage/test-suite/cddl-files/babbage.cddl#L187

jorbuedo: I'll ask, but I'm not sure the onchain pool retirement is what management are looking for. It's a good information to have, specially for other pools. 
But they have a specific list, and another list to migrate to. And they may want to specify an earlier date than the actual retirement, so users have time to migrate without loosing rewards.

stackchain: @pedromtcosta @jorbuedo when creating the retirement certificate it can be created with an specific epoch ahead, so since is current_epoch + 1, it can be planned ahead, and API can return it, it would be nice to receive the retirement `epoch` and `date` (beginning of epoch) from the API so the FE doesn't need to do any math to get the countdown. Also @pedromtcosta the pool can be registered again, which means API need to handle retirement->registration->retirement->registration. (ty: @vsubhuman)

jorbuedo: That's good info to have, but doesn't solve the current request to have a list of retiring pools and a list of suggested pools to transition to. We may not even have access to the retiring pools to add the certificate.

stackchain: The list of retiring pools is no longer required with this implementation, whenever a pool the user is delegating is about to be retired the pop will show, the only thing needed is the suggestion (transition endpoint), in this scenario.

jorbuedo: We had a call, decided to go with current approach just for now, based on this API https://github.com/Emurgo/yoroi-lib/pull/136

Any future decisions on this, API improvements and support for other pools will be followed up separately.

jorbuedo:
@stackchain @michaeljscript @banklesss
Ready for merge on my side. 4.28.0 should get a regression to make sure the lib updates didn't break anything

stackchain (approved):
This is done but is blocked by the hardcoded date on the package, after merging a full regression will be required (swap create/cancel) due to CSL bump cc: @rahulnr7

stackchain:
> This is done but is blocked by the hardcoded date on the package, after merging a full regression will be required (swap create/cancel) due to CSL bump cc: @rahulnr7

Bringing it in, the package can be updated later.

PR #3195, opened 2024-04-14
chore(portfolio): base sync portfolio

stackchain:


.vscode/settings.json
```diff
@@ -1,4 +1,13 @@
 {
     "java.configuration.updateBuildConfiguration": "disabled",
-    "typescript.tsdk": "node_modules/typescript/lib"
+    "typescript.tsdk": "node_modules/typescript/lib",
+    "jestrunner.debugOptions": {
```
stackchain, line 4: jest parallelizes tests, but for that it uses JSON.stringify, which can't serialize bigint. it's needed to run-in-band, or with max-workers = 1.

apps/wallet-mobile/scripts/create-mocked-token-infos.js
stackchain, line 1: to create mocked records

apps/wallet-mobile/src/AppNavigator.tsx
```diff
@@ -248,9 +251,9 @@ type FirstAction = 'auth-with-pin' | 'auth-with-os' | 'request-new-pin' | 'first
 const getFirstAction = (
   isAuthOsSupported: boolean,
   authSetting: AuthSetting,
-  agreement: LegalAgreement | undefined,
+  legalAgreement: LegalAgreement | undefined | null,
```
stackchain, line 254: tiny refactor for consistency

apps/wallet-mobile/src/features/Portfolio/common/usePortfolioBalanceManager.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    balanceManager.hydrate()
+    return {
+      balanceManager,
+      balanceStorage,
+    }
+  }, [tokenManager, walletId])
+}
+
+const primaryTokenInfo = createPrimaryTokenInfo({
```
stackchain, line 64: temporary for testing purposes only

apps/wallet-mobile/src/features/Portfolio/useCases/PortfolioScreen.tsx
stackchain, line 1: testing purposes for now

apps/wallet-mobile/src/migrations/4_9_0.test.ts
```diff
@@ -30,7 +30,7 @@ describe('migrateAuthSetting', () => {
 
   // if the store is inconsistent we favor OS, so the user can disable on device and it will ask for a new pin
   it('old store is pin + os (inconsistent), method = "os"', async () => {
-    await rootStorage.join('appSettings/').multiSet([
+    await rootStorage.join('appSettings/').multiSet<string | boolean>([
```
stackchain, line 33: multi-set from now on requires the types

apps/wallet-mobile/src/yoroi-wallets/auth/auth.ts
```diff
@@ -159,9 +159,10 @@ export const disableAllEasyConfirmation = async (storage: App.Storage) => {
   })
 
   const updateWalletJSONs = walletIds.map(async (walletId) => {
-    const walletJSON: Record<string, unknown> & {isEasyConfirmationEnabled: boolean} = await walletStorage
+    const walletJSON: null | (Record<string, unknown> & {isEasyConfirmationEnabled: boolean}) = await walletStorage
```
stackchain, line 162: anything read from storage can be `null` eventually.

packages/common/package.json
```diff
@@ -58,8 +58,8 @@
     "prepublish:prod": "yarn build:release",
     "publish:prod": "npm publish --scope yoroi --access public",
     "release": "release-it",
-    "test": "jest",
-    "test:watch": "jest --watch --debug",
+    "test": "jest --maxWorkers=1",
```
stackchain, line 61: needed for bigint

packages/common/src/api/fetchData.ts
```diff
@@ -1,6 +1,13 @@
 import axios, {AxiosError, AxiosRequestConfig, AxiosResponse} from 'axios'
 import {Api} from '@yoroi/types'
 
+// Module augmentation to extend AxiosRequestConfig
+declare module 'axios' {
+  export interface AxiosRequestConfig {
+    metadata?: {startTime: number; duration: number}
```
stackchain, line 7: tracks request execution time

packages/common/src/cache/cache-record-schema-maker.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const cacheRecordSchemaMaker = <T>(
+  recordSchema: ZodType<T>,
+): ZodType<App.CacheRecord<T>> => {
+  return z.object({
+    record: recordSchema,
+    expires: z.number().nonnegative(),
+    hash: z.string(),
+    // Zod is not infering the type correctly when using interface
+  }) as unknown as ZodType<App.CacheRecord<T>>
```
stackchain, line 12: zod loses reference when inferring with another schema as part of the current schema

packages/common/src/numbers/as-atomic-value.test.ts
```diff
@@ -0,0 +1,36 @@
+import {Numbers} from '@yoroi/types'
+
+import {asAtomicValue} from './as-atomic-value'
+
+describe('asQuantity', () => {
```
stackchain:
```suggestion
describe('asAtomicValue', () => {
```

packages/common/src/numbers/as-atomic-value.test.ts
stackchain, line 1: still considering it

packages/common/src/queue/queue-task.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  it('should destroy the queue and stop executing tasks', async () => {
+    const queue = queueTaskMaker()
+
+    const task1 = jest.fn().mockResolvedValue(undefined)
+    const task2 = jest.fn().mockResolvedValue(undefined)
+
+    queue.destroy()
+
+    queue.enqueue(task1)
```
stackchain, line 30: current tasks are always processed, rxjs definition.

packages/common/src/storage/adapters/async-storage.test.ts
stackchain, line 1: with this change when switching branches, if the packages are not built it will display 0 wallets, cuz `Object/` will be the path instead of `/` (while not merged/rebased after merged)

packages/common/src/storage/adapters/async-storage.ts
```diff
@@ -102,7 +107,7 @@ export const mountAsyncMultiStorage = <T = unknown>(
     dataFolder,
     keyExtractor,
     serializer = JSON.stringify,
-    deserializer = parseSafe as (item: string | null) => Nullable<T>,
+    deserializer = parseSafe as (item: string | null) => T | null,
```
stackchain, line 110: TS is dumb, it doesn't expand Generics, so Generics become opaque, and any null/undefined dropped by asserts won't work with Generics, sad.

packages/common/src/storage/helpers/storage-deserializer-maker.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * Creates a storage deserializer function based on the provided mapping.
+ * Since BigInt and BigNumber are not supported by JSON.parse, it stores as strings and revive as BigInt or BigNumber.
+ *
+ * @param mapping - The mapping of keys to reviver types.
+ * @returns JSON object revived according to the mapping, string -> BigInt or BigNumber.
+ */
+export const storageDeserializerMaker = (mapping: StorageReviverMapping) => {
```
stackchain, line 20: serializing is simple, by finding BigInt/BigNumber properties, for deserializing a map is needed, it doesn't take paths, so if `x` is a mapped, any `x` level will be considered whether BI/BN, therefore bigint names **must** be unique. eg. `a.b.c.x` x = BN/BI, also for arrays it is needed to embed in a object (it can serlialize, but not deserialize)

packages/common/src/storage/helpers/storage-serializer.ts
```diff
@@ -0,0 +1,19 @@
+import BigNumber from 'bignumber.js'
+
+/**
+ * Serializes the given object into a JSON string, with support for BigInt and BigNumber values.
+ * Since BigInt and BigNumber are not supported by JSON.stringify, it serialize them as strings.
+ *
+ * @param toStringify - The object to be serialized.
+ * @returns A stringfied JSON object with BigInt and BigNumber values serialized as strings.
+ */
+export const storageSerializer = (toStringify: unknown) => {
```
stackchain, line 10: wrap, JSON.stringify(data, storageSerializer) won't work, it needs to be used directly. storageSerializer(data)

packages/common/src/utils/arrays.test.ts
stackchain, line 1: helpers = app context 
utils = generics / library
therefore move as utils.

packages/common/src/utils/predicates.ts
stackchain, line 1: considering overload

packages/portfolio/src/adapters/dullahan-api/transformers.ts
```diff
@@ -0,0 +1,50 @@
...
+  apiTokenInfosResponse: DullahanApiTokenInfosResponse,
+) => {
+  const apiCachedTokenInfos: Record<
+    Portfolio.Token.Id,
+    Api.ResponseWithCache<Portfolio.Token.Info>
+  > = {}
+
+  return Object.entries(apiTokenInfosResponse).reduce(
+    (acc, [id, tokenInfoWithCache]) => {
+      const castedId = id as Portfolio.Token.Id
```
stackchain, line 23: casting is safe here, lost by reduce.

packages/portfolio/src/adapters/dullahan-api/types.ts
```diff
@@ -0,0 +1,7 @@
+import {Api, Portfolio} from '@yoroi/types'
+
+export type DullahanApiTokenInfosResponse = Readonly<{
+  [key: Portfolio.Token.Id]: Api.ResponseWithCache<
+    Omit<Portfolio.Token.Info, 'nature'>
```
stackchain, line 5: from API it receives only secondary, primary is static local.

packages/portfolio/src/adapters/mmkv-storage/balance-storage-maker.test.ts
```diff
@@ -0,0 +1,213 @@
+import {storageSerializer} from '@yoroi/common'
+import {App, Portfolio} from '@yoroi/types'
+
+import {portfolioBalanceStorageMaker} from './balance-storage-maker' // Adjust the path accordingly
```
stackchain:
```suggestion
import {portfolioBalanceStorageMaker} from './balance-storage-maker'
```

packages/portfolio/src/balance-manager.ts
```diff
@@ -0,0 +1,331 @@
...
+  })
+  const updatePrimary = primaryTokenUpdaterWrapper({
+    primaryTokenInfo: primaryToken.info,
+    storagePrimaryBalanceBreakdown: storage.primaryBalanceBreakdown,
+  })
+
+  const subscription = tokenManager.observable
+    .pipe(
+      filter(() => isHydrated),
+      filter((dtoEvent) => isNotTriggeredBySelf(sourceId)(dtoEvent)),
```
stackchain, line 70: ignores events started by itself

packages/portfolio/src/balance-manager.ts
```diff
@@ -0,0 +1,331 @@
...
+      filter(
+        (dtoEvent) =>
+          isEventTokenManagerSync(dtoEvent) &&
+          hasStaleTokenInfo(secondaries)(dtoEvent),
```
stackchain, line 74: it should also check the hash, yet, since it doesn't load 2 wallets at same time is safe for now.

packages/portfolio/src/balance-manager.ts
```diff
@@ -0,0 +1,331 @@
...
+        info: primaryTokenInfo,
+      },
+      true,
+    )
+    storagePrimaryBalanceBreakdown.clear()
+    storagePrimaryBalanceBreakdown.save(newPrimaryBreakdown)
+    return freeze({newPrimaryBreakdown, newPrimaryTokenBalance}, true)
+  }
+
+// TODO list
```
stackchain, line 321: ignore it for now

packages/portfolio/src/helpers/create-cached-unknown-token-info.ts
```diff
@@ -0,0 +1,16 @@
+import {App, Portfolio} from '@yoroi/types'
+import {freeze} from 'immer'
+
+import {createUnknownTokenInfo} from './create-unknown-token-info'
+
+export const createCachedUnknownTokenInfo = (
```
stackchain, line 6: api fails won't block the wallet anymore, tokens will be created as unknown and will be stale by default, so it will refresh them every chance until get something from API.

packages/portfolio/src/helpers/is-nft.ts
```diff
@@ -0,0 +1,4 @@
+import {Portfolio} from '@yoroi/types'
+
+export const isNft = (tokenInfo: Portfolio.Token.Info): boolean =>
```
stackchain, line 3: there is not relation with the test that should happen in BE.

stackchain: filtering purposes only.

packages/types/src/portfolio/price.ts
stackchain, line 1: skip for now

packages/portfolio/src/transformers/deserializers.ts
```diff
@@ -0,0 +1,62 @@
...
+  quantity: StorageReviverType.AsBigInt,
+}
+
+const tokenBalanceDeserializer = (jsonString: string | null) => {
+  if (jsonString == null) return null
+  const record = storageDeserializerMaker(tokenBalanceReviverMapping)(
+    jsonString,
+  )
+  const parsed = parseTokenBalance(record)
+  return parsed ?? null
```
stackchain, line 34: from storage expect null not undefined

apps/wallet-mobile/scripts/create-mocked-token-infos.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+const generatedTokens = generateTokenInfos(nftCryptoKitty, 50)
+
+const apiResponseTokenInfos = generatedTokens.reduce((acc, token, index) => {
+  acc[token.id] = [200, token, `hash${index + 1}`, 3600]
+  return acc
+}, {})
+
+console.log(JSON.stringify(apiResponseTokenInfos, null, 2))
```
michalszorad: Log not needed

apps/wallet-mobile/scripts/create-mocked-token-infos.js
```diff
@@ -0,0 +1,42 @@
...
+  const tokens = []
+  for (let i = 0; i < count; i++) {
+    // Clone the base token object to create a new token
+    const newToken = {...baseToken}
+
+    // Modify properties to make the token unique
+    newToken.id = `${baseToken.id.split('.')[0]}.${Buffer.from(String(i)).toString('hex')}`
+    newToken.name = `${baseToken.name.split(' ')[0]} #${i}`
+    newToken.reference = `${baseToken.reference.split('.')[0]}.${Buffer.from(String(i)).toString('hex')}`
+    newToken.fingerprint = `asset${i}s7nlt45cc82upqewvjtgu7g97l7eg483c6wu${i}`
```
michalszorad:
```suggestion
    const id = `${baseToken.id.split('.')[0]}.${Buffer.from(String(i)).toString('hex')}`
    const name = `${baseToken.name.split(' ')[0]} #${i}`
    const reference = `${baseToken.reference.split('.')[0]}.${Buffer.from(String(i)).toString('hex')}`
    const fingerprint = `asset${i}s7nlt45cc82upqewvjtgu7g97l7eg483c6wu${i}`
    
    const newToken = {...baseToken, id, name, reference, fingerprint}
```

apps/wallet-mobile/src/features/Portfolio/common/usePortfolioBalanceManager.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const usePortfolioBalanceManager = ({
+  tokenManager,
+  walletId,
+}: {
+  tokenManager: Portfolio.Manager.Token
+  walletId: YoroiWallet['id']
+}) => {
+  return React.useMemo(() => {
+    const balanceStorageMounted = mountMMKVStorage<Portfolio.Token.Id>({path: `balance/${walletId}/`})
```
michalszorad, line 16: Previously we had the path `${walletId}/feature/`, is it supposed to be the other way here?

apps/wallet-mobile/src/yoroi-wallets/auth/auth.ts
```diff
@@ -264,7 +265,7 @@ const messages = defineMessages({
   },
 })
 
-export type AuthSetting = 'pin' | 'os' | undefined
+export type AuthSetting = 'pin' | 'os' | undefined | null
```
michalszorad: The `LegalAgreement` has a defined type without `null | undefined` and the `getFirstAction` uses `legalAgreement: LegalAgreement | undefined | null` so I am wondering if the similar approach would be better here.

Eg 
```
export type AuthSetting = 'pin' | 'os'
```
in getFirstAction
```
authSetting: AuthSetting | undefined | null

File: apps/wallet-mobile/src/yoroi-wallets/cardano/transactionManager/transactionManager.test.ts [hunk 27]
```diff
@@ -32,7 +33,7 @@ describe('transactionManager', () => {
     DeviceInfo.getVersion = () => '9.9.9'
 
     const mockStorage = rootStorage.join('txs/')
-    mockStorage.multiSet([
+    mockStorage.multiSet<Transaction | [string]>([
```
michalszorad: should be there `await` before `mockStorage.multiSet`

apps/wallet-mobile/src/yoroi-wallets/memos/memosManager.ts
```diff
@@ -33,7 +33,7 @@ export const makeMemosManager = async (storage: App.Storage) => {
   } as const
 }
 
-const filterCorruptEntries = (tuples: ReadonlyArray<[string, string | undefined]>) => {
+const filterCorruptEntries = (tuples: ReadonlyArray<[string, string | undefined | null]>) => {
```
michalszorad:
```suggestion
const filterCorruptEntries = (tuples: ReadonlyArray<[string, unknown]>) => {
```

packages/common/package.json
michalszorad: Why is there a limitation?

packages/common/src/api/fetchData.ts
```diff
@@ -40,27 +47,27 @@ export type FetchData = <T, D = any>(
  *
  * @example
  * ```typescript
- * // Example of a GET request
+ * //// Example of a GET request
```
michalszorad:
```suggestion
 * // Example of a GET request
```

packages/common/src/cache/cache-manage-multi-request.ts
```diff
@@ -0,0 +1,136 @@
+import {Api, App} from '@yoroi/types'
+import {freeze} from 'immer'
+
+import {cacheResolveRecordsSource} from './cache-resolve-records-source'
+import {isRight} from '../utils/monads'
+import {hasEntryValue} from '../utils/predicates'
+
+export const cacheManageMultiRequest = async <K extends string, V>({
```
michalszorad: For readability 
```tsx
type CacheManagerMultiRequestParams = ...
export const cacheManageMultiRequest = async <K extends string, V>({} : CacheManagerMultiRequestParams)

File: packages/common/src/cache/cache-manage-multi-request.ts [hunk 31]
```diff
@@ -0,0 +1,136 @@
...
+
+  let toSaveNewFromApi: Map<K, App.CacheRecord<V>> = new Map()
+  let recordsFromApi: Record<K, Api.ResponseWithCache<V>> = {} as Record<
+    K,
+    Api.ResponseWithCache<V>
+  >
+
+  if (toFetch.length > 0) {
+    const apiResponse = await request(toFetch)
+    // if the request fails, we will show unknown tokens and wont block the user
```
michalszorad, line 59: If we have 5 requests and 1 of them fails, does it act like all of them failed?

packages/common/src/cache/cache-manage-multi-request.ts
michalszorad: What are the benefits of cached multi requests manager instead of having a 1 request = 1 cache manager

packages/common/src/cache/cache-record-schema-maker.ts
michalszorad: Is it possible to use `z.object<MyType>` to avoid type casting?

packages/common/src/cache/cache-record-schema-maker.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+export const isCacheRecord = <T>(
```
michalszorad, line 15: `createTypeGuardFromSchema` could be reused? By using it we will avoid writing the `data is App.CacheRecord<T>` manually which should avoid mistakes in the implementation

packages/common/src/cache/extract-entry-cache-info.ts
```diff
@@ -0,0 +1,9 @@
+import {App} from '@yoroi/types'
+
+export const extractEntryCacheInfo = <
```
michalszorad, line 3: What is the benefit of having this function? It seems to only convert `undefined` to `null`

packages/common/src/observer/observer.ts
```diff
@@ -1,23 +1,23 @@
 import {App} from '@yoroi/types'
+import {freeze} from 'immer'
+import {Subject, Subscription} from 'rxjs'
```
michalszorad, line 3: 🙌

packages/common/src/storage/adapters/mmkv-storage.ts
```diff
@@ -1,17 +1,22 @@
...
-): App.Storage<false> => {
+export const mountMMKVStorage = <Key extends string = string>(
+  {
+    path,
+    id = 'default.mmkv',
+  }: {
+    path: App.StorageFolderName
+    id?: string
+  },
+  {instance}: {instance?: MMKV} = {},
```
michalszorad, line 18: Could be `instance` in the 1st argument as well?

packages/common/src/storage/helpers/storage-deserializer-maker.ts
```diff
@@ -0,0 +1,57 @@
...
+    switch (mapping[key]) {
+      case StorageReviverType.AsBigInt:
+        return value == null ? value : BigInt(value)
+      case StorageReviverType.AsBigNumber:
+      default:
+        return value === null ? value : new BigNumber(value)
+    }
+  }
+
+  const convertProperties = (obj: any): unknown => {
```
michalszorad: it may be safer to use `unknown` instead of `any`

packages/common/src/storage/helpers/storage-deserializer-maker.ts
```diff
@@ -0,0 +1,57 @@
...
+    if (Array.isArray(obj)) {
+      return obj.map((item) => convertProperties(item))
+    } else if (obj !== null && typeof obj === 'object') {
```
michalszorad: `else` not needed if you have `return` above

packages/common/src/storage/helpers/storage-deserializer-maker.ts
michalszorad: How does it convert the properties?

packages/common/src/storage/helpers/storage-deserializer-maker.ts
```diff
@@ -0,0 +1,57 @@
...
+      Object.keys(obj).forEach((key) => {
+        obj[key] = convertProperties(obj[key])
```
michalszorad, line 36: Modifies original object

packages/common/src/utils/predicates.ts
```diff
@@ -0,0 +1,6 @@
+export const hasValue = <V>(value: V | null | undefined): value is V =>
```
michalszorad, line 1: 🎉

michalszorad: I imagine this could be reused in some of the files where `!= null` is used

packages/portfolio/src/adapters/dullahan-api/api-maker.ts
```diff
@@ -0,0 +1,106 @@
...
+          ReadonlyArray<Api.RequestWithCache<Portfolio.Token.Id>>
+        >({
+          method: 'post',
+          url: config.tokenInfos,
+          data: idsWithCache,
+          headers: {
+            'Content-Type': 'application/json',
+            'Accept': 'application/json',
+          },
+        }).then((response) => {
```
michalszorad: async/await instead of .then

packages/portfolio/src/adapters/dullahan-api/transformers.ts
```diff
@@ -0,0 +1,50 @@
...
+      if (!parseSecondaryTokenInfoWithCacheRecord(castedTokenInfoWithCache))
+        throw new Api.Errors.ResponseMalformed(
+          'Unable to parse the token-info response',
+        )
+
+      const [statusCode] = castedTokenInfoWithCache
+
+      if (statusCode === Api.HttpStatusCode.NotModified) {
+        acc[castedId] = castedTokenInfoWithCache
+        return acc
```
michalszorad: It is a new object, so nothing should break, but would always suggest to use spread in these cases

```suggestion
        return {...acc, [castedId]: castedTokenInfoWithCache}
```

packages/portfolio/src/adapters/dullahan-api/types.ts
michalszorad: What is this `DullahanApi`?

packages/portfolio/src/adapters/mmkv-storage/token-storage-maker.test.ts
```diff
@@ -0,0 +1,282 @@
+import {cacheRecordMaker, storageSerializer} from '@yoroi/common'
+import {App, Portfolio} from '@yoroi/types'
+
+import {portfolioTokenStorageMaker} from './token-storage-maker' // Adjust the path accordingly
```
michalszorad, line 4:
```suggestion
import {portfolioTokenStorageMaker} from './token-storage-maker' //TODO: Adjust the path accordingly
```

packages/portfolio/src/validators/amount.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import {TokenIdSchema} from './token-id'
+
+export const AmountSchema = z.object({
+  id: TokenIdSchema,
+  quantity: z.bigint(),
+})
+
+export const isAmount = (data: unknown): data is Portfolio.Amount => {
+  return AmountSchema.safeParse(data).success
```
michalszorad, line 12: `createTypeGuardFromSchema`

packages/portfolio/src/validators/token-balance.ts
```diff
@@ -0,0 +1,22 @@
...
+export const TokenBalanceSchema = z.object({
+  info: TokenInfoSchema,
+  balance: z.bigint(),
+  lockedInBuiltTxs: z.bigint(),
+})
+
+export const isTokenBalance = (
+  data: unknown,
+): data is Portfolio.Token.Balance => {
+  return TokenBalanceSchema.safeParse(data).success
```
michalszorad, line 15: `createTypeGuardFromSchema`

packages/portfolio/src/helpers/sorting.test.ts
```diff
@@ -0,0 +1,160 @@
...
+import {tokenInfoMocks} from '../adapters/token-info.mocks'
+import {
+  alpha,
+  sortTokenBalances,
+  sortTokenInfos,
+  toEnd,
+  toStart,
+} from './sorting'
+
+describe('sorting', () => {
```
michalszorad, line 14: We have the same test in `apps/wallet-mobile/src/utils/sorting.test.ts` should it be removed from wallet-mobile?

packages/portfolio/src/helpers/sorting.ts
```diff
@@ -0,0 +1,104 @@
+import {Portfolio} from '@yoroi/types'
+
+export const alpha =
```
michalszorad, line 3: Should this be removed from wallet-mobile?

packages/portfolio/src/validators/primary-balance-breakdown.ts
```diff
@@ -0,0 +1,26 @@
...
+      redeemableAfter: z.number(),
+      source: z.string(),
+    }),
+  ),
+})
+
+export const isPrimaryBalanceBreakdown = (
+  data: unknown,
+): data is Portfolio.BalancePrimaryBreakdown =>
+  PrimaryBalanceBreakdownSchema.safeParse(data).success
```
michalszorad, line 20: `createTypeGuardFromSchema`

packages/types/src/api/cache.ts
```diff
@@ -0,0 +1,7 @@
+import {ApiHttpStatusCode} from './status-code'
+
+export type ApiResponseRecordWithCache<T> =
+  | [StatusCode: ApiHttpStatusCode.Ok, Record: T, ETag: string, MaxAge: number]
+  | [StatusCode: ApiHttpStatusCode.NotModified, MaxAge: number]
```
michalszorad, line 5: I'm thinking this could be an object for easier usage

packages/types/src/portfolio/info.ts
```diff
@@ -0,0 +1,41 @@
+import {
+  PortfolioTokenApplication,
+  PortfolioTokenId,
+  PortfolioTokenNature,
+  PortfolioTokenStatus,
+  PortfolioTokenType,
+} from './token'
+
+type CommonTokenInfo = {
```
michalszorad, line 9: Will this be later reused instead of `Balance.TokenInfo` which we use in utils such as `useTokenInfo`?

packages/portfolio/src/validators/token-type.ts
```diff
@@ -0,0 +1,14 @@
+import {Portfolio} from '@yoroi/types'
+import {z} from 'zod'
+
+export const TokenTypeSchema = z.nativeEnum(Portfolio.Token.Type)
+
+export const isTokenType = (data: unknown): data is Portfolio.Token.Type => {
+  return TokenTypeSchema.safeParse(data).success
```
michalszorad: `createTypeGuardFromSchema`

packages/portfolio/src/validators/token-property-type.ts
```diff
@@ -0,0 +1,18 @@
...
+import {z} from 'zod'
+
+export const TokenPropertyTypeSchema = z.nativeEnum(
+  Portfolio.Token.PropertyType,
+)
+
+export const isTokenPropertyType = (
+  data: unknown,
+): data is Portfolio.Token.PropertyType => {
+  return TokenPropertyTypeSchema.safeParse(data).success
```
michalszorad: `createTypeGuardFromSchema` on other places as well

packages/types/src/api/cache.ts
stackchain: the reason is that on mobile it will transfer "keys", like 200 "status", which is not necessary, so that's why using a positional array. so we save users' bandwidth with metadata

packages/common/package.json
stackchain: the way the jest was built to parallelize tests https://github.com/jestjs/jest/issues/11617

packages/common/src/cache/cache-record-schema-maker.ts
stackchain: checking

packages/portfolio/src/adapters/dullahan-api/types.ts
stackchain: new headless engine (WIP yet) https://github.com/Emurgo/dullahan

apps/wallet-mobile/scripts/create-mocked-token-infos.js
stackchain:
```suggestion
```

apps/wallet-mobile/src/yoroi-wallets/auth/auth.ts
stackchain: let me check

apps/wallet-mobile/src/yoroi-wallets/cardano/transactionManager/transactionManager.test.ts
stackchain: I think this would explain some random errors while running the pipelines

packages/common/src/api/fetchData.ts
stackchain: let me check but I think it needs to be double for the hover desc

packages/common/src/cache/cache-manage-multi-request.ts
stackchain: 4k items, 4k network round-trips. multi, it can batch, and figure out the best batch size, based on the network conditions.

packages/common/src/cache/cache-manage-multi-request.ts
stackchain: yap, that's precise, cuz it's hitting once now, next iteration is to batch it, 100/250/500 still unknown, this will mark all those tokens as unknown, but they will be stale right away, so next sync will try fetching them again, and also it can implement retry logic (at least 3 times) before building as unknown.

stackchain: like it is on tx-history right now.

packages/common/src/cache/cache-record-schema-maker.ts
stackchain: I remember I tested a lot of stuff couldn't get it working 
```
Type 'ZodType<CacheRecord<T>, ZodTypeDef, CacheRecord<T>>' does not satisfy the constraint 'ZodRawShape'.
  Index signature for type 'string' is missing in type 'ZodType<CacheRecord<T>, ZodTypeDef, CacheRecord<T>>'
```

stackchain: the biggest issue on Zod is template literals, there is a PR coming to fix it, it looks it will be ready on next release.

apps/wallet-mobile/src/yoroi-wallets/auth/auth.ts
stackchain: aaa, the return type of the deserializers always must have T | `null` to don't break storage check, cuz empty records are expected to respond as `null`

stackchain: fixed, dropped the null on the getter fn

apps/wallet-mobile/src/features/Portfolio/common/usePortfolioBalanceManager.ts
stackchain: @michaeljscript portfolio will work differently (it will require a different remove either), the idea is to the dashboard to aggregate all your assets value later, (per network), so it's not final still work in progress. I'm guessing the final version will look like `{chainId}/{balances}/{walletId}`

stackchain: or maybe it can stick with /wallets/{ids}/.. but then it will require a discovery process per chain id. (as I said WIP)

apps/wallet-mobile/src/yoroi-wallets/cardano/transactionManager/transactionManager.test.ts
stackchain: fixed

packages/common/src/cache/extract-entry-cache-info.ts
stackchain: it transforms/extracts the api cache info `[,tokeninfo,etag,maxage]` -> to the storage cache record -> `{expires: <when>, hash: <whatever, in this case eTag from api>}`, deals also with not-found items

packages/common/src/cache/cache-record-schema-maker.ts
stackchain: I think for the other I can leverage it for this no, cuz it wraps within

packages/common/src/storage/adapters/mmkv-storage.ts
stackchain: Think of it as `deps` to inject, so you won't use it as a client, unless if testing it.

packages/common/src/storage/helpers/storage-deserializer-maker.ts
stackchain: looking bad, let me refactor it.

packages/common/src/storage/helpers/storage-deserializer-maker.ts
stackchain: this is ok, this object is created right to process it.

packages/common/src/storage/helpers/storage-deserializer-maker.ts
stackchain: fixed, leveraged isRecord.

packages/common/src/utils/predicates.ts
stackchain: it makes sense, in this case I won't overload with [, null] -> hasEntryValue.

packages/types/src/portfolio/info.ts
stackchain: Balance** will be completely removed, for some time we will have both, until transition everything into Portfolio.

packages/portfolio/src/adapters/dullahan-api/api-maker.ts
stackchain: refacoted for consistency

packages/portfolio/src/adapters/mmkv-storage/token-storage-maker.test.ts
stackchain: :') as I mentioned above, still not sure what'd be the best approach to aggregate the entire balance per network.

packages/common/src/storage/helpers/storage-deserializer-maker.ts
stackchain: to deserialize a map is needed, so there is a factory to build the maps (which are based on keys only and not paths`.a.b.c` not supported), and positional/non-positional arrays are not supported for now.

packages/portfolio/src/validators/amount.ts
stackchain: 🤦🏻 updating all validators

stackchain: ok, now I remembered, template literals issues, I can update everything that doesn't contain a template literal (aka. Token.Id)

stackchain: we will be able to leverage it only after this is merged: https://github.com/colinhacks/zod/pull/1786

packages/common/src/utils/predicates.ts
stackchain: not anymore (using in other places)

packages/common/src/cache/cache-record-schema-maker.ts
stackchain: yap, only after this probably https://github.com/colinhacks/zod/pull/1786

.vscode/settings.json
stackchain: -> https://github.com/jestjs/jest/issues/11617

packages/common/src/api/fetchData.ts
stackchain: fixed

packages/common/src/cache/cache-manage-multi-request.ts
stackchain: refactored for readability

packages/portfolio/src/helpers/sorting.test.ts
stackchain: it will after merging and adopting it, cuz after merging it, the balances will be always sorted and ready to go syncrously from package.

stackchain: it was added few more fucntionalies was well, like to cover for unknowns, scams, etc...

packages/portfolio/src/helpers/sorting.ts
stackchain: yap, after full adoption.

packages/portfolio/src/validators/token-type.ts
stackchain: I've updated the validators that are not using template literals

michalszorad (approved):

PR #3196, opened 2024-04-15
refactor(theme): themes flexibility 

stackchain:
- [x] Covered with 100% tests
- [x] Renamed to use Tailwind CSS selectors (1 update is needed on Figma)
- [x] Added storage (selecting will persist)
- [x] Added detectTheme - will resolve from 'system' if empty and 'sytem' now is also a valid option
- [x] Refactored tokens (not liking the current definition) specially lineHeight
- [x] Renamed from `-` to `_` so it can be accessed 
- [x] Added tiny README.md
- [x] `useTheme` is designed to manage theme selection not to be consumed by the UI
- [x] `useThemeColor` is designed to consume colors
- [x] Moved from calculated to static 
- [x] Split Palette into Base (dark/light) and Theme (exposed officially used)

It is using `as const` therefore making the properties read-only I'm guessing react-native will complain, it can leverage `Writable` or Draft from 'immer'


stackchain:
@SorinC6 can you take it from here? It will require an update on wallet-mobile, big one.
- [ ] Fix initialization (a storage is required, check Portfolio for example)
- [ ] ~Fix consume from useTheme -> useThemeColor~
- [ ] ~Fix styles (need to import atoms) to everything that is not theme (color) based~
- [ ] Fix styles colors
---
It could also export from Theme the atoms, yet, IMHO is an overkill, unless is planned a `compact` / `big` version.

stackchain:
@SorinC6 please check with Design, if they approve the Base->Theme palette naming convention we can go on with the implementation

nistadev (approved):

PR #3213, opened 2024-04-23
fix(wallet-mobile): NetworkError instance

jorbuedo:
related: https://emurgo.atlassian.net/browse/YOMO-1356

This one was confusing. Not sure why we have `NetworkError` exported from `@yoroi/common`, but it's not the same one the wallet is throwing, so don't mix them up because then `instanceof` will fail.


banklesss:
maybe it's worth to still use the common package error but catching it by instance and use the UI error messages?

jorbuedo:
> maybe it's worth to still use the common package error but catching it by instance and use the UI error messages?

Well, sure, we can use the one in common everywhere instead. None of the uses I see use the message from the error itself.

apps/wallet-mobile/src/features/SetupWallet/useCases/CreateWallet/WalletDetailsScreen.tsx
```diff
@@ -233,7 +233,7 @@ export const WalletDetailsScreen = () => {
   ])
 
   React.useEffect(() => {
-    if (showRestoreWalletInfoModal) showModalTipsPassword()
+    if (showRestoreWalletInfoModal === true) showModalTipsPassword()
```
stackchain:
```suggestion
    if (showRestoreWalletInfoModal) showModalTipsPassword()
```

stackchain:
> > maybe it's worth to still use the common package error but catching it by instance and use the UI error messages?
> 
> Well, sure, we can use the one in common everywhere instead. None of the uses I see use the message from the error itself.

The network error inside wallet is legacy, there are 2 fetchers, the legacy fetch, and the fetchData, which is the new one. There will be few places to replace. Please update them all to use -> from types Api.Error.Network, from types, the errors inside common can be dropped.

apps/wallet-mobile/src/features/SetupWallet/useCases/CreateWallet/WalletDetailsScreen.tsx
jorbuedo: Not sure why this was complaining for me, saying `showRestoreWalletInfoModal` was `any` and required a specific check, despite it being a boolean.

stackchain: I have weird messages on it too sometimes with VSCode + eslint, then I close (like windows) I close VSCode and open it again, it stops complaining, it looks this rule for boolean has come conflict with VSC or something is not working between them.

stackchain (approved):

PR #3215, opened 2024-04-23
fix(wallet-setup): empty wallet list state

banklesss:
Related to:

- [YOMO-1301](https://emurgo.atlassian.net/browse/YOMO-1301)

[YOMO-1301]: https://emurgo.atlassian.net/browse/YOMO-1301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/wallet-mobile/src/features/Menu/Menu.tsx
```diff
@@ -187,7 +187,6 @@ const useNavigateTo = () => {
   const prefetchStakingInfo = usePrefetchStakingInfo(wallet)
 
   return {
-    allWallets: () => navigation.navigate('app-root', {screen: 'wallet-selection'}),
```
banklesss, line 190: not used anymore in the revamp design

apps/wallet-mobile/src/navigation.tsx
```diff
@@ -514,6 +542,28 @@ export const useWalletNavigation = () => {
...
+    },
+  })
+
+  const navigateToWalletSelectionSync = React.useCallback(() => {
+    refetch()
+  }, [refetch])
+
+  return navigateToWalletSelectionSync
+}
+
```
banklesss, line 566: To determine if the user has wallets, it is necessary to use the `listWallets` method.

When deleting a wallet, its metadata is not deleted at that time, but rather when the app is launched. When using the `listWallets` method, a check is performed to filter the IDs of the wallets that are going to be deleted.

In short, it is necessary to perform a `refetch` to obtain an updated value of `hasWallets`, that is needed to determine if we are going to show the wallets list screen or the setup wallet screen

michalszorad (approved):

apps/wallet-mobile/src/navigation.tsx
```diff
@@ -514,6 +536,24 @@ export const useWalletNavigation = () => {
...
+        return
+      }
+
+      resetToWalletSetup()
+    },
+  })
+
+  return refetch
+}
+
```
banklesss: To determine if the user has wallets, it is necessary to use the listWallets method.

When deleting a wallet, its metadata is not deleted at that time, but rather when the app is launched. When using the listWallets method, a check is performed to filter the IDs of the wallets that are going to be deleted.

In short, it is necessary to perform a refetch to obtain an updated value of hasWallets, that is needed to determine if we are going to show the wallets list screen or the setup wallet screen when a wallet is removed.

apps/wallet-mobile/src/features/WalletManager/SelectWalletFromList/SelectWalletFromListScreen.tsx
```diff
@@ -146,20 +145,16 @@ const SupportTicketLink = () => {
...
   const strings = useStrings()
   const {styles} = useStyles()
-  const {walletImplementationIdChanged, reset: resetSetupWallet} = useSetupWallet()
+  const {reset: resetSetupWallet} = useSetupWallet()
+  const {resetToWalletSetup} = useWalletNavigation()
 
   return (
     <Button
       onPress={() => {
         resetSetupWallet()
```
stackchain: I'm guessing resetSetupWallet -> re-initialize state and resetToWalletSetup navigates to it? may worth adding a comment on top

apps/wallet-mobile/src/navigation.tsx
stackchain: Roger, but this will be way simpler to achieve with the changes in the walletManager, (we can't pick right now), create a ticket for 4.28 to refactor it

stackchain: Question invalidating the query cache didn't do the job? It should, can listWallets mind for deleted ones.

banklesss: i tested this but it doesn't work. I'm doing it wrong?

```
export const useNavigateToWalletSelectionSync = () => {
  const {resetToWalletSetup, resetToWalletSelection} = useWalletNavigation()
  const walletManager = useWalletManager()
  const queryClient = useQueryClient()

  const {hasWallets} = useHasWallets(walletManager, {
    onSuccess: () => queryClient.invalidateQueries({queryKey: ['hasWallets']}),
  })

  if (hasWallets) {
    return resetToWalletSelection
  }

  return resetToWalletSetup
}
```

banklesss: i fixed the queryKey [here](https://github.com/Emurgo/yoroi/pull/3215/commits/6f0aec71c20595e301fc9f52be693581ddd4e263#diff-74c3148fcce592101f4de41e6d2c636df4d4c9145956b49662585dbece349b40L703) but it doesn't work yet : (

stackchain: I think a helper would do the work
```
  export const whetherHasWallets = (
    walletManager: WalletManager,
    withWallets: () => void,
    withoutWallets: () => void
  ) => {
    const hasWallets = walletManager.listWallets().length > 0

    if (hasWallets) return withWallets()

    return withoutWallets()
}
```

banklesss: FIXED.

The problem was in using incorrect queryId to invalidate.

apps/wallet-mobile/src/yoroi-wallets/hooks/index.ts
```diff
@@ -700,7 +700,7 @@ export const useHasWallets = (
   options?: UseQueryOptions<Array<WalletMeta>, Error, boolean>,
 ) => {
   const query = useQuery({
-    queryKey: ['walletMetas'],
+    queryKey: ['hasWallets'],
```
michalszorad: I'd suggest to have `hasWallets` as a const or export a function to invalidate the query with this key, so we don't have the strings "hasWallets" in multiple places

PR #3233, opened 2024-05-03
fix(setup wallet): fix handle invalid word message

banklesss:
Related to [YOMO-1396](https://emurgo.atlassian.net/browse/YOMO-1396)

[YOMO-1396]: https://emurgo.atlassian.net/browse/YOMO-1396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/features/SetupWallet/common/MnemonicInput/MnemonicInput.tsx
+++ b/apps/wallet-mobile/src/features/SetupWallet/common/MnemonicInput/MnemonicInput.tsx
@@ -23,8 +23,7 @@ export const MnemonicInput = ({
   onFocus,
   mnenonicRefs,
   mnemonic,
-  addInputErrorIndex,
-  removeInputErrorIndex,
+  onError,
 }: {
   length: number
   isValidPhrase: boolean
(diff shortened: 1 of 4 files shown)


apps/wallet-mobile/src/features/SetupWallet/common/MnemonicInput/MnemonicInput.tsx
```diff
@@ -115,6 +123,8 @@ const MnemonicWordsInput = ({
   suggestedWords,
   setSuggestedWords,
   onFocus,
+  addInputErrorIndex,
+  removeInputErrorIndex,
```
michalszorad: Prop names for functions usually start with `on`. Also thinking if an `onError({index})` and `onErrorCleared({index})` would be better. What are your thoughts?

banklesss: yeah that's a good point

michalszorad (approved):

apps/wallet-mobile/src/features/SetupWallet/common/MnemonicInput/MnemonicInput.tsx
```diff
@@ -227,10 +236,18 @@ const MnemonicWordInput = React.forwardRef<MnemonicWordInputRef, MnemonicWordInp
 
           if (suggestedWords.length <= 0) {
             setError('error')
```
stackchain: shouldn't it be a boolean?

PR #3235, opened 2024-05-03
fix(setup wallet): add wallet screen back button

banklesss:
Related to [YOMO-1402](https://emurgo.atlassian.net/browse/YOMO-1402)

[YOMO-1402]: https://emurgo.atlassian.net/browse/YOMO-1402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/features/SetupWallet/SetupWalletNavigator.tsx
+++ b/apps/wallet-mobile/src/features/SetupWallet/SetupWalletNavigator.tsx
@@ -37,6 +37,12 @@ export const SetupWalletNavigator = () => {
       <Stack.Screen
         name="setup-wallet-choose-setup-type"
         component={ChooseSetupTypeScreen}
+        options={{title: strings.addNewWalletTitle}}
+      />
+
+      <Stack.Screen
+        name="setup-wallet-choose-setup-type-init"
+        component={ChooseSetupTypeScreen}
         options={{title: strings.addNewWalletTitle, headerLeft: () => null}}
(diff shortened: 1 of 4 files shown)


apps/wallet-mobile/src/features/SetupWallet/SetupWalletNavigator.tsx
```diff
@@ -37,6 +37,12 @@ export const SetupWalletNavigator = () => {
       <Stack.Screen
         name="setup-wallet-choose-setup-type"
         component={ChooseSetupTypeScreen}
+        options={{title: strings.addNewWalletTitle}}
+      />
+
+      <Stack.Screen
+        name="setup-wallet-choose-setup-type-init"
+        component={ChooseSetupTypeScreen}
         options={{title: strings.addNewWalletTitle, headerLeft: () => null}}
```
stackchain, line 46: dropping the left button can be accomplished by changing the screen options when mounting it, any particular reason to keep 2 routes?

banklesss: I thought that this way the route would be better defined because both paths are accessed from different screens.

My question was how to determine what is the case for hiding or not within the screen with that feature. Also I thought it would make it a little difficult to understand the different routes

PR #3274, opened 2024-05-27
feature(wallet-mobile): NFTs traits

stackchain:

- [x] Added support to NFT traits.
- [x] Updated prod api for mainnet (backend zero)

##### Ticket

YOMO-1457

> [!NOTE]

### DIFF
--- a/packages/portfolio/src/translators/reactjs/usePortfolioTokenTraits.tsx
+++ b/packages/portfolio/src/translators/reactjs/usePortfolioTokenTraits.tsx
@@ -1,0 +1,36 @@
+import {isRight} from '@yoroi/common'
+import {Chain, Portfolio} from '@yoroi/types'
+import {UseQueryOptions, useQuery} from 'react-query'
+
+export function usePorfolioTokenTraits(
+  {
+    id,
+    getTokenTraits,
+    network,
+  }: {
+    id: Portfolio.Token.Id
+    getTokenTraits: Portfolio.Api.Api['tokenTraits']
+    network: Chain.SupportedNetworks
+  },
+  options?: UseQueryOptions<
+    Portfolio.Token.Traits,
+    Error,
+    Portfolio.Token.Traits,
+    [Chain.SupportedNetworks, 'usePorfolioTokenTraits', Portfolio.Token.Id]
+  >,
+) {
+  const query = useQuery({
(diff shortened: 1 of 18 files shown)


packages/portfolio/src/validators/token-traits.ts
```diff
@@ -0,0 +1,25 @@
...
+export const TraitSchema = z.object({
+  type: z.string(),
+  value: z.string(),
+  rarity: z.string(),
+})
+
+const TraitsSchema = z.object({
+  totalItems: z.number().nonnegative(),
+  traits: z.array(TraitSchema),
+  // useful only if client doen't have the info
```
stackchain, line 13:
```suggestion
  // useful only if client doesn't have the info
```

packages/portfolio/src/translators/reactjs/usePortfolioTokenTraits.tsx
```diff
@@ -0,0 +1,36 @@
...
+    queryFn: async () => {
+      const response = await getTokenTraits(id)
+      if (isRight(response)) return response.value.data
+      throw new Error('usePorfolioTokenTraits')
+    },
+  })
+
+  return {
+    ...query,
+    tokenTraits: query.data,
```
banklesss, line 34: If i am not wrong traits its a property of data right? `query.data.traits`

would it be a good idea to get the traits so you dont need to use `tokenTraits.traits`?

or does `query.data` contain more useful properties?

banklesss: ok i see it includes total items property

banklesss:
```
 return {
    ...query,
    tokenTraits: query.data.traits,
    totalItems: query.data.totalItems,
 }
```

Maybe like this?

To avoid `tokenTraits.traits`

banklesss (approved):

michalszorad (approved):

jorbuedo (approved):

packages/portfolio/src/translators/reactjs/usePortfolioTokenTraits.tsx
stackchain: yap, some info are not important cuz it does have the `info` before hitting it.

PR #3285, opened 2024-06-07
feat(pool-transition) Update pool transition api and staking

jorbuedo:

New Api uses S3 file to check pool transition.
Also updates adapool bias param for ordering.


stackchain (approved):

zuzunker (commented):
LGTM, with the only note that the version of the lib is expected to change from alpha at some point

PR #3286, opened 2024-06-09
feature(wallet-mobile): portfolio UI

stackchain:
##### Ticket

YOMO-

> [!NOTE]


stackchain:
@lephuochoai 

I'll be merging since it doesn't prevent releases, but it's a big blocker for another change that is happening, the list of fixes will be shared on slack.

- [ ] there is a flickering while scrolling up token details

PR #3288, opened 2024-06-10
feat(catalyst): dark theme

banklesss:

Applying Dark Theme to Catalyst feature.

##### Ticket

YOMO-1474

> [!NOTE]


apps/wallet-mobile/src/components/Button/Button.tsx
```diff
@@ -90,7 +90,7 @@ export const Button = (props: ButtonProps) => {
             textStyles,
           ]}
         >
-          {title}
+          {title.toLocaleUpperCase()}
```
michalszorad: Was this agreed a design decision to always use uppercase?

banklesss: changed from here https://github.com/Emurgo/yoroi/pull/3288/files#diff-33ab3a81ad151023035e7e2668df9f2e265da93493fe6afca547ed4eaedcb8cfL145

😉

banklesss: I'm going to restore the change because i just noticed it is affecting some implementation of the button that disables that function.

apps/wallet-mobile/src/components/Button/Button.tsx
```diff
@@ -138,7 +138,7 @@ const useStyles = () => {
       borderColor: color.primary_c600,
     },
     text: {
-      color: color.gray_cmin,
+      color: isDark ? color.white_static : color.gray_cmin,
```
stackchain: @banklesss can you check with Dima, it shouldn't have this type of check in the theme, it should use a `sub-token` -> `color.title-1`

stackchain (approved):

PR #3327, opened 2024-06-25
fix(wallet-mobile): balances - protocolparams

stackchain:

- Protocol params can change per epoch after Conway, added persistence to it
- Ignoring API errors, falling back to hardcoded values 
- Considering consuming it straight from network-manager since is too tight to flavour


stackchain:
@banklesss the issue with no balances was some throwing not handled, another step to a more off-line first after this PR.

banklesss:
>@banklesss the issue with no balances was some throwing not handled, another step to a more off-line first after this PR.

let me know if i can help you with anything 👍🏻

banklesss (approved):

PR #3328, opened 2024-06-25
chore(all): Bump react query

jorbuedo:
##### Ticket

[YOMO-1407
](https://emurgo.atlassian.net/browse/YOMO-1407)

Updating react-query comes with a lot of changes, and it got stuck with the ract-native bump.
With only this bit, we can run it for a while and make sure this part at least works.

Be careful with query functions returning `undefined`, as that is now and invalid return value.


packages/resolver/src/translators/reactjs/hooks/useResolverSetShowNotice.test.tsx
```diff
@@ -27,17 +27,15 @@ describe('useResolverSetShowNotice', () => {
       resolverManager: mockResolverManager,
     })
 
-    const {result, waitFor: waitForHook} = renderHook(
-      () => useResolverSetShowNotice(),
-      {wrapper},
-    )
-
+    renderHook(() => useResolverSetShowNotice(), {wrapper})
+    /*
```
jorbuedo, line 31: @banklesss can't get the commented part to run 😞

michalszorad (approved):

PR #3333, opened 2024-06-26
chore(wallet-mobile): Rename Layout to LinearTransition

jorbuedo:

This causes some noise in the react-native bump, because it conflicts all the time with other changes in develop.

`Layout` is an alias for `LinearTransition` that react-native-reanimated is deprecating.
https://github.com/software-mansion/react-native-reanimated/blob/3.1.0/src/reanimated2/layoutReanimation/defaultTransitions/LinearTransition.ts#L48

image

By having it already changed in develop, it'll make life easier when upgrading the dependency.

### DIFF
--- a/apps/wallet-mobile/ios/Podfile.lock
+++ b/apps/wallet-mobile/ios/Podfile.lock
@@ -482,7 +482,7 @@ PODS:
     - React-Core
   - RNPermissions (3.8.4):
     - React-Core
-  - RNReanimated (3.3.0):
+  - RNReanimated (3.5.0):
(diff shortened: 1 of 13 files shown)


banklesss (approved):

stackchain (approved):

stackchain:
@jorbuedo update the pods too.

jorbuedo:
Double checking the version that supports this change, but not so much that it breaks other things or needs a bump in react-native... 1 sec.

michalszorad (approved):

PR #3337, opened 2024-06-27
chore: Adds reset and build scripts without running tests

jorbuedo:

`yarn reset:dev` to clean install everything without running tests.


packages/api/package.json
```diff
@@ -47,6 +47,7 @@
   ],
   "scripts": {
     "build": "yarn tsc && yarn lint && yarn test --ci --silent && yarn clean && bob build",
+    "build:dev": "yarn clean && bob build",
```
michalszorad: `yarn tsc` as well?

michalszorad: Same applies below in other packages

jorbuedo: I don't know, it's adding more time. Depends if we care about tsc when running this.

michalszorad: I'd say to keep tsc as well, so it's easier to spot TS bugs locally

michalszorad (approved):

stackchain:
@jorbuedo bringing in after rebasing.

PR #3338, opened 2024-06-27
refactor(common): array keys in queries 

banklesss:

using array keys in queries 

> [!NOTE]


apps/wallet-mobile/src/legacy/Staking/PoolTransition/usePoolTransition.tsx
```diff
@@ -83,7 +83,7 @@ export const usePoolTransition = () => {
 
   return {
     ...poolTransitionQuery,
-    isLoading: isLoading || poolTransitionQuery.isLoading,
```
banklesss: https://tanstack.com/query/v4/docs/framework/react/guides/disabling-queries#isinitialloading

michalszorad (approved):

jorbuedo (approved):

banklesss:
to merge when the react native bump is finsihed

stackchain:
@banklesss can you use this PR to update the key management transform all into arrays only, so when it comes the time to bump it, would be simpler to do it? (can use the revert to check all places that were updated)

michalszorad (approved):

PR #3390, opened 2024-07-11
fix(wallet-mobile): Dark theme nft placeholder

jorbuedo:
##### Ticket

[YOMO-1657
](https://emurgo.atlassian.net/browse/YOMO-1657)

Can't reproduce original issue (placeholder not showing)


apps/wallet-mobile/src/assets/img/nft-placeholder-dark.png
stackchain, line 1: Can't it leverage the `tintColor` property? (when the image it is not transparent), then it could use the same image.

apps/wallet-mobile/src/components/NftPreview/NftPreview.tsx
```diff
@@ -1,10 +1,12 @@
 import {isString} from '@yoroi/common'
+import {useTheme} from '@yoroi/theme'
 import {Balance} from '@yoroi/types'
 import React from 'react'
 import {Image, ImageStyle, View} from 'react-native'
 import SkeletonPlaceholder from 'react-native-skeleton-placeholder'
 
-import placeholder from '../../assets/img/nft-placeholder.png'
+import placeholderLight from '../../assets/img/nft-placeholder.png'
+import placeholderDark from '../../assets/img/nft-placeholder.png'
```
stackchain:
```suggestion
import placeholderDark from '../../assets/img/nft-placeholder-dark.png'
```

stackchain (approved):

apps/wallet-mobile/src/assets/img/nft-placeholder-dark.png
jorbuedo: I can check but there's 2 colors (background and foreground), I don't think tinting it would result in the same result.

jorbuedo: Maybe with a new image that had the foreground as transparent we could try to tint the background part. As long as the background of the component had the right color, could work maybe.

PR #3412, opened 2024-07-18
fix(wallet-mobile): Restore pin keyboard style

jorbuedo:

- Restores pin keyboard
- Fix tx history panel background

### DIFF
--- a/apps/wallet-mobile/src/components/NumericKeyboard.tsx
+++ b/apps/wallet-mobile/src/components/NumericKeyboard.tsx
@@ -3,7 +3,6 @@ import React from 'react'
 import {StyleSheet, TouchableHighlight, View, ViewProps} from 'react-native'
 
 import {Backspace} from './Icon/Backspace'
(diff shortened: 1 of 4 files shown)


apps/wallet-mobile/src/components/NumericKeyboard.tsx
```diff
@@ -136,30 +108,19 @@ const useStyles = () => {
...
-    },
-    keyboardKeyDisabled: {
-      backgroundColor: color.gray_c100,
-    },
-    emptyKey: {
+    specialKey: {
       flex: 1,
       alignItems: 'center',
       justifyContent: 'center',
-      borderRadius: 5,
```
stackchain, line 162: just confirming: no more rounded corners?

jorbuedo: No, back to the previous keyboard. We talked about it yesterday and there never was any intention from product to change it to look more like iOS.

PR #3422, opened 2024-07-23
fix(tx-list): background color

banklesss:
##### Ticket

YOMO-1701

> [!NOTE]


apps/wallet-mobile/src/features/ListTxHistory/useCases/ListTxHistory/TxHistory.tsx
```diff
@@ -103,7 +103,7 @@ const useStyles = () => {
     panel: {
       flex: 1,
       paddingTop: 8,
-      backgroundColor: color.gray_cmin,
+      backgroundColor: isDark ? color.gray_c50 : color.white_static,
```
banklesss: exception

jorbuedo: That's `bg_color_high`, already changes to `gray_c50` or white depending on dark or not.
I changed them all to that and then changed this one back because it looked bad with the items in it, didn't realize it was the items the ones that needed to be set to that as well.

stackchain (approved):

PR #3459, opened 2024-07-26
feature(network-switcher): header info and left overs

banklesss:

Implementation of the missing things and the header feature

##### Ticket

YOMO-1797
YOMO-1516


apps/wallet-mobile/src/WalletNavigator.tsx
```diff
@@ -186,7 +188,14 @@ export const WalletNavigator = () => {
         >
           <Stack.Screen
             name="wallet-selection"
-            options={{title: strings.walletSelectionScreenHeader}}
+            options={{
+              title: strings.walletSelectionScreenHeader,
+              headerTitle: ({children}) => (
+                <NetworkTag disabled directChangeOnDevActive>
```
banklesss: `directChangeOnDevActive`

😉

banklesss: deprecated

apps/wallet-mobile/src/features/Settings/ChangeNetwork/NetworkTag.tsx
```diff
@@ -3,32 +3,82 @@ import {Chain} from '@yoroi/types'
...
 
+  const onPress = () => {
+    if (directChangeActive) {
+      const networks: Array<Chain.SupportedNetworks> = [
+        Chain.Network.Mainnet,
+        Chain.Network.Preprod,
+        Chain.Network.Sancho,
+      ]
+
+      const nextNetwork = networks[(networks.indexOf(network) + 1) % networks.length]
```
stackchain: @banklesss can you confirm with Product but I believe it will always return to mainnet when it's not.

banklesss: I just confirmed with Product that the wallet list screen is the exception and you can directly change the network when you tap on the header without needing to go to the network list screen

I copied that logic directly from the previous implementation with some minor changes.

Please tell me if you think it needs any improvement

banklesss:
Converted to draft to cover possible new missing parts

apps/wallet-mobile/src/features/Settings/ChangeNetwork/NetworkTag.tsx
banklesss: I think I misunderstood the message, is that something you think is incorrect?

Product wanted the network switcher in the wallet list as it was originally designed: without needing to go to the network list and sequential change.

Please tell me if there is something I am missing!

apps/wallet-mobile/src/features/Exchange/useCases/CreateExchangeOrderScreen/CreateExchangeOrderScreen.tsx
```diff
@@ -126,32 +136,30 @@ export const CreateExchangeOrderScreen = () => {
           >
             <SelectBuyOrSell disabled={isLoading} />
 
-            <Space height="xl" />
-
-            <EditAmount disabled={isLoading} />
-
-            <Space height="_2xs" />
+            {isPreprod && orderType === 'buy' ? (
+              <ShowPreprodNotice />
```
michalszorad, line 145: Should this be shown for sancho as well?

apps/wallet-mobile/src/features/Portfolio/useCases/PortfolioTokensList/PortfolioWalletTokenList/PortfolioWalletTokenList.tsx
```diff
@@ -77,7 +82,7 @@ export const PortfolioWalletTokenList = () => {
         <View>
           <Spacer height={16} />
 
-          <BuyADABanner />
+          {network === Chain.Network.Preprod ? <PreprodFaucetBanner /> : <BuyADABanner />}
```
michalszorad: May need to be shown for sancho as well

apps/wallet-mobile/src/features/Exchange/useCases/CreateExchangeOrderScreen/CreateExchangeOrderScreen.tsx
banklesss: yeah good point

i assume is not specified because sancho is only displayed for us, developers 🤓

michalszorad (approved):

jorbuedo (approved):

PR #3464, opened 2024-07-29
fix(tx-list): updating with new txs

banklesss:
##### Ticket

YOMO-1689

### DIFF
--- a/apps/wallet-mobile/src/features/Transactions/useCases/TxList/TxList.tsx
+++ b/apps/wallet-mobile/src/features/Transactions/useCases/TxList/TxList.tsx
@@ -18,11 +18,19 @@ export const TxList = (props: Props) => {
   const {wallet} = useSelectedWallet()
 
   const filter = useTxFilter()
-  const transactions = useTransactionInfos({wallet})
-  const filteredTransactions = React.useMemo(() => filterTransactions(transactions, filter), [transactions, filter])
+  const transactionInfos = useTransactionInfos({wallet})
+  const filteredTransactions = React.useMemo(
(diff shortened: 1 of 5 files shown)


apps/wallet-mobile/src/features/Transactions/useCases/TxList/TxList.tsx
```diff
@@ -18,12 +18,20 @@ export const TxList = (props: Props) => {
...
+    () => filterTransactions(transactionInfos, filter),
+    [transactionInfos, filter],
+  )
 
   const [loadedTxs, setLoadedTxs] = React.useState(filteredTransactions.slice(0, batchSize))
   const [currentIndex, setCurrentIndex] = React.useState(batchSize)
 
+  React.useEffect(() => {
+    setLoadedTxs(filteredTransactions.slice(0, batchSize))
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
michalszorad, line 32: Does it need a `setCurrentIndex` as well?

jorbuedo (approved):

apps/wallet-mobile/src/features/Transactions/useCases/TxList/TxList.tsx
banklesss: Transactions can change, but the index depends on the user's movements

so if new transactions appear, the user will have the list as he had it (tested)

stackchain: hmm, I think this would be affecting all current infinity lists.

PR #3477, opened 2024-08-01
fix(swap): tx

banklesss:

Wrong metadata naming

##### Ticket

YOMO-1800

> [!NOTE]

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/common/useSwapTx.ts
+++ b/apps/wallet-mobile/src/features/Swap/common/useSwapTx.ts
@@ -28,8 +28,7 @@ export const useSwapTx = (options?: UseMutationOptions<YoroiUnsignedTx, Error, {
 
   const {wallet, meta} = useSelectedWallet()
   const mutation = useMutationWithInvalidations({
-    mutationFn: (data) =>
-      wallet.createUnsignedTx({entries: data.entries, auxiliaryData: metadata, addressMode: meta.addressMode}),
+    mutationFn: (data) => wallet.createUnsignedTx({entries: data.entries, metadata, addressMode: meta.addressMode}),
     invalidateQueries: [['useCreateOrder']],
     ...options,
   })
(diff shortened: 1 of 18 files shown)


apps/wallet-mobile/src/features/Swap/common/useSwapTx.ts
```diff
@@ -28,7 +28,8 @@ export const useSwapTx = (options?: UseMutationOptions<YoroiUnsignedTx, Error, {
 
   const {wallet, meta} = useSelectedWallet()
   const mutation = useMutationWithInvalidations({
-    mutationFn: (data) => wallet.createUnsignedTx({entries: data.entries, metadata, addressMode: meta.addressMode}),
```
jorbuedo: Do we know since when `createUnsignedTx` expects `auxiliaryData` instead of `metadata`?

jorbuedo (approved):

apps/wallet-mobile/src/features/Swap/common/useSwapTx.ts
banklesss: the wallet class has changed a lot recently so at some point it changed and we didn't update that function accordingly

jorbuedo: Typescript should have complained that the function inside the wallet class doesn't have the same arguments as the type.

stackchain: yap the issue is since both are optionals, it looks like TS (linter) at least, just disregard them, @banklesss I think it should be `metadata` since `auxiliaryData` was the old name for the tx metadatum, update the ledger part, and make sure that all is passing as `metadata` please.

banklesss: @stackchain 

i just found the ledger `auxiliaryData` is needed because this [type](https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/blob/ee543694deaa9dd39d049d3efb888515bf058faa/src/types/public.ts#L1705) 

that type is part of this [type](https://github.com/Emurgo/yoroi/blob/develop/apps/wallet-mobile/src/features/Discover/common/ledger.ts#L201)

banklesss: `createUnsignedTx` with metadata is only used to make swaps

PR #3499, opened 2024-08-07
feature(wallet-mobile): Adds token activity to show non ada prices

jorbuedo:
##### Ticket
[YOMO-1495
](https://emurgo.atlassian.net/browse/YOMO-1495)

### DIFF
--- a/apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
+++ b/apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
@@ -1,4 +1,4 @@
-import {isNonNullable} from '@yoroi/common'
+import {invalid, isNonNullable} from '@yoroi/common'
 import {Portfolio} from '@yoroi/types'
 import React from 'react'
 import {useQuery, useQueryClient} from 'react-query'
@@ -93,7 +93,9 @@ export const PortfolioTokenActivityProvider = ({children}: Props) => {
   return <PortfolioTokenActivityContext.Provider value={value}>{children}</PortfolioTokenActivityContext.Provider>
 }
 
-export const usePortfolioTokenActivity = () => React.useContext(PortfolioTokenActivityContext) ?? {}
+export const usePortfolioTokenActivity = () =>
+  React.useContext(PortfolioTokenActivityContext) ??
+  invalid('usePortfolioTokenActiviy requires PortfolioTokenActivitiyProvider')
 
 type PortfolioTokenActivityContext = {
   aggregatedBalances?: Portfolio.Token.AmountRecords
--- a/apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
+++ b/apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
@@ -1,0 +1,101 @@
+import {isNonNullable} from '@yoroi/common'
+import {Portfolio} from '@yoroi/types'
+import React from 'react'
+import {useQuery, useQueryClient} from 'react-query'
+import {merge, switchMap} from 'rxjs'
+
+import {time} from '../../../kernel/constants'
+import {queryInfo} from '../../../kernel/query-client'
+import {useSelectedNetwork} from '../../WalletManager/common/hooks/useSelectedNetwork'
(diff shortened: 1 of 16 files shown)


banklesss:
i think we need to adapt the price from tx list.

The logic there is similar

jorbuedo:
> i think we need to adapt the price from tx list.
> 
> The logic there is similar

Right now TxList only shows prices for the Ada part of any tx. In the portfolio view for specific tokens, it shows the number of different assets in the tx. 
So unless we want to add up the total value of a tx with the containing assets, there's nothing to change yet.

banklesss (approved):

apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+const useTokenActivity = () => {
+  const {walletManager} = useWalletManager()
+  const {
+    networkManager: {tokenManager},
+  } = useSelectedNetwork()
+  const [aggregatedBalances, setAggregatedBalances] = React.useState<Portfolio.Token.AmountRecords>()
+  const [tokenIds, setTokenIds] = React.useState<Portfolio.Token.Id[]>()
+  const queryClient = useQueryClient()
+
+  React.useEffect(() => {
```
michalszorad, line 22: Is it worth extracting it into a custom hook as we had before? Eg `useAggregatedBalances`

jorbuedo: Here's my reasoning.
AggregatedBalances and TokenActivity could be separate, they're independent features.
But they basically use the same subscription, so I think it's a waste of resources to have them both running the same thing when it can be just executed once.
Having this as an extracted hook may give the impression that it can be reused, which there's no need to.

apps/wallet-mobile/src/features/Portfolio/common/helpers/aggregatePrimaryAmount.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  return Object.values(tokenAmountRecords).reduce(
+    (totalAmount, tokenAmount) => {
+      const tokenPrimaryPrice = tokenActivity?.[tokenAmount.info.id]?.price24h.close ?? new BigNumber(0)
+
+      const quantity =
+        tokenAmount.info.id === primaryTokenInfo.id
+          ? tokenAmount.quantity
+          : BigInt(
+              amountBreakdown(tokenAmount).bn.times(tokenPrimaryPrice).shiftedBy(primaryTokenInfo.decimals).toFixed(0),
```
stackchain, line 26: `amountBreakdown` has `.str` I think is accomplishing the same.

apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
```diff
@@ -0,0 +1,101 @@
...
+type Props = {
+  children: React.ReactNode
+}
+
+export const PortfolioTokenActivityProvider = ({children}: Props) => {
+  const value = useTokenActivity()
+  return <PortfolioTokenActivityContext.Provider value={value}>{children}</PortfolioTokenActivityContext.Provider>
+}
+
+export const usePortfolioTokenActivity = () => React.useContext(PortfolioTokenActivityContext) ?? {}
```
stackchain: it helps when writing stories.
```suggestion
export const usePortfolioTokenActivity = () => React.useContext(PortfolioTokenActivityContext) ?? invalid('usePortfolioTokenActiviy requires PortfolioTokenActivitiyProvider')
```

stackchain (approved):
looking real good

apps/wallet-mobile/src/features/Portfolio/common/helpers/aggregatePrimaryAmount.ts
jorbuedo: For the `toFixed` you mean? After getting the bignumber, multiply and shift, it's a bignumber, not an amountBreakdown, so it doesn't have the .str anymore.

stackchain: I missed the `.times` yap

PR #3534, opened 2024-08-16
fix(receive): qr content

banklesss:
##### Ticket

YOMO-1847

### DIFF
--- a/apps/wallet-mobile/src/components/ShareQRCodeCard/ShareQRCodeCard.tsx
+++ b/apps/wallet-mobile/src/components/ShareQRCodeCard/ShareQRCodeCard.tsx
@@ -13,7 +13,8 @@ import {Space} from '../Space/Space'
 import {CaptureShareQRCodeCard} from './CaptureShareQRCodeCard/CaptureShareQRCodeCard'
 
 type ShareQRCodeCardProps = {
-  content: string
+  qrContent: string
+  shareContent: string
   title: string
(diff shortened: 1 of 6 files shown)


apps/wallet-mobile/src/components/ShareQRCodeCard/ShareQRCodeCard.tsx
```diff
@@ -13,7 +13,8 @@ import {Space} from '../Space/Space'
 import {CaptureShareQRCodeCard} from './CaptureShareQRCodeCard/CaptureShareQRCodeCard'
 
 type ShareQRCodeCardProps = {
-  content: string
+  qrContent: string
+  shareContent: string
```
michalszorad, line 17: What is the difference between these two please?

stackchain: I believe shared is carrying all text, while qrContent is just the link, yet needs confirmation from @banklesss

banklesss: Yeah you are right @stackchain. Maybe we can improve the wording. I can take a look at it

michalszorad (approved):
Maybe just the wording of params qrContent and shareContent, as it was not quite clear from the beginning, but looks good.

PR #3540, opened 2024-08-16
chore(portfolio): support new API response for multi token when 1 fail

stackchain:

- **chore(types): new cache response for multi token - 500 with message**
- **chore(portfolio): updated multi token to handle any error with 500 record**
- **chore(root): bumped types/links/portfolio versions**
- **chore(common): updated for new API cache record]**

##### Ticket

YOMO-1844

> [!NOTE]


jorbuedo (approved):
Should we log them in some way, or would the backend check up on the failing records?

stackchain:
> Should we log them in some way, or would the backend check up on the failing records?

There is a tool, openobserver for logging, in case something is odd we can query the logs for "[500," there.

PR #3561, opened 2024-08-26
fix(send): edit send amount empty state

banklesss:

When returning to the previous screen from the "edit amount"  screen with the value set to zero, the token remained selected, affecting the routes.

##### Ticket

YOMO-1852


apps/wallet-mobile/src/features/Send/useCases/ListAmountsToSend/EditAmount/EditAmountScreen.tsx
```diff
@@ -53,6 +65,25 @@ export const EditAmountScreen = () => {
     setInputValue(atomicBreakdown(initialQuantity, amount.info.decimals).bn.toFormat())
   }, [amount.info.decimals, initialQuantity])
 
+  React.useLayoutEffect(() => {
```
michalszorad: What if I press hardware button or swipe left? Shoult it be just a useEffect with a cleanup function that changes the amount?

banklesss: yeah good point

banklesss: will have a look

banklesss: done

added `isFocused` because typing triggers useEffect/useFocusEffect return function

stackchain (approved):

PR #3564, opened 2024-08-28
fix(common): styling minors

banklesss:

Styling minors.

##### Ticket

YOMO-1865
YOMO-1455

### DIFF
--- a/apps/wallet-mobile/src/components/Button/Button.tsx
+++ b/apps/wallet-mobile/src/components/Button/Button.tsx
@@ -156,7 +156,7 @@ const useStyles = () => {
       opacity: 0.5,
     },
     buttonDisabledText: {
-      color: color.gray_cmin,
+      color: color.text_gray_min,
     },
     shelleyTheme: {
       backgroundColor: color.primary_c500,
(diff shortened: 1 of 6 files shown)


apps/wallet-mobile/src/components/Button/Button.tsx
```diff
@@ -153,6 +154,10 @@ const useStyles = () => {
     },
     buttonDisabled: {
       opacity: 0.5,
+      color: color.gray_cmin,
```
banklesss:
```suggestion
```

michalszorad (approved):

apps/wallet-mobile/src/components/Button/Button.tsx
```diff
@@ -154,6 +155,9 @@ const useStyles = () => {
     buttonDisabled: {
       opacity: 0.5,
     },
+    buttonDisabledText: {
+      color: color.gray_cmin,
```
stackchain: text, should be `text_..min`

stackchain: confirm if is the case for the buttons as well.

banklesss: text_gray_min is not created 👀

will have a look

banklesss: done

PR #3570, opened 2024-08-28
feature(discover): intl filter options

banklesss:

Internationalization of discover filter options

##### Ticket

YOMO-1866


apps/wallet-mobile/src/features/Discover/common/useStrings.tsx
```diff
@@ -229,5 +243,114 @@ export const messages = Object.freeze(
...
+      case 'Trading Tools':
+        return strings.filterChildOptionsTradingTools
+
+      case 'DAO':
+        return strings.filterChildOptionsDAO
+
+      case 'Decentralised Storage':
+        return strings.filterChildOptionsDecentralisedStorage
+
+      default:
```
michalszorad, line 352: Can we default to the input value instead of null? This is to allow us to dynamically change the options in case of emergency (even though I don't see any reason why would it happen right now)

michalszorad: Ahh I see you already have the defaults. OK!

michalszorad (approved):

PR #3586, opened 2024-09-02
refactor(wallet-mobile): drop legacy primary token

stackchain:

- **chore(portfolio): export ptId + bump v1.0.2**
- **refactor(wallet-mobile): portfolio token**

##### Ticket

YOMO-

> [!NOTE]


apps/wallet-mobile/src/features/Swap/common/helpers.ts
```diff
@@ -80,15 +80,15 @@ export const parseOrderTxMetadata = (metadataJson: string, primaryTokenId: strin
 }
 
 const normalisePrimaryTokenId = (tokenId: string, primaryTokenId: string) => {
-  return tokenId === '.' || tokenId === '' ? primaryTokenId : tokenId
+  return tokenId === ptId || tokenId === '' ? primaryTokenId : tokenId
```
michalszorad: If the `primaryTokenId` from portfolio gets changed to anything else than `'.'` then it may cause issues here, as we need to check for `'.'` and `''`. Maybe I'd suggest to add another check `tokenId === '.' || tokenId === '' || tokenId === ptId ? primaryTokenId : tokenId`

stackchain: it's actually better and correct, since it needs to keep all legacy ids, updating it

stackchain: added a list of `swapPtTokenIds`

michalszorad (approved):

PR #3593, opened 2024-09-03
fix(wallet-mobile): fix app background color

jorbuedo:
##### Ticket

[YOMO-1873](https://emurgo.atlassian.net/browse/YOMO-1873)

[YOMO-1873]: https://emurgo.atlassian.net/browse/YOMO-1873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/jestSetup.ts
+++ b/apps/wallet-mobile/jestSetup.ts
@@ -8,6 +8,10 @@ logger.disable()
 
 global.fetch = fetch
 
+jest.mock('expo-system-ui', () => ({
+  setBackgroundColorAsync: async () => undefined,
+  getBackgroundColorAsync: async () => 'black',
+}))
 jest.mock('react-native-device-info', () => ({getVersion: () => '1.5.1'}))
 jest.mock('react-native-randombytes', () => require('crypto').randomBytes)
 jest.mock('react-native-background-timer', () => {})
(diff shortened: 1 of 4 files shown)


banklesss (approved):

stackchain:
@jorbuedo can you check the minimal react-native version?

jorbuedo:
> @jorbuedo can you check the minimal react-native version?

SystemUI is available since expo sdk 49 which is the one we're in: https://docs.expo.dev/versions/v49.0.0/sdk/system-ui/
I installed it with the expo tool and it set the version to 2 something, despite latest being 3 something.
And it's working locally.

So I don't know what's complaining about, I'll try named import instead of *

stackchain:
> > @jorbuedo can you check the minimal react-native version?
> 
> SystemUI is available since expo sdk 49 which is the one we're in: https://docs.expo.dev/versions/v49.0.0/sdk/system-ui/ I installed it with the expo tool and it set the version to 2 something, despite latest being 3 something. And it's working locally.
> 
> So I don't know what's complaining about, I'll try named import instead of *

maybe is jest, then it needs to be added to the ignore pattern, and mocked in the jest setup.

jorbuedo:
> maybe is jest, then it needs to be added to the ignore pattern, and mocked in the jest setup.

Yes, just check, it's jest the one complaining

PR #3604, opened 2024-09-06
fix(swap): screen freezing when changing tabs with keyboard open

banklesss:
##### Ticket

YOMO-1916

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/SwapNavigator.tsx
+++ b/apps/wallet-mobile/src/features/Swap/SwapNavigator.tsx
@@ -68,7 +68,7 @@ export const SwapTabNavigator = () => {
             listeners={{
               tabPress: (e) => {
                 // if the keyboard is open, the user needs to close first the keyboard
-                // then, press again the tab change screen
+                // then, press again the tab to change screen
                 // to avoid screen freezing
                 if (isKeyboardOpen) {
                   Keyboard.dismiss()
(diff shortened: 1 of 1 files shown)


apps/wallet-mobile/src/features/Swap/SwapNavigator.tsx
```diff
@@ -62,9 +64,37 @@ export const SwapTabNavigator = () => {
...
+                  e.preventDefault()
+                }
+              },
+            }}
+            name="token-swap"
+            component={StartSwapOrderScreen}
+          />
 
-          <Tab.Screen name="orders" getComponent={() => ListOrders} />
+          <Tab.Screen
```
jorbuedo, line 83: Why do we use `component` in one and `getComponent` with a function that returns the component in the other one? (not PR related, just saw it)

apps/wallet-mobile/src/features/Swap/SwapNavigator.tsx
```diff
@@ -62,9 +64,37 @@ export const SwapTabNavigator = () => {
           })}
           style={styles.tab}
         >
-          <Tab.Screen name="token-swap" component={StartSwapOrderScreen} />
+          <Tab.Screen
+            listeners={{
```
jorbuedo, line 68: Seems like an iOS only bug, could it be happening in other screens?
Maybe abstract the whole listener so it can be spread into any Tab.Screen?

apps/wallet-mobile/src/features/Swap/SwapNavigator.tsx
stackchain: getComponent is lazy, since the start order is the first render, the second tab is only loaded when clicked. 
this bug looks like some animation fighting problem

stackchain (approved):
approving for now, lets check after bumping .74 if it still persist.

PR #3619, opened 2024-09-11
fix(wallet-mobile): adds log for token activity error

jorbuedo:

Previous log (called directly by react-query when throwing)
```
{"message": undefined, "responseData": {"reasons": {"": [Array]}}, "status": 500}
```

With this change:
 ```
{"endpoint":"tokenActivity","status":500,"responseData":{"reasons":{"":["unexpected error"]}}}
```

### DIFF
--- a/apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
+++ b/apps/wallet-mobile/src/features/Portfolio/common/PortfolioTokenActivityProvider.tsx
@@ -7,6 +7,7 @@ import {useQuery, useQueryClient} from 'react-query'
 import {merge, switchMap} from 'rxjs'
 
 import {time} from '../../../kernel/constants'
+import {logger} from '../../../kernel/logger/logger'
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/common/src/api/fetchData.ts [hunk 1]
```diff
@@ -102,6 +102,8 @@ export const fetchData: FetchData = <T, D = any>(
         const message = error.response.statusText
         const responseData = error.response.data
 
+        console.error('fetchData error: ', config.url)
```
banklesss:
```suggestion
```

banklesss (approved):

packages/common/src/api/fetchData.ts
jorbuedo: Ugh, commited that by mistake

stackchain:
it is returning `500`, just a quick recall that by not throwing react-query will return `isSuccess: true, isError: false`, even when failing, with `data: null`, I think is ok for debugging, but it looks like the issue might be sitting in the request.

PR #3628, opened 2024-09-13
chore(types): added epoch to protocol params api response

stackchain:
## Description / Change(s) / Related issue(s)

- **chore(api): added epoch to protocol-params api**
- **chore(types): added epoch to protocol params api response**
- **fix(api): protocol params test**
##### Ticket

YOMO-

> [!NOTE]
> Please create the ticket if missing it.

PR #3629, opened 2024-09-13
chore(wallet-mobile): update protocol params

stackchain:

- Wraps up #3628  

##### Ticket

YOMO-

> [!NOTE]

### DIFF
--- a/apps/wallet-mobile/src/features/WalletManager/network-manager/network-manager.ts
+++ b/apps/wallet-mobile/src/features/WalletManager/network-manager/network-manager.ts
@@ -72,6 +72,7 @@ export const protocolParamsPlaceholder = freeze({
   coinsPerUtxoByte: '4310',
   poolDeposit: '500000000',
   keyDeposit: '2000000',
+  epoch: 509,
 })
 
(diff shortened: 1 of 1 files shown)


jorbuedo (approved):

apps/wallet-mobile/src/features/WalletManager/network-manager/network-manager.ts
```diff
@@ -72,6 +72,7 @@ export const protocolParamsPlaceholder = freeze({
   coinsPerUtxoByte: '4310',
   poolDeposit: '500000000',
   keyDeposit: '2000000',
+  epoch: 509,
```
jorbuedo, line 75: Dangerous, I assume we don't actually use this and is just there to satisfy the type

banklesss (approved):

apps/wallet-mobile/src/features/WalletManager/network-manager/network-manager.ts
stackchain: it's actually safe, it can be the last epoch before voting begins (static), if the request to params fails between epochs and it actually needs to fallback definitely a warning about using outdated values are a good idea though.

PR #3634, opened 2024-09-17
refactor(wallet-mobile): drop indexes export + unused exports/code

stackchain:
##### Ticket

YOMO-

> [!NOTE]


banklesss (approved):

apps/wallet-mobile/src/components/ScrollableView/ScrollableView.tsx
```diff
@@ -1,16 +1,6 @@
 import * as React from 'react'
-import {ScrollView, ScrollViewProps} from 'react-native'
+import {ScrollView} from 'react-native'
 
 const ScrollViewContext = React.createContext<{current: ScrollView | null}>({current: null})
 
 export const useScrollView = () => React.useContext(ScrollViewContext).current
```
michalszorad: Is this `useScrollView`  ever used? Seems to be redundant without the provider

michalszorad (approved):

apps/wallet-mobile/src/components/ScrollableView/ScrollableView.tsx
stackchain: dunno, I was just pruning unused stuff, will check in another PR, still on going.

jorbuedo: Seems to be used to get `isScrollBarShown` to add the top border of the action footers

stackchain: @banklesss it also looks that in the new Restore funnel it was used to auto scroll, but it is wrapped within a regular view instead of ScrollableView, I've marked you to check it.

PR #3645, opened 2024-09-23
refactor(staking): withdrawal dialog

banklesss:

Refactor of the legacy staking withdrawal dialog

##### Ticket

YOMO-1899

### DIFF
--- a/apps/wallet-mobile/src/legacy/Dashboard/WithdrawStakingRewards/WithdrawStakingRewards.tsx
+++ b/apps/wallet-mobile/src/legacy/Dashboard/WithdrawStakingRewards/WithdrawStakingRewards.tsx
@@ -6,42 +6,51 @@ import {ScrollView, StyleSheet, Text, View} from 'react-native'
 import {Boundary} from '../../../components/Boundary/Boundary'
 import {Button} from '../../../components/Button/Button'
 import {Checkbox} from '../../../components/Checkbox/Checkbox'
+import {useModal} from '../../../components/Modal/ModalContext'
 import {PleaseWaitView} from '../../../components/PleaseWaitModal'
 import {Space} from '../../../components/Space/Space'
 import {Warning} from '../../../components/Warning/Warning'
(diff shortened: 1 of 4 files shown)


apps/wallet-mobile/src/legacy/Dashboard/WithdrawStakingRewards/WithdrawStakingRewards.tsx
```diff
@@ -6,42 +6,51 @@ import {ScrollView, StyleSheet, Text, View} from 'react-native'
...
-  const [state, setState] = React.useState<
-    {step: 'form'; withdrawalTx: undefined} | {step: 'confirm'; withdrawalTx: YoroiUnsignedTx}
-  >({step: 'form', withdrawalTx: undefined})
+  const {closeModal, openModal} = useModal()
+  const {resetToTxHistory} = useWalletNavigation()
+
+  const handleOnConfirm = async (withdrawalTx: YoroiUnsignedTx) => {
+    closeModal()
+
+    await delay(1000)
```
jorbuedo, line 35: I don't love the artificial delay, is it for aesthetics or does it break without it?
Would it make sense to add a new `switchModal` to useModal, that acts like close+open without it breaking?

michalszorad (approved):

apps/wallet-mobile/src/legacy/Dashboard/WithdrawStakingRewards/WithdrawStakingRewards.tsx
banklesss: I put it because the next dialog opens before the other one closes, forcing an extra back, and it goes to the previous screen, that is, the menu.

stackchain: Definitely the Modal context needs a refactor to deal with an array of modals, instead of a single one.

banklesss: that would be rally good

PR #3652, opened 2024-09-23
feat(wallet-mobile): dev builds automatically invalidate token image on error

jorbuedo:

Seems like a bad thing to do in prod, but should be fine to clear out failing images during dev and qa.

related https://emurgo.atlassian.net/browse/YOMO-1920


stackchain:
I think is a very good idea for prod too.

jorbuedo:
@stackchain there's images that simply will never work, because there's no image. Having thousands of clients spamming invalidate seems wasteful.

stackchain:
> @stackchain there's images that simply will never work, because there's no image. Having thousands of clients spamming invalidate seems wasteful.

yap :/

PR #3663, opened 2024-09-30
feature(wallet-mobile): prefetch ada prices

jorbuedo:
##### Ticket

https://emurgo.atlassian.net/browse/YOMO-1943


jorbuedo:
Double checking cache shenanigans

apps/wallet-mobile/src/features/Portfolio/common/PortfolioProvider.tsx
```diff
@@ -57,6 +63,22 @@ export const PortfolioProvider = ({
...
+    queryFn: async () => {
+      const response = await tokenManager.api.tokenHistory(
+        '279c909f348e533da5808898f87f9a14bb2c3dfbbacccd631d927a3f.534e454b',
+        Portfolio.Token.HistoryPeriod.OneDay,
+      )
+      if (isRight(response)) return true
+      return false
+    },
+  })
+  const isTokenHistoryApiAvailable = data ?? false
```
michalszorad: Maybe refactor for easier read to `useIsTokenHistoryApiAvailable()`

apps/wallet-mobile/src/features/Portfolio/common/hooks/useGetPortfolioTokenChart.ts
```diff
@@ -70,45 +78,49 @@ export const useGetPortfolioTokenChart = (
...
+    queryKey: ['ptPriceHistory', timeInterval],
+    queryFn: ptPriceQueryFn,
+  })
+
+  const ptQuery = useQuery({
     enabled: tokenInfo && isPrimaryToken(tokenInfo.info),
+    staleTime: time.oneMinute,
     ...options,
-    queryKey: ['useGetPortfolioTokenChart', tokenInfo?.info.id ?? '', timeInterval, currency],
+    queryKey: ['useGetPortfolioTokenChart', '.', timeInterval, currency],
```
michalszorad: `primaryTokenId` instead of `.`

apps/wallet-mobile/src/features/Portfolio/common/hooks/useGetPortfolioTokenChart.ts
```diff
@@ -70,45 +78,49 @@ export const useGetPortfolioTokenChart = (
...
-          .map((ticker) => {
-            const value = ticker.prices[validCurrency]
-            if (value === undefined) return undefined
-            const {changePercent, changeValue} = priceChange(initialPrice, value)
-            const label = new Date(ticker.timestamp).toLocaleString(languageCode, {
-              dateStyle: 'short',
-              timeStyle: 'short',
-            })
-            return {label, value, changePercent, changeValue}
+      // force queryFn to be async, otherwise it takes longer and doesn't show isFetching
```
michalszorad, line 101: What do you mean that it takes longer?

apps/wallet-mobile/src/kernel/query-client.ts
```diff
@@ -1,7 +1,24 @@
 import {freeze} from 'immer'
 import {QueryClient} from 'react-query'
 
+import {ptPriceQueryFn, TokenChartInterval} from '../features/Portfolio/common/hooks/useGetPortfolioTokenChart'
+
 const queryClient = new QueryClient()
+
+Object.values(TokenChartInterval).forEach((TokenChartInterval) => {
```
michalszorad: Can we extract this to a fn `prefetchPrimaryTokenPrices`?

apps/wallet-mobile/src/features/Portfolio/common/hooks/useGetPortfolioTokenChart.ts
jorbuedo: Seems like a react-query bug, the fn just takes like half a second to start running.

apps/wallet-mobile/src/features/Portfolio/common/hooks/useGetPortfolioTokenChart.ts
jorbuedo: It's just a static string. Any string as long as it doesn't match any secondary token id would do.
I'll call it `pt` if you don't like the `.`

apps/wallet-mobile/src/kernel/query-client.ts
jorbuedo: Seems a like an unnecessary overhead to declare a function and call it immediately, but sure.

apps/wallet-mobile/src/features/Portfolio/common/hooks/useGetPortfolioTokenChart.ts
michalszorad: Sounds good. Previously we had to replace all `''` with `'.'` which led to a lot of issues, so just wanted to avoid anything similar popping up again

apps/wallet-mobile/src/kernel/query-client.ts
michalszorad: Yeah, I thought the function would be located in a different file - where other price history stuff is, otherwise good

michalszorad (approved):

apps/wallet-mobile/src/kernel/query-client.ts
jorbuedo: The queryFn `ptPriceQueryFn` is on another file. But to run the prefetch itself I need the queryClient, which is either here or inside a hook with useQueryClient.

Come to think of it, I could pass queryClient as paramater from here, and thus move the function, but seems a lot of spreading to multiple files what is just needed once here.

jorbuedo:
@stackchain This contains 2 things:

- Prefeching ada prices so that its graphs don't take 4 seconds to load.
- Hiding non ada charts while the endpoint isn't working, which product asked for.

Is it the method to check for the endpoint being live that you don't like?

stackchain:
> @stackchain This contains 2 things:
> 
> * Prefeching ada prices so that its graphs don't take 4 seconds to load.
> * Hiding non ada charts while the endpoint isn't working, which product asked for.
> 
> Is it the method to check for the endpoint being live that you don't like?

On top of my head, when the feature is actually active for all tokens, for those tokens without data the behaviour would be same, so I believe there is no need for this implementation, yet needs confirmation.

jorbuedo:
> On top of my head, when the feature is actually active for all tokens, for those tokens without data the behaviour would be same, so I believe there is no need for this implementation, yet needs confirmation.

Oh I agree, but it's something product asked for specifically because they wanted to hide the section until the API was ready.
But the 1 point about prefecthing ada prices is independent of that and would just improve perf.

apps/wallet-mobile/src/features/Portfolio/useCases/PortfolioTokenDetails/PortfolioTokenChart/TokenChartToolBar.tsx
```diff
@@ -21,7 +21,7 @@ export const TokenChartToolbar = ({timeInterval, disabled, onChange}: Props) =>
 
   return (
     <View style={styles.chartToolbar}>
-      {Object.values(TOKEN_CHART_INTERVAL).map((itv) => (
+      {Object.values(TokenChartInterval).map((itv) => (
```
stackchain, line 24: I think it is missing translations it'd show `24 H`, `1 W` for all languages.

stackchain: the interval label from TokenPerformance maybe should be moved to `useStrings` 🤷🏻

stackchain (approved):
lg, just 1 point regarding translations, I think only he chart label is affected cc: @jorbuedo

apps/wallet-mobile/src/features/Portfolio/common/hooks/useGetPortfolioTokenChart.ts
```diff
@@ -70,45 +78,49 @@ export const useGetPortfolioTokenChart = (
...
+      await delay(0)
+
+      const tickers = ptPriceQuery?.data ?? []
+      if (tickers.length === 0) throw new Error('No PT price data')
```
stackchain, line 105:
```suggestion
      if (tickers.length === 0) return []
```

apps/wallet-mobile/src/features/Portfolio/useCases/PortfolioTokenDetails/PortfolioTokenChart/TokenChartToolBar.tsx
jorbuedo: Mm yeah, that was never translated.

apps/wallet-mobile/src/features/Portfolio/common/hooks/useGetPortfolioTokenChart.ts
jorbuedo: We're showing the placeholder when fetching is done and data is undefined
```
          {!data ? <ChartPlaceholder /> : <TokenChart onValueSelected={handleChartSelected} dataSources={data} />}
```


If we returned an empty array, it would be used as an actual chart with no points. Plus the warning it shows when it's an error.

apps/wallet-mobile/src/features/Portfolio/useCases/PortfolioTokenDetails/PortfolioTokenChart/TokenChartToolBar.tsx
stackchain: create a task please

PR #3664, opened 2024-09-30
fix(wallet-mobile): offline crashes

jorbuedo:
##### Ticket

[YOMO-1944](https://emurgo.atlassian.net/browse/YOMO-1944)

[YOMO-1944]: https://emurgo.atlassian.net/browse/YOMO-1944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jorbuedo:
@stackchain I assume you don't like this one because it's a provisional solution and those api shouldn't throw in the first place? Offline support isn't expected yet, but at least not crashing would be an improvement.

stackchain:
> @stackchain I assume you don't like this one because it's a provisional solution and those api shouldn't throw in the first place? Offline support isn't expected yet, but at least not crashing would be an improvement.

I believe the issue here is structural on Yoroi, so the problem is deeper, I'd like to avoid short fixes, so far no user has complained about it, I believe these days no one is using it offline, so I'd like to plan this change after 0.74/0.75/0.76 bump.

stackchain:
> > @stackchain I assume you don't like this one because it's a provisional solution and those api shouldn't throw in the first place? Offline support isn't expected yet, but at least not crashing would be an improvement.
> 
> I believe the issue here is structural on Yoroi, so the problem is deeper, I'd like to avoid short fixes, so far no user has complained about it, I believe these days no one is using it offline, so I'd like to plan this change after 0.74/0.75/0.76 bump.

Yet, leave this one open, considering to include on 5.1, since a lot of focus on Anzens, maybe the bump will get delayed.

stackchain (approved):

PR #3695, opened 2024-10-18
fix(wallet-mobile): navigation weirdness in Send

jorbuedo:
##### Ticket
https://emurgo.atlassian.net/browse/YV-30

### DIFF
--- a/apps/wallet-mobile/src/features/Send/common/navigation.ts
+++ b/apps/wallet-mobile/src/features/Send/common/navigation.ts
@@ -1,19 +1,17 @@
 import {useNavigation} from '@react-navigation/native'
 import {useRef} from 'react'
 
-import {
-  AppRouteNavigation,
-  TxHistoryRouteNavigation,
-  TxHistoryRoutes,
-  useOverridePreviousRoute,
-} from '../../../kernel/navigation'
+import {AppRouteNavigation, TxHistoryRouteNavigation} from '../../../kernel/navigation'
 
 export const useNavigateTo = () => {
   const navigation = useNavigation<TxHistoryRouteNavigation & AppRouteNavigation>()
 
   return useRef({
     selectedTokens: () => navigation.navigate('send-list-amounts-to-send'),
-    addToken: () => navigation.navigate('send-select-token-from-list'),
+    addToken: (removePrevious?: boolean) => {
+      if (removePrevious) navigation.pop()
+      navigation.navigate('send-select-token-from-list')
+    },
     startTx: () => navigation.navigate('send-start-tx'),
     editAmount: () => navigation.navigate('send-edit-amount'),
     reader: () => navigation.navigate('scan-start', {insideFeature: 'send'}),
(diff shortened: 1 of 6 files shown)


apps/wallet-mobile/src/kernel/navigation.tsx
```diff
@@ -650,66 +649,3 @@ const getFocusedRouteName = (state: Partial<NavigationState> | NavigationState['
...
-        routeNames: newRoutes.map((r) => r.name),
-      }
-      navigation.reset(newState)
-    })
-
-    return () => task.cancel()
-  }, [navigation, initialRouteId, routesToKeep])
-}
-
-export function useOverridePreviousRoute<RouteName extends string>(previousRouteName: RouteName) {
```
jorbuedo, line 686: It was only used precisely for those Send navigation overrides, and it's buggy which caused the issue.
Turns out adjusting the history state in a hook that executes every render, based on a route name that can change between renders just isn't safe.

apps/wallet-mobile/src/features/Send/useCases/ListAmountsToSend/ListAmountsToSendScreen.tsx
```diff
@@ -76,7 +74,7 @@ export const ListAmountsToSendScreen = () => {
     // use case: redirect to add token screen if there is no token left
     if (selectedTokensCounter === 1) {
       clearSearch()
-      navigateTo.addToken()
+      navigateTo.addToken(true)
```
stackchain: @jorbuedo just for readability can you pass it as object? `{sholdPop: boolean}` something like it

apps/wallet-mobile/src/kernel/navigation.tsx
stackchain: nice catch!

stackchain (approved):

banklesss (approved):

apps/wallet-mobile/src/features/Send/common/navigation.ts
```diff
@@ -1,19 +1,17 @@
...
+import {AppRouteNavigation, TxHistoryRouteNavigation} from '../../../kernel/navigation'
 
 export const useNavigateTo = () => {
   const navigation = useNavigation<TxHistoryRouteNavigation & AppRouteNavigation>()
 
   return useRef({
     selectedTokens: () => navigation.navigate('send-list-amounts-to-send'),
-    addToken: () => navigation.navigate('send-select-token-from-list'),
+    addToken: (removePrevious?: boolean) => {
+      if (removePrevious) navigation.pop()
```
stackchain:
```suggestion
    addToken: ({shouldPopPrevious}: {shouldPopPrevious?: boolean} = {shouldPopPrevious = false}) => {
      if (shouldPopPrevious) navigation.pop()
```

apps/wallet-mobile/src/features/Send/useCases/ListAmountsToSend/ListAmountsToSendScreen.tsx
stackchain:
```suggestion
      navigateTo.addToken({shouldPopPrevious: true})
```

PR #3707, opened 2024-10-24
feature(wallet-mobile): new tx review for withdraw staking rewards

banklesss:

New tx review for withdraw staking rewards feature.

Includes a operations refactor.

##### Ticket

YV-41


apps/wallet-mobile/src/features/ReviewTx/common/operations.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+      case CertificateTypes.StakeDelegation: {
+        const poolKeyHash = (CertificateData as Certificate[CertificateTypes.StakeDelegation]).pool_keyhash ?? null
+        if (poolKeyHash == null) return acc
+        return [...acc, <StakeDelegateOperation key={index} poolId={poolKeyHash} />]
+      }
+
+      case CertificateTypes.VoteDelegation: {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const drep: any = (CertificateData as Certificate[CertificateTypes.VoteDelegation]).drep
```
banklesss: typescripp gets crazy when you mix two shapes.

drep is `'AlwaysAbstain' | 'AlwaysNoConfidence' |  {KeyHash: string} |  {ScriptHash: string}` ans it comes from CSL

jorbuedo: But why is it an issue it has that shape? You can just check for it and it doesn't require the `any`

```
        const drep = (CertificateData as Certificate[CertificateTypes.VoteDelegation]).drep

        if (drep === 'AlwaysAbstain') return [...acc, <AbstainOperation key={index} />]
        if (drep === 'AlwaysNoConfidence') return [...acc, <NoConfidenceOperation key={index} />]

        const drepId = ('KeyHash' in drep ? drep.KeyHash : drep.ScriptHash) ?? ''
```

banklesss: yeah ! forgot about this

jorbuedo (approved):

apps/wallet-mobile/src/features/ReviewTx/common/ReviewTxProvider.tsx
```diff
@@ -152,6 +168,7 @@ type ReviewTxActions = {
   onErrorChanged: (onError: ReviewTxState['onError']) => void
```
stackchain, line 168: is not a good idea to have objects that are not serializable in the state, (eg. if we decide to leverage persisted state this will fail), and context also can quickly become a problem

stackchain:
@banklesss bring it in, but create a task to remove all callbacks out of the state.

apps/wallet-mobile/src/features/ReviewTx/common/ReviewTxProvider.tsx
banklesss: Yeah. It is a temporal solution until the common error screens are implemented

PR #3717, opened 2024-10-30
feature(wallet-mobile): new transaction for discovery (dapps)

banklesss:

Finally, it has been decided to simplify the integration of the new tx review with the dapps so that the UI is ready and the UAT can begin as soon as possible. 

The elimination of the use of unsignedTx, whose work has been started, will be done in the following PRs.

##### Ticket

YV-44


apps/wallet-mobile/src/features/Discover/useDappConnectorManager.ts
```diff
@@ -55,14 +66,18 @@ export const useDappConnectorManager = () => {
...
+            navigateToTxReview({
+              onConfirm: () => {
                 if (!shouldResolve) return
                 shouldResolve = false
-                resolve(tx)
+                signTxWithHW(
+                  {cbor, partial},
+                  {
+                    onSuccess: (signature) => resolve(signature),
+                    onError: (error) => logger.error('ReviewTransaction::handleOnConfirm', {error}),
```
michalszorad: error should be probably passed to `reject`

stackchain (approved):
lgtm, 1 detail pointed by @michaeljscript

PR #3759, opened 2024-12-10
fix(wallet-mobile): address details styling

banklesss:
##### Ticket

YV-159

### DIFF
--- a/apps/wallet-mobile/src/components/Button/Button.tsx
+++ b/apps/wallet-mobile/src/components/Button/Button.tsx
@@ -154,7 +154,6 @@ const useStyles = ({
     container: {
       backgroundColor,
       ...atoms.flex,
-      ...atoms.flex_grow,
       ...atoms.flex_row,
       ...atoms.align_start,
       ...atoms.justify_center,
--- a/apps/wallet-mobile/src/components/Button/Button.tsx
+++ b/apps/wallet-mobile/src/components/Button/Button.tsx
@@ -154,8 +154,8 @@ const useStyles = ({
     container: {
       backgroundColor,
       ...atoms.flex,
-      ...atoms.flex_row,
       ...atoms.flex_grow,
+      ...atoms.flex_row,
       ...atoms.align_start,
       ...atoms.justify_center,
       ...shape,
--- a/apps/wallet-mobile/src/components/Button/Button.tsx
+++ b/apps/wallet-mobile/src/components/Button/Button.tsx
@@ -155,6 +155,7 @@ const useStyles = ({
       backgroundColor,
       ...atoms.flex,
       ...atoms.flex_row,
+      ...atoms.flex_grow,
       ...atoms.align_start,
       ...atoms.justify_center,
       ...shape,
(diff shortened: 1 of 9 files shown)


apps/wallet-mobile/src/components/Button/Button.tsx
```diff
@@ -154,7 +154,6 @@ const useStyles = ({
     container: {
       backgroundColor,
       ...atoms.flex,
-      ...atoms.flex_grow,
```
banklesss: @jorbuedo I checked other buttons and it seems like this change has no negative implications. Maybe I'm missing something.

jorbuedo: Uff, I wouldn't be so sure. Had quite a bit of trouble trying to simplify that so that it worked for all cases, and couldn't get it right. What I do is when I don't want it to grow, wrap it in a View with whatever position you want.

apps/wallet-mobile/src/components/Clipboard/ClipboardProvider.tsx
```diff
@@ -24,12 +24,12 @@ export const ClipboardProvider = ({children}: Props) => {
...
   const [copied, setCopied] = React.useState<CopiedState | null>(null)
   const {height, width} = useWindowDimensions()
-  const copy: ClipboardContext['copy'] = ({text, feedback = 'Copied', event}) => {
+  const copy: ClipboardContext['copy'] = ({text, feedback = 'Copied', event, offsetX = 0, offsetY = 0}) => {
     Clipboard.setString(text)
     setCopied({
       feedback,
-      locationY: event ? event.nativeEvent.pageY - 50 : height * 0.85,
-      locationX: (event?.nativeEvent.pageX ?? width * 0.5) - feedback.length * 4,
+      locationY: (event ? event.nativeEvent.pageY - 50 : height * 0.85) + offsetY,
```
jorbuedo: Seems like a lot of extra props and calculations per usage just to not clip it.
We could calculate here the screen size and put it as a max position, so it doesn't overflow it.
Alternatively, if you don't pass the `event`, the message shows in the middle bottom part of the screen instead of on top of the button.

banklesss: To make the exact calculations you need the feedback text component size before rendering. Looking for a workaround

banklesss: I've tried to get the approximate size of the feedback text in several different ways but it's highly unreliable.

Let me know if you can think of a better solution, if not, I don't think this solution is that problematic TBH

jorbuedo: @banklesss try now

jorbuedo (approved):

apps/wallet-mobile/src/components/Clipboard/ClipboardProvider.tsx
banklesss: yeah it works thanks

PR #3768, opened 2024-12-17
fix(wallet-mobile): collateral dialog closing

banklesss:


apps/wallet-mobile/src/features/Settings/useCases/changeWalletSettings/ManageCollateral/ManageCollateralScreen.tsx
```diff
@@ -95,6 +95,8 @@ export const ManageCollateralScreen = () => {
   const isLoading = isLoadingTx || isLoadingCollateral
 
   const handleGenerateCollateral = () => {
+    closeModal()
```
michalszorad, line 98: We may need to fix also the issue where it generates new collateral even if we already generated collateral before.

michalszorad: Maybe if the collateral already exists not show the modal?

michalszorad (approved):

apps/wallet-mobile/src/features/Settings/useCases/changeWalletSettings/ManageCollateral/ManageCollateralScreen.tsx
banklesss: The button that triggers the modal is hidden when `hasCollateral` is true, do you think any further improvement is needed?

banklesss: ok its approved thanks !

PR #3780, opened 2025-01-07
fix(wallet-mobile): error boundary button layout

banklesss:
##### Ticket

YV-174


apps/wallet-mobile/src/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -51,23 +52,25 @@ const ErrorView = ({state}: {state: State}) => {
 
   return (
     <View style={styles.root}>
+      <Spacer height={70} />
```
jorbuedo, line 55: Do we still have to mix Space and Spacer?

stackchain: yap we do :(, we will be able to get rid of it, only when creating our design system components.

stackchain (approved):

PR #3809, opened 2025-02-03
feat(wallet-mobile): remove trailing zeros from tx list amounts

jorbuedo:

Simplest change with just the trailing zeros removed for the token amount (not the paired balance subtext) in tx list items (tx history and portfolio token tx list).

The "feature" still needed some defining for the rest of the amounts in the app.

##### Ticket

https://emurgo.atlassian.net/browse/YV-132

image

### DIFF
--- a/apps/wallet-mobile/src/yoroi-wallets/utils/format.test.ts
+++ b/apps/wallet-mobile/src/yoroi-wallets/utils/format.test.ts
@@ -37,17 +37,17 @@ describe('formatAda', () => {
 
 describe('formatAdaFractional', () => {
   it('formats zero', () => {
-    expect(formatTokenFractional(asQuantity(0), primaryTokenInfoMainnet)).toEqual('.000000')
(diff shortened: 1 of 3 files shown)


stackchain:
@jorbuedo how about taking only the decimals, and moving it to `@yoroi/common` -> /numbers/fractional-formatter, does it seem doable, dunno about what is consuming it.

jorbuedo:
@stackchain it's only used by TxListItem, and were it is right now it's among plenty of other formatting related things, like the integer part for that same string. So I don't think moving the fractional part to common would be enough. I'd wait to see if there's an actual spec for all formats in the app to move it or refactor everything else.

stackchain:
> @stackchain it's only used by TxListItem, and were it is right now it's among plenty of other formatting related things, like the integer part for that same string. So I don't think moving the fractional part to common would be enough. I'd wait to see if there's an actual spec for all formats in the app to move it or refactor everything else.

makes sense, collocated up to the global definitions

stackchain (approved):

PR #3819, opened 2025-02-10
feature(wallet-mobile): tx review swap cancellation details

banklesss:
**NOTE**: take into consideration this will me adapted to the new swap feature code.

##### Ticket

YV-182

### DIFF
--- a/apps/wallet-mobile/src/features/Swap/common/PoolIcon/PoolIcon.tsx
+++ b/apps/wallet-mobile/src/features/Swap/common/PoolIcon/PoolIcon.tsx
@@ -20,6 +20,7 @@ const icons: Record<Swap.PoolProvider, FunctionComponent<{size?: number}>> = {
   muesliswap_v3: Icon.MuesliSwap,
   muesliswap_v4: Icon.MuesliSwap,
   minswap: Icon.MinSwap,
+  minswapv2: Icon.MinSwap,
   sundaeswap: Icon.SundaeSwap,
   wingriders: Icon.WingRiders,
   vyfi: Icon.VyfiSwap,
--- a/packages/types/src/swap/pool.ts
+++ b/packages/types/src/swap/pool.ts
@@ -2,6 +2,7 @@ import {PortfolioTokenId} from '../portfolio/token'
 
 export type SwapPoolProvider =
   | 'minswap'
+  | 'minswapv2'
   | 'sundaeswap'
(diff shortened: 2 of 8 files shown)


michalszorad (approved):

apps/wallet-mobile/src/features/Swap/common/PoolIcon/PoolIcon.tsx
```diff
@@ -20,6 +20,7 @@ const icons: Record<Swap.PoolProvider, FunctionComponent<{size?: number}>> = {
   muesliswap_v3: Icon.MuesliSwap,
   muesliswap_v4: Icon.MuesliSwap,
   minswap: Icon.MinSwap,
+  minswapv2: Icon.MinSwap,
```
stackchain, line 23: it looks like the convention here is _v

banklesss: is coming in this way from the API 🧐

packages/types/src/swap/pool.ts
```diff
@@ -2,6 +2,7 @@ import {PortfolioTokenId} from '../portfolio/token'
 
 export type SwapPoolProvider =
   | 'minswap'
+  | 'minswapv2'
```
stackchain, line 5: v2 is a protocol, why add it as another pool ?

stackchain: you can get it by the getByProtocol

apps/wallet-mobile/src/features/Swap/common/PoolIcon/PoolIcon.tsx
stackchain: hmm, as `MINSWAPV2` ?

stackchain: maybe it should use Swap.Protocol, and not Swap.Provider.

stackchain: pool provider will die.

stackchain:
@banklesss sorry, I'm mixing new updates with old ones. :/

stackchain:
> @banklesss sorry, I'm mixing new updates with old ones. :/

I thought it was on swap/refactor branch not in the develop

PR #3839, opened 2025-02-19
fix(wallet-mobile): disable gestures for modals

jorbuedo:

Seems like modals in iOS are still able to be closed by a swipe down, not sure how. Not sure if this would prevent it but can't hurt to try.

##### Ticket

https://emurgo.atlassian.net/browse/YV-240?focusedCommentId=30503


apps/wallet-mobile/src/AppNavigator.tsx
```diff
@@ -202,12 +202,19 @@ export const AppNavigator = () => {
...
                   ...(Platform.OS === 'android' && {...TransitionPresets.DefaultTransition}), // overriding general navigation settings
                   cardStyle: {backgroundColor: 'transparent'}, // this is needed for the modal to be transparent
                 }}
               >
-                <Stack.Screen name="modal" component={ModalScreen} />
+                <Stack.Screen
+                  name="modal"
+                  component={ModalScreen}
+                  options={{
+                    gestureEnabled: false,
```
michalszorad, line 215: Should this be disabled for all modals, not just disclaimers?

jorbuedo: Mmm, I don't know. The screen itself already has an outside click or swipe down from the header to close in case it's allowed. So I think it should be fine to remove all native gestures.
Not sure how we would add this only for the non discardable modals, unless we divided them to different screens... I think it's fine this way.

michalszorad (approved):

PR #3842, opened 2025-02-20
feat(swap): Remotely configure buying side of swap

jorbuedo:

This will need to be ported into the swap refactor branch as well, not necessarily in the same place.

##### Ticket

https://emurgo.atlassian.net/browse/YV-256

### DIFF
--- a/packages/swap/package.json
+++ b/packages/swap/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@yoroi/swap",
-  "version": "2.0.1",
+  "version": "2.1.0",
   "description": "The Swap package of Yoroi SDK",
   "keywords": [
     "yoroi",
--- a/packages/swap/package.json
+++ b/packages/swap/package.json
@@ -119,7 +119,7 @@
     ],
     "coverageThreshold": {
       "global": {
-        "branches": 100,
+        "branches": 99,
         "functions": 100,
         "lines": 100,
         "statements": 100
--- a/packages/swap/package.json
+++ b/packages/swap/package.json
@@ -119,7 +119,7 @@
     ],
     "coverageThreshold": {
       "global": {
-        "branches": 99,
+        "branches": 100,
         "functions": 100,
         "lines": 100,
         "statements": 100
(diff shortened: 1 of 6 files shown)


packages/swap/package.json
```diff
@@ -119,7 +119,7 @@
     ],
     "coverageThreshold": {
       "global": {
-        "branches": 100,
+        "branches": 99,
```
michalszorad: You can keep it 100% if you make the params of `getSwapConfig` required instead of them being optional

stackchain:
```suggestion
        "branches": 100,
```

jorbuedo: But I don't want them required. I don't want to have to pass fetchData when I use this.

michalszorad:
```ts
it('should not require passing a param', () => {
    expect(getSwapConfig()).toBeDefined()
  })
```

jorbuedo: That would send a remote request to the file, which I don't want.

jorbuedo: The difference with the one in dapps is the factory. I would have to instead of having a simple function, have a function that returns a function, just so I can test that the factory works without actually calling the function inside. It's just stupid.

packages/swap/src/adapters/getSwapConfig.ts
```diff
@@ -1,33 +1,34 @@
...
-  request: FetchData
-} = initialDeps): Promise<SwapConfig> => {
-  const response = await request<SwapConfig>({
-    url: 'https://daehx1qv45z7c.cloudfront.net/swapConfig.json',
-  })
+    if (isLeft(response)) throw getApiError(response.error)
 
-  if (isLeft(response)) throw getApiError(response.error)
+    if (!SwapConfigResponseSchema.safeParse(response.value.data).success) {
+      throw new Error(
```
stackchain: `Api.Errors.ResponseMalformed`

packages/swap/src/adapters/getSwapConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+type SwapConfig = z.infer<typeof SwapConfigResponseSchema>
+
+export const getSwapConfigApiMaker =
+  (
+    {
+      request,
+    }: {
+      request: FetchData
+    } = {request: fetchData},
+  ) =>
```
michalszorad: The `const initialDeps = freeze({request: fetchData}, true)` made it a bit more clear to read. Maybe having a type for the options `type Options = {request:  FetchData}` would make it more readable too

stackchain (approved):

PR #3843, opened 2025-02-21
fix(wallet-mobile): Fix address path for message signing and TX review message parsing

michalszorad:

Fix address path for message signing and TX review message parsing

##### Ticket

[YV-245](https://emurgo.atlassian.net/browse/YV-245)

[YV-245]: https://emurgo.atlassian.net/browse/YV-245?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/wallet-mobile/src/features/ReviewTx/common/hooks/useFormattedMetadata.tsx
+++ b/apps/wallet-mobile/src/features/ReviewTx/common/hooks/useFormattedMetadata.tsx
@@ -43,16 +43,13 @@ export const formatMetadata = async (
 }
 
 const parseMsg = (msg: Array<string> | string): string => {
-  if (Array.isArray(msg)) {
-    const message = msg.join('')
-    try {
-      return JSON.parse(message)
-    } catch {
-      return message
-    }
-  }
+  const messageToParse = Array.isArray(msg) ? msg.join('') : msg
 
-  return msg
+  try {
+    return JSON.parse(messageToParse)
+  } catch {
+    return messageToParse
+  }
 }
 
 export const useFormattedMetadata = ({
--- a/apps/wallet-mobile/src/features/ReviewTx/common/hooks/useFormattedMetadata.tsx
+++ b/apps/wallet-mobile/src/features/ReviewTx/common/hooks/useFormattedMetadata.tsx
@@ -1,4 +1,5 @@
 import {MetadataJsonSchema} from '@emurgo/cross-csl-core'
+import {isString} from '@yoroi/common'
 import _ from 'lodash'
 import {useQuery} from 'react-query'
 
@@ -46,7 +47,9 @@ const parseMsg = (msg: Array<string> | string): string => {
   const messageToParse = Array.isArray(msg) ? msg.join('') : msg
 
   try {
-    return JSON.parse(messageToParse)
+    const parsed = JSON.parse(messageToParse)
+    if (isString(parsed)) return parsed
+    return JSON.stringify(parsed)
   } catch {
     return messageToParse
   }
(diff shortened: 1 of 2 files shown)


apps/wallet-mobile/src/features/ReviewTx/common/hooks/useFormattedMetadata.tsx
```diff
@@ -42,7 +42,11 @@ export const formatMetadata = async (
   }
 }
 
-const parseMsg = (msg: Array<string>) => {
+const parseMsg = (msg: Array<string> | string) => {
+  if (typeof msg === 'string') {
+    return msg
+  }
+
   if (msg.length > 1) {
```
stackchain: I think is misleading, maybe make it more declarative, -> `if Array.isArray` ..., first, and then fall back to string, cuz [].join still works, unless is needed to check if it is an array and there is at least 1 element.

michalszorad: Yes, it seems that there are different rules when the length is 1 and when the length is > 1. I'll do a small refactor and ping you to check again if it looks better. It may be good to hear @banklesss thoughts

stackchain: maybe something like `const string2sign = Array.isArray(msg) ? msg.join() : msg`, states clear that signing expects a string only

michalszorad: Note this is not used for signing, just displaying on the TX review screen

michalszorad: I've changed it to the following
```ts
const messageToParse = Array.isArray(msg) ? msg.join('') : msg

  try {
    return JSON.parse(messageToParse)
  } catch {
    return messageToParse
  }
```

apps/wallet-mobile/src/features/ReviewTx/common/hooks/useFormattedMetadata.tsx
```diff
@@ -42,20 +42,21 @@ export const formatMetadata = async (
...
-  if (typeof msg === 'string') {
-    return msg
-  }
-
-  if (msg.length > 1) {
+const parseMsg = (msg: Array<string> | string): string => {
+  if (Array.isArray(msg) && msg.length > 1) {
     const message = msg.join('')
     try {
       return JSON.parse(message)
```
stackchain: ~parsing a string? shouldn't it be a stringily, im confused here.~

stackchain: aaa nvm, got it, if it's a valid JSON return as JSON otherwise as string, makes sense now.

stackchain (approved):

apps/wallet-mobile/src/features/ReviewTx/common/hooks/useFormattedMetadata.tsx
```diff
@@ -42,16 +42,14 @@ export const formatMetadata = async (
...
 
-const parseMsg = (msg: Array<string>) => {
-  if (msg.length > 1) {
-    const message = msg.join('')
-    try {
-      return JSON.parse(message)
-    } catch {
-      return message
-    }
+const parseMsg = (msg: Array<string> | string): string => {
```
stackchain, line 46: @michaeljscript -> I know what is bugging me, the return type can be whatever, --> JSON.parse('1') = `number` not string, maybe parse, and then stringify after? :dunno:

michalszorad: Yeah good point

michalszorad:
```ts
const parsed = JSON.parse(messageToParse)
if (isString(parsed)) return parsed
return JSON.stringify(parsed)
```

PR #3846, opened 2025-02-24
chore(wallet-mobile): drop disable status for action buttons

stackchain:
##### Ticket

YV-260

### DIFF
--- a/apps/wallet-mobile/src/features/Transactions/useCases/TxHistory/ActionsBanner.tsx
+++ b/apps/wallet-mobile/src/features/Transactions/useCases/TxHistory/ActionsBanner.tsx
@@ -3,7 +3,7 @@ import {useSwap} from '@yoroi/swap'
 import {useTheme} from '@yoroi/theme'
 import {useTransfer} from '@yoroi/transfer'
 import {Chain} from '@yoroi/types'
-import React from 'react'
+import * as React from 'react'
 import {GestureResponderEvent, StyleSheet, View} from 'react-native'
 
 import {Button, ButtonType} from '../../../../components/Button/Button'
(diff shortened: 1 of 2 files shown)


apps/wallet-mobile/src/features/Transactions/useCases/TxHistory/ActionsBanner.tsx
```diff
@@ -22,7 +22,7 @@ import {useSelectedWallet} from '../../../WalletManager/common/hooks/useSelected
 import {useWalletManager} from '../../../WalletManager/context/WalletManagerProvider'
 import {useStrings} from '../../common/strings'
 
-export const ActionsBanner = ({disabled = false}: {disabled: boolean}) => {
+export const ActionsBanner = () => {
```
stackchain, line 25: disabled was added for the first load, since now it doesnt open the wallet before a proper sync (exception for dev build), it is no longer needed.

michalszorad (approved):

apps/wallet-mobile/src/features/Transactions/useCases/TxHistory/ActionsBanner.tsx
jorbuedo: What happened with the check for pool transition?

stackchain: is it part of tx loading state?

stackchain: I see two things here, 1st possible a clash with another modal if the user open whatever right after loading the UI, but this would happen anyway if any other action has a modal, 2nd all pools were already migrated, so no worries (for now), the 1st issue will be solved only after building a queue for the modal.

PR #3849, opened 2025-02-25
feat(wallet-mobile): Add FCM Token to About screen

jorbuedo:
image

##### Ticket
https://emurgo.atlassian.net/browse/YV-247

### DIFF
--- a/apps/wallet-mobile/src/features/Settings/useCases/changeAppSettings/About/About.tsx
+++ b/apps/wallet-mobile/src/features/Settings/useCases/changeAppSettings/About/About.tsx
@@ -1,18 +1,27 @@
+import messaging from '@react-native-firebase/messaging'
 import {useNavigation} from '@react-navigation/native'
 import {useTheme} from '@yoroi/theme'
 import React from 'react'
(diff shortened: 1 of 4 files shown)


apps/wallet-mobile/src/features/Settings/useCases/changeAppSettings/About/About.tsx
```diff
@@ -29,6 +38,18 @@ export const About = () => {
 
         <ValueText>{commit}</ValueText>
       </Row>
+
+      {FCMToken && (
+        <Row>
+          <LabelText>{`FCM Token`}</LabelText>
```
michalszorad: Translations

jorbuedo: Is that something that we want to translate? 🤔

michalszorad: Maybe, we have translations for almost everything, so not sure why this would be an exception

michalszorad (approved):

PR #3852, opened 2025-02-26
fix(wallet-mobile): Fix tx review accordion and send loading button

jorbuedo:
##### Ticket

https://emurgo.atlassian.net/browse/YV-263

### DIFF
--- a/apps/wallet-mobile/src/features/ReviewTx/common/Accordion.tsx
+++ b/apps/wallet-mobile/src/features/ReviewTx/common/Accordion.tsx
@@ -32,8 +32,8 @@ export const Accordion = ({label, children}: {label: string; children: React.Rea
           styles.childrenContainer,
           {
             maxHeight: animatedHeight.interpolate({
-              inputRange: [0, 1],
-              outputRange: [0, 1000],
+              inputRange: [0, 0.9, 1],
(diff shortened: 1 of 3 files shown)


apps/wallet-mobile/src/features/ReviewTx/common/Accordion.tsx
```diff
@@ -32,8 +32,8 @@ export const Accordion = ({label, children}: {label: string; children: React.Rea
           styles.childrenContainer,
           {
             maxHeight: animatedHeight.interpolate({
-              inputRange: [0, 1],
-              outputRange: [0, 1000],
+              inputRange: [0, 0.9, 1],
+              outputRange: [0, 1000, 9999999], // can't mix numbers and strings, and appending % to the result is crashing the app, so just use a BIG number
```
michalszorad, line 36: Will bigger number mean a faster animation?

stackchain (approved):

apps/wallet-mobile/src/features/ReviewTx/common/Accordion.tsx
jorbuedo: That's why I added another step, so from 0 to 0.9 we interpolate the same as before, leaving the animation similar, and then the last 0.1 goes to the big number.

michalszorad (approved):

PR #3856, opened 2025-03-03
fix(wallet-mobile): Use explorer data when pool info is missing

jorbuedo:

Proposal for fallback pool info using explorer data.

##### Ticket

https://emurgo.atlassian.net/browse/YV-275


stackchain:
i think is a great idea, @jorbuedo can u update the router to use getComponent if it is using component, to lazy load it, we need to do a mass review on the navigator is loading all screens mostly in one shot

stackchain (approved):

PR #3859, opened 2025-03-07
fix(wallet-mobile): dapps collateral request

banklesss:
##### Ticket

YV-192


apps/wallet-mobile/src/yoroi-wallets/cardano/cip30/cip30.ts
```diff
@@ -196,7 +190,7 @@ class CIP30Extension {
     }
   }
 
-  async sendReorganisationTx(cbor: string, witnesses: TransactionWitnessSet): Promise<CSL.TransactionUnspentOutput> {
+  async sendReorganisationTx(cbor: string, witnesses: TransactionWitnessSet): Promise<void> {
```
michalszorad: Is this method still needed?

apps/wallet-mobile/src/features/Swap/useCases/StartOrderSwapScreen/ListOrders/OpenOrders.tsx
```diff
@@ -728,17 +732,19 @@ const NoOrdersYet = () => {
   )
 }
 
-const useShowCollateralNotFoundAlert = (wallet: YoroiWallet) => {
+export const useShowCollateralNotFoundAlert = (
+  wallet: YoroiWallet,
+  onCollateralNotFoundPress?: () => void,
+  onCollateralPendingPress?: () => void,
```
michalszorad: Can we use options `{wallet, onCollateralNotFoundPress, onCollateralPendingPress}`?

apps/wallet-mobile/src/features/Discover/useDappConnectorManager.tsx
```diff
@@ -214,3 +271,42 @@ const useSignDataWithHW = () => {
     [confirmHWConnection, wallet, meta, closeModal],
   )
 }
+
+export const useDappCollateralRequestUtils = (wallet: YoroiWallet) => {
+  const [dappIds, setDappsIds] = React.useState<Array<string>>([])
```
michalszorad, line 276: Why do we need to track which dApps requested collateral?

apps/wallet-mobile/src/features/Discover/useDappConnectorManager.tsx
```diff
@@ -39,12 +43,33 @@ export const useDappConnectorManager = () => {
...
       return new Promise<string>((resolve, reject) => {
         let shouldResolve = true
-        return manager.getDAppList().then(({dapps}) => {
+        return manager.getDAppList().then(async ({dapps}) => {
+          const dappsConnected = await manager.listAllConnections()
+          const matchingDappConnection =
+            activeTabOrigin != null ? dappsConnected.find((dapp) => dapp.dappOrigin.includes(activeTabOrigin)) : null
+
+          if (matchingDappConnection?.dappOrigin != null) {
+            const isDappRequestingCollateral = dappCollateralRequestUtils.getIsDappRequestingCollateral(
```
michalszorad, line 54: Is the idea to block the user from signing only for the specific dapp which requested collateral? If yes, does it handle the case when I have 2 tabs open and want to place 2 different TXs - one that requires collateral and 2nd that does not require collateral?

apps/wallet-mobile/src/yoroi-wallets/cardano/cip30/cip30.ts
banklesss: need to be removed yes!

apps/wallet-mobile/src/features/Discover/useDappConnectorManager.tsx
banklesss: It is to be able to identify and reject the request of the parallel tx related to the collateral tx. both tx are triggered simultaneously

banklesss: > one that requires collateral and 2nd that does not require collateral?

extreme edge case, but yes maybe we need to add extra identifier

banklesss: thanks!

banklesss:
converted to draft to apply changes and wait for copy changes

michalszorad (approved):

jorbuedo (approved):

PR #3862, opened 2025-03-11
feature(banners): banners package

stackchain:
##### Ticket

YV-284


apps/wallet-mobile/src/kernel/constants.ts
```diff
@@ -77,3 +77,6 @@ export const time = freeze({
   // session here means while the wallet is open
   session: Infinity,
 })
+
+// NOTE: bech32 'drep1ygr9tuapcanc3kpeyy4dc3vmrz9cfe5q7v9wj3x9j0ap3tswtre9j'
+export const yoroiDRepIdHex = '220655f3a1c76788d839212adc459b188b84e680f30ae944c593fa18ae'
```
stackchain, line 82: @michaeljscript the endpoint is returning the hash of the vote drep delegation.

apps/wallet-mobile/ios/Podfile.lock
```diff
@@ -153,15 +153,9 @@ PODS:
   - libwebp/sharpyuv (1.3.2)
   - libwebp/webp (1.3.2):
     - libwebp/sharpyuv
-<<<<<<< Updated upstream
```
michalszorad, line 156: 👍

packages/banners/src/translators/reactjs/useBanner.ts
```diff
@@ -0,0 +1,21 @@
+import {Banners} from '@yoroi/types'
+import {freeze} from 'immer'
+import * as React from 'react'
+
+export function useBanner<K extends string = string>({
+  id,
+  manager,
```
michalszorad, line 7: Is it not better to use context to access manager?

michalszorad (approved):

packages/banners/src/translators/reactjs/useBanner.ts
stackchain: for testing it is not, this also decouple the context from the model, for the new hooks they are receiving only the function needed which is even simpler, here it would need at least 2 so for now is receiving the entire module.

PR #3881, opened 2025-03-24
fix(portfolio): further toTokenActivity breaking

jorbuedo:
##### Ticket

https://emurgo.atlassian.net/browse/YV-306


packages/portfolio/src/adapters/dullahan-api/transformers.ts
```diff
@@ -63,7 +63,8 @@ export const toTokenActivity = (
       const [statusCode, tokenActivityData] = response
       if (statusCode !== Api.HttpStatusCode.Ok) return acc
 
-      TokenActivityResponseSchema.parse(tokenActivityData)
+      if (!TokenActivityResponseSchema.safeParse(tokenActivityData).success)
```
stackchain: I'm thinking about it, if we switch to a optimistic approach, maybe we should update all ? 🤔

jorbuedo: I don't know right now if the other endpoints benefit from throwing or not, but this one being `multi` I never want the whole thing killed because 1 of the entries was malformed. Just discard that one and move on.

But yeah, probably the others could use the same treatment

jorbuedo: We could keep the general malformed error on the api if we wanted for the case where all of them are malformed, like

```
if (entries.length > 0 && transformedEntries.length === 0) return {tag: 'left' ...}
```

But I'm not sure that provides any value either, when I call this from the app I expect the results and I'm fine with 0 results.

stackchain:
### TODO (portfolio only for now)

- [ ] add the logger as dependency
- [ ] fix other requests to be optmistic
- [ ] log problems when parsing with logger.error (will send to sentry if sharing crashes)

PR #3891, opened 2025-04-02
chore(wallet-mobile): Refactor wallet transactions getters

michalszorad:

When working on the notifications feature I noticed that whenever we access the `wallet.transactions` we re-parse the metadata which slows down the app considerably. The refactoring of some getters in Yoroi wallet should lead to better performance.


banklesss (approved):

stackchain:
we should blow this thing up and write a proper model.

PR #3979, opened 2025-07-15
rn79: modal first version

banklesss:

First modal version. 

Discord's react-native-bottom-sheet is a fork of https://github.com/gorhom/react-native-bottom-sheet and this one is currently being maintained. Both repos are controlled by the same user.

For now, the structure and usability of the old modal have been very simply imitated. The implementation will evolve based on needs and the new possibilities offered by the new modal.

the dev menu has been recovered

##### Ticket

YV-505


stackchain:
@banklesss this will change a lot the clients, yet is fine, lets try to leverage a pseudo Portal to move the Modal out of react tree, and the `canDiscard` is a very important option for when clicking outside, on blue-sky they have brought the code internal instead of adding as a package, maybe we should consider do the same for some tweaks, for now is good, only the `canDiscard` is relevant, and eventually a queue for the modals.

PR #3998, opened 2025-07-31
Update resolver package

folarinemurgo:
Bumped up Bumped up cross-csl-core and cross-csl-nodejs to 8.0 alpha and react-query to 5.81.5

##### Ticket
YV-465

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependencies to the latest versions.
  * Incremented package version for maintenance release.


stackchain:
can you confirm the new sync is working? also any use of react-query is no longer needed, should be replaced by a try - catch.

PR #4039, opened 2025-08-14
Fix UI of initial screen

rahulnr7:
Before -> After

Screenshot 2025-08-14 at 11 19 42 AM


apps/mobile/src/ui/YoroiLogo/YoroiLogo.tsx
```diff
@@ -23,7 +23,7 @@ export const YoroiLogo = () => {
       <Space.Height.sm />
 
       <Text style={{...a.body_2_md_regular, color: p.gray_900}}>
```
stackchain: @rahulnr7 some atoms are "themed" you can check in the .cursor ... you will have to import `const {atoms: ta} = useTheme()`, they come from the theme hook and not from the static atoms just FIY, mainly background/text.
```suggestion
      <Text style={[ta.body_2_md_regular, color: p.gray_900]}>
```

stackchain (approved):

PR #4081, opened 2025-08-21
doc: Adds TODO for bugifx

jorbuedo:
This is a bug in production


banklesss (approved):

apps/mobile/src/wallets/cardano/common/signatureUtils.ts
```diff
@@ -210,6 +210,7 @@ const needsToSignWithStakingKey = (tx: CSL_TYPES.Transaction) => {
     if (certificate.asVoteRegistrationAndDelegation()?.hasValue()) return true
   }
 
+  // TODO: REVISIT This needs to check if the withdrawal matches our wallet, aribitrary tx from dapps can have withdrawals on other wallets and we don't want to sign them with our staking key. In this case Anzens is failing.
```
stackchain: @lisicky AFAIK if we pass partial to the sign it will only look for request signatures that we can actually sign without throwing or complaining if there is none, is it correct ?

PR #4091, opened 2025-08-26
fix(mobile): signatures

stackchain:


mobile/app.config.js
banklesss, line 1: nightly is used in two config files 🤔

stackchain: this config is overwritten, yet it should be DEV, previously dev = prod. let me revert for prod icons / bundler

banklesss (approved):

PR #4107, opened 2025-08-29
Fixed multi token send crash

folarinemurgo:
Fixed multi token send crash
Updated strings

##### Ticket
YV-525

### DIFF
--- a/mobile/src/features/Send/useCases/ListAmountsToSend/EditAmount/EditAmountScreen.tsx
+++ b/mobile/src/features/Send/useCases/ListAmountsToSend/EditAmount/EditAmountScreen.tsx
@@ -52,19 +52,8 @@ export const EditAmountScreen = () => {
     targets,
   } = useTransfer()
 
-  if (
-    !targets ||
-    selectedTargetIndex < 0 ||
-    selectedTargetIndex >= targets.length
-  ) {
-    return null
-  }
-
-  if (!selectedTokenId) {
-    return null
-  }
-
-  const amount = targets[selectedTargetIndex]?.entry?.amounts?.[selectedTokenId]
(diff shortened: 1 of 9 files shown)


mobile/src/features/Send/useCases/ListAmountsToSend/EditAmount/EditAmountScreen.tsx
```diff
@@ -119,23 +127,40 @@ export const EditAmountScreen = () => {
...
+  if (!selectedTokenId) {
+    return null
+  }
+
+  if (!amount) {
+    return null
+  }
 
   return (
     <KeyboardAvoidingView style={[a.flex_1, {backgroundColor: p.bg_color_max}]}>
```
stackchain:
```suggestion
    <KeyboardAvoidingView style={[a.flex_1, ta.bg_color_max]}>
```

stackchain:
@folarinemurgo this chain of optional amount complicates the logic, my takeaway here is, go to the place that is navigating to the EditAmount and only navigate if the target -> amount is present, pass it as navigation param and useSafeParams on the other side as props, this will eliminate all optional chains and also the useEffect to navigate directly to select, cuz it can be tested in the even handler

folarinemurgo:
> @folarinemurgo this chain of optional amount complicates the logic, my takeaway here is, go to the place that is navigating to the EditAmount and only navigate if the target -> amount is present, pass it as navigation param and useSafeParams on the other side as props, this will eliminate all optional chains and also the useEffect to navigate directly to select, cuz it can be tested in the even handler

@stackchain Pushed an updated

stackchain (approved):

PR #4110, opened 2025-08-29
Fix empty search results screen for swap tokens

rahulnr7:
https://emurgo.atlassian.net/browse/YV-576


mobile/src/ui/Analytics/Analytics.tsx
```diff
@@ -83,11 +81,11 @@ const Notice = ({
             }}
             title={strings.ui.skip}
           />
+
+          <Space.Height.lg />
```
stackchain, line 85: @rahulnr7 I just realized that I'm breaking the JSX rules by having a lowercase component Q.Q `.lg` maybe we need to refactor all usage of Space.Height / Width to something like, H_lg, W_lg, so on and so forth Q.Q

stackchain (approved):

PR #4124, opened 2025-09-04
fix(mobile): modal avoiding keyboard

banklesss:
Working solution. Styling needs a review. Its only applied to tx review at the moment.

Future modals with inputs **NEED TO USE THE CUSTOM MODAL TEXT INPUT.**


mobile/src/ui/Modal/TextInput.tsx
banklesss, line 1: This is a candidate to replace ui folder TextInput and thus avoid React Native paper.

stackchain: @banklesss im not comfortable with that is mixing 2 things in one, IMHO let's refactor the current INPUT to fix it, and then wrap the ModalScreen with the react-native-avoid-keyboard (dunno the exact name)

banklesss:
refactoring

banklesss:
Closing in favour of https://github.com/Emurgo/yoroi/pull/4163

PR #4125, opened 2025-09-04
fix(swap): Add muesliswap partner to limit quote and create

jorbuedo:
@stackchain how are we doing the publish to npm now? Is it manual or action?


SorinC6 (approved):

stackchain:
> @stackchain how are we doing the publish to npm now? Is it manual or action?

is manual, need to jump in the scripts -> run buld_packages -> publish them (the ones that have changed) according the dependency tree

stackchain (approved):

PR #4141, opened 2025-09-09
fix(mobile): missing send funnel banner

banklesss:
https://emurgo.atlassian.net/browse/YV-595


mobile/src/features/Send/useCases/StartMultiTokenTx/StartMultiTokenTxScreen.tsx
```diff
@@ -106,6 +107,8 @@ export const StartMultiTokenTxScreen = () => {
         >
           <ShowErrors />
 
+          <NotifySupportedNameServers />
```
stackchain, line 110: AFAIK it was asked to drop it.

PR #4157, opened 2025-09-12
fix screen share crash

jorbuedo:
https://emurgo.atlassian.net/browse/YV-593

Screen sharing enable/disable, which was only in Android, seems to be broken.
This will at least prevent the crash, and disable it for non dev. 
Making it work will need more work down the line.


banklesss (approved):

stackchain:
-> closed, I incorporated in my PR.

PR #4188, opened 2025-09-18
fix swap card and padding

rahulnr7:
https://emurgo.atlassian.net/browse/YV-640


jorbuedo:
https://github.com/Emurgo/yoroi/pull/4189

PR #4203, opened 2025-09-19
fix(mobile): Invalidate notifications when removing one

jorbuedo:


stackchain (approved):

stackchain:
@jorbuedo dunno why some preprod `.remove` are missing `await` since is not mainnet i'm not that concern, dunno if Michal did on purpose, it looks like it is just missing

PR #4232, opened 2025-09-25
fix(common, mobile): Refactor swap and send number parsing with a new common parser

jorbuedo:
Fixes https://emurgo.atlassian.net/browse/YV-661

- Creates common parseNumberFromText with support for everything text to sanitized, number, format and quantity, with locale and keyboard mismatch for decimals
- Refactor swap and send number parsing with a new common parser
- Use system locale for number formats
- Formatting with BigNumber uses system locale

---

> [!NOTE]
> Adds a new locale-aware number parser and switches swap/send flows and formatting to it, using system number locale throughout.
> 
> - **Common**:
>   - **New Parser**: Add `utils/parse-number-from-text` with locale-aware sanitization, precision, numeric value, and atomic `quantity` output; expose via `@yoroi/common` and update `to-bigint` docs.
>   - **Exports**: Reorder/adjust `index.ts` exports; remove legacy `numbers/parse-decimal` and its tests.
> - **i18n/Localization**:
>   - Use system number locale (`getSystemNumberLocale`) and configure `BigNumber` `FORMAT` accordingly.
>   - Improve language resolution with base-language fallback.
> - **Wallet Utils**:
>   - Replace `Quantities.parseFromText` with `parseNumberFromText`; update tests and expectations.
> - **Swap**:
>   - Parse and sanitize all numeric inputs (amounts, wanted price, slippage) with `parseNumberFromText`.
>   - Use parsed `numericValue`/`quantity` for API calls, balance checks, and estimate propagation.
>   - Locale-format numeric values in `EstimateSummary` and `TransactionSummary`.
> - **Send (Edit Amount)**:
>   - Replace custom input with `TextInput`, auto-select-on-focus, and use `atomicBreakdown(...).str`; tweak UI (buttons/divider).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b2be61bd3da43858c7d069a08db9ede6c5ac363e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mobile/src/features/Swap/common/SwapProvider.tsx
```diff
@@ -334,13 +338,19 @@ export const SwapProvider = ({children}: {children: React.ReactNode}) => {
...
+                wantedPrice: parseNumberFromText({
+                  text: state.wantedPrice,
+                }).numericValue,
               }),
             }
           : {
-              amountOut: Number(state.tokenOutInput.value),
+              amountOut: parseNumberFromText({
+                text: state.tokenOutInput.value,
+              }).numericValue,
```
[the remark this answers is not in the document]
jorbuedo: No it doesn't.

rahulnr7 (approved):
Good to merge. 
Update - further changes require to handle locales from OS

jorbuedo:
@stackchain formatting seems too work now when parsing inputs and most of where we display them. An exception to this is swap, because lot of numbers go from the aggregators api to the UI without formatting them first. That makes the swap screen mix commas and dots for decimals if the user's system uses commas. Should we format swap numbers as well, or not format the input there?

mobile/packages/common/utils/parse-number-from-text.ts
stackchain, line 1: keep kebab case for packages - non react stuff

mobile/packages/common/utils/parse-number-from-text.ts
```diff
@@ -0,0 +1,211 @@
...
+    throw new Error('Invalid quantity')
+  }
+  return bn.toString(10) as Balance.Quantity
+}
+
+export type ParseNumberFromTextOptions = {
+  /** The input text to parse */
+  text: string
+  /** Token denomination (e.g., 6 for ADA) - optional for unitless numbers */
+  denomination?: number
```
stackchain, line 18: there is no need to be optional, make it mandatory

jorbuedo: This function can be used for number inputs that are not tokens, or not converted to base units, like slippage, wanted price, inputs and outputs of swap (since we're using decimals right now). Forcing adding denomination is confusing. The way it is right now, if denomination is undefined in the input, quantity is undefined in the output.

stackchain: yap, it looks like it should be split, asQuantity always take denomination while (the former to-decimal) was for UI mainly, maybe it should be specialized so we dont end up always checking for possible undefined 🤔

jorbuedo: I don't see why we need to split it, it's basically a lot of the same logic, it just can be used for simple numbers, and optionally to auto give you a Quantity if you provide a denomination.
The Quantities utils in mobile still uses 
```
  parseFromText: (
    text: string,
    denomination: number,
    format: Numbers.Locale,
    precision = denomination,
  ): [string, Balance.Quantity] => {
  ```
  
  So if you're using that you still have denomination required and a non-undefined output.

jorbuedo: So to summarize, the main functionality for the function is:

Text => Sanitized text, used for inputs where the user is the one providing the initial Text
+ Numeric value always available

Optionally, you can pass inputs to get outputs:

format => Formatted value. Unlike sanitized text, this one formats thousand separators too and removes trailing comas, etc. This one is for showing, not for inputs.
denomination => Quantity, based on the denomination it calculates the integer value, truncating smaller decimals.

More optionally, the last input is:

precision => It limits the number of decimals the user can write. Defaults to denomination if exists, or 12 if it doesn't. They are kind of related because usually if you're typing Ada with denomination 6, you want to also allow 6 decimals. But it's not necessarily tied to it, you may want to limit slippage input to 2 decimals, and it has nothing to do with denomination. You could even limit the decimals of an Ada input to just 3, and the least 3 lovelace digits would be forced to be 000.

jorbuedo:
Added formatting to the rest of the swap numbers, so if your OS numbers are using commas, both inputs and outputs on all screens will use commas. For the input, even though comma would be preferred, a dot will be interpreted as a comma and changed to it, to take into account mismatches between locale and keyboard.

I'm smoke testing a bit and so far so good.

stackchain (approved):

PR #4274, opened 2025-10-14
Update packages to support 16k alignment

lisicky:


Amurmurmur (approved):

mobile/bun.lock
stackchain, line 1: @lisicky I think the bun.lock can be deleted, at least for now

Amurmurmur: we will switch to bun no need to delete.

PR #4276, opened 2025-10-16
fix: prevent multiple displays of operations notice modal

AbdullayevAbdurashid:


mobile/src/features/ReviewTx/useCases/ReviewTxScreen/ReviewTx/Overview/OverviewTab.tsx
```diff
@@ -730,6 +730,7 @@ const useShowOperationsNotice = (operations: Operations) => {
   const {openModal} = useModal()
   const strings = useStrings()
   const screenHeight = useWindowDimensions().height
+  const hasShownModal = React.useRef(false)
```
banklesss: i think this is redundant with useQuery. Maybe the api of useQuery changed in the new version. Can you have a look?

mobile/src/features/ReviewTx/useCases/ReviewTxScreen/ReviewTx/Overview/OverviewTab.tsx
```diff
@@ -730,6 +730,7 @@ const useShowOperationsNotice = (operations: Operations) => {
   const {openModal} = useModal()
   const strings = useStrings()
   const screenHeight = useWindowDimensions().height
+  const hasOpenedInSession = React.useRef(false)
```
banklesss: this is not needed. It needs to be able to update state after the modal is shown and read the value update so the modal its now shown anymore.

banklesss (approved):

PR #4296, opened 2025-10-27
fix(swap): When creating a cancel swap tx, try all adapters if hinted one fails

jorbuedo:
ticket: https://emurgo.atlassian.net/browse/YV-714

---

> [!NOTE]
> Enhances cancel flow to try alternative adapters in parallel when the hinted one fails or returns empty CBOR, returning the first valid CBOR response.
> 
> - **Swap Manager (`mobile/packages/swap/manager.ts`)**:
>   - **cancel()**: Introduces fallback logic.
>     - Tries the hinted adapter first; if response is not Right with non-empty `cbor`, fan out to other adapters in parallel.
>     - Returns the first response with valid `cbor`; otherwise returns the initial response.
> - **Tests (`mobile/packages/swap/manager.test.ts`)**:
>   - Adds comprehensive `cancel()` tests covering:
>     - Delegation per order aggregator (Dexhunter, Muesliswap, Minswap).
>     - Fallback when initial adapter returns empty `cbor` or a Left error.
>     - Behavior when no adapter returns valid `cbor`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit a1139523fe53775533f8a31392126da35c75fadc. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


stackchain (approved):
@jorbuedo lgtm, but I'd impl by going with allSettled right away, and checking if in the return the initial adapter has returned ok, cuz this way it will fire twice against the initial adapter, (which is not a problem)

stackchain:
> @jorbuedo lgtm, but I'd impl by going with allSettled right away, and checking if in the return the initial adapter has returned ok, cuz this way it will fire twice against the initial adapter, (which is not a problem)

I think is wrapped within the api response monad, so going with promise.all right away also would do the job.

PR #4297, opened 2025-10-27
chore(wallet): increase default timer to lock app

jorbuedo:
Comment says should be configurable, and we could definitely add this to user settings. But until that's done, I think 5 minutes is more reasonable than 1 second. Right now it's impossible to switch apps without the whole app locking and loosing where you are. Use cases for this are copy pasting addresses or dreps from an explorer. Or checking seed phrases for recovery from another app.

---

> [!NOTE]
> Increase auto-logout background timer in `AuthProvider` from 1 second to 5 minutes.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 413804ae1d0fa17f26b77bfc140c2994a7125eb4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mobile/src/features/Auth/context/AuthProvider.tsx
+++ b/mobile/src/features/Auth/context/AuthProvider.tsx
@@ -26,7 +26,7 @@ export const AuthProvider: React.FC<React.PropsWithChildren<Props>> = ({
 
   // NOTE: This should be configurable
   useBackgroundTimer({
-    after: time.minutes(1),
+    after: time.minutes(5),
     execute: () => {
       if (loggedState.status === 'logged-in') {
         logger.debug('logout (auto)', {origin: 'AuthProvider', type: 'user'})
--- a/mobile/src/features/Auth/context/AuthProvider.tsx
+++ b/mobile/src/features/Auth/context/AuthProvider.tsx
@@ -26,7 +26,7 @@ export const AuthProvider: React.FC<React.PropsWithChildren<Props>> = ({
 
   // NOTE: This should be configurable
   useBackgroundTimer({
-    after: time.minutes(5),
+    after: time.minutes(1),
     execute: () => {
       if (loggedState.status === 'logged-in') {
         logger.debug('logout (auto)', {origin: 'AuthProvider', type: 'user'})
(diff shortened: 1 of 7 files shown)


mobile/src/features/Auth/context/AuthProvider.tsx
```diff
@@ -26,7 +26,7 @@ export const AuthProvider: React.FC<React.PropsWithChildren<Props>> = ({
 
   // NOTE: This should be configurable
   useBackgroundTimer({
-    after: time.seconds(1),
```
stackchain, line 29: oww boy, this was a mistake introduced by me while testing, i believe it was 1m before

mobile/src/features/Auth/context/AuthProvider.tsx
```diff
@@ -26,7 +26,7 @@ export const AuthProvider: React.FC<React.PropsWithChildren<Props>> = ({
...
+    after: time.minutes(5),
```
stackchain, line 29:
```suggestion
    after: time.minutes(1),
```

mobile/src/features/Auth/context/AuthProvider.tsx
stackchain: updated to 1m

banklesss:
in thsi case we can revert https://github.com/Emurgo/yoroi/pull/4281

mobile/src/features/Auth/context/AuthProvider.tsx
jorbuedo: I think it should be more than just 1, copying stuff from explorers or changing OS permissions may take more than a minute for a lot of users. And I don't see much security concern between 1 and 5.

mobile/src/features/Auth/context/AuthProvider.tsx
jorbuedo: I think it should be more than just 1, copying stuff from explorers or changing OS permissions may take more than a minute for a lot of users. And I don't see much security concern between 1 and 5.

stackchain:
> in thsi case we can revert #4281

unless while requesting privileges we want the user to stay logged in

stackchain:
@rahulnr7 this is acutally a bug, it would be nice to include in the next release. cc: @Amurmurmur

mobile/src/features/Auth/context/AuthProvider.tsx
stackchain: agreed

rahulnr7 (approved):

PR #4301, opened 2025-10-28
feature(mobile): implement posthog analytics

banklesss:
This is the implementation of PostHog analytics. The API documentation is [here](https://github.com/Emurgo/yoroi/blob/cd006b93628a82fcd2923bc67a87cfdec8315dc0/mobile/src/features/Analytics/README.md).

Everything is in the Analytics folder in case we decide to create a package to share between codebases. I’ve tried to make it platform-agnostic. Any suggestions are welcome.

---

> [!NOTE]
> Introduce PostHog analytics with typed events and app-wide wiring, add consent UX, refactor status bar handling, and update env/config/dependencies.
> 
> - **Analytics**:
>   - Add `features/Analytics` module: `AnalyticsRootProvider`, PostHog client adapter, typed events/enums, route-to-event map, `useAnalyticsTracking`, and README.
>   - Consent: analytics notice and settings toggle; persist consent via storage.
> - **App Integration**:
>   - Wire tracking across flows: ReviewTx (view/submit with asset props), Send, Swap (review/cancel), Staking (delegate/withdraw), Governance, DApp connector, Register Catalyst, Manage Collateral, Pool Transition, Portfolio, Notifications (buy banner), Tx History (exchange).
>   - Extend `ReviewTxRoutes` with `context` and pass through call sites.
>   - Enhance `useFormattedTx` to expose `areTokenInfosLoaded` for gated analytics.
> - **Navigation/Status Bar**:
>   - Replace `useStatusBar` with `applyStatusBarForRoute`; update `RouterContainer` to apply status bar and support `onRouteChange`; remove old hook/tests.
> - **Config/Dependencies**:
>   - Add `posthog-react-native` and Expo modules; update `eas.json` to include `EXPO_PUBLIC_POSTHOG_KEY/HOST` for dev/preview/prod; minor version adjustments.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7c9441a428aa2b28557f2c19e2287c68334edaa4. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mobile/PlatformShell.tsx
```diff
@@ -6,30 +7,71 @@ import {
...
 import {RouterContainer} from '~/kernel/navigation/RouterContainer'
+import {
+  initInstallationId,
+  metricsEnabledStorageKeyManager,
+} from '~/kernel/storage/storages'
 import {ModalProvider} from '~/ui/Modal/context/ModalContext'
 
 import {BackgroundTimerProvider} from './src/hooks/BackgroundTimerContext'
 
+function AnalyticsInitializer({children}: React.PropsWithChildren) {
```
jorbuedo: This looks more like a hook than a component, and should probably be declared inside the analytics feature. Since it depends on useAnalyticsContext and in order to not need to be a component child of the provider, it could probably be just done inside the same useAnalyticsContext hook.

mobile/src/features/Analytics/hooks/usePageViewTracking.ts
```diff
@@ -0,0 +1,49 @@
...
+    React.useCallback(() => {
+      trackEvent(event, properties)
+    }, [trackEvent, event, properties]),
+  )
+}
+
+/**
+ * Hook for tracking page views with a delay (useful for screens with loading states)
+ */
+export function useDelayedPageViewTracking<T extends AnalyticsEvent>(
```
jorbuedo: Seems redundant to have 2 hooks doing the same, could just add the optional delay to the main one.

mobile/PlatformShell.tsx
banklesss: its a way to generate the client in a platform agnostic way. Its intended.

mobile/src/features/Analytics/hooks/usePageViewTracking.ts
banklesss: good point. thanks

mobile/src/kernel/navigation/types.tsx
```diff
@@ -242,6 +242,7 @@ export type ReviewTxRoutes = {
     receiverCustomTitle?: React.ReactNode
     details?: ReviewDetailsProps
     createdBy?: React.ReactNode
+    context?: string
```
jorbuedo: I see you've added these context all over the place, but I'm not sure what it does since `navigateToTxReview` hasn't been modified to use it?

banklesss: yes its updated https://github.com/Emurgo/yoroi/pull/4301/files#diff-fbbb230673d8f05c3a2fd43278ed9918a42ac9c4da5497ed44008a1022c03ef5R87

mobile/src/features/Swap/useCases/CreateOrder/SwapMainScreen.tsx
```diff
@@ -35,6 +36,9 @@ export const SwapMainScreen = () => {
   const swapForm = useSwap()
   const {openModal, closeModal} = useModal()
   const navigateTo = useNavigateTo()
+
+  usePageViewTracking('Swap Initiated')
```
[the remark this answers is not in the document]
banklesss: ignore. the events its described in this way

mobile/src/features/ReviewTx/useCases/ReviewTxScreen/ReviewTxScreen.tsx
```diff
@@ -1,20 +1,50 @@
...
+        asset_ticker: asset.tokenInfo.ticker ?? asset.tokenInfo.name ?? '',
+      })
+    }
+  })
+
+  return {
+    type: context ?? '',
+    asset_count: uniqueAssets.size,
+    asset_list: JSON.stringify(Array.from(uniqueAssets.values())),
+  }
```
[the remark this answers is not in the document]
banklesss: fixed already

mobile/PlatformShell.tsx
jorbuedo: I don't know what that means. Platform agnostic to what, iOS/Android? Or Posthog/Amplitude/SomeFutureAlternative?

- The component is declared in the middle of PlatformShell, but it's an analytics component. So it should be declared inside the Analytics feature.
- It returns nothing but the children, so it's not really a component, just an effect, so it should be just a hook.
- Depends on an existing hook (useAnalyticsContext) so in order to avoid having to have a tree, the logic can just be added to that same hook. 
- Alternatively if you don't want to merge the logic hook and the context one, just add the same logic to the provider itself.

banklesss: The idea is that since metrics needs to be initialized, you can import posthog there for React Native, just as is currently being done.

If you decide to move analytics to a package; extension or ctrl wallet can use its posthog dependency to inject its own client in a similar way.

banklesss: so `posthog-react-native` is used in RN code. not in the package in case we finally decide to create a package

mobile/src/features/Analytics/helpers/createPosthogClient.ts
```diff
@@ -0,0 +1,27 @@
...
+    capture: (event, properties) => sdk.capture(event, properties),
+    install: (campaign, source) => {
+      if (currentUserId) sdk.identify(currentUserId, {campaign, source})
+    },
+    identify: (userId, traits) => {
+      currentUserId = userId
+      if (userId) sdk.identify(userId, traits)
+    },
+  }
+}
```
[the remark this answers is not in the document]
banklesss: fixed already

mobile/src/features/Analytics/hooks/usePageViewTracking.ts
```diff
@@ -0,0 +1,49 @@
...
+  useFocusEffect(
+    React.useCallback(() => {
+      const timer = setTimeout(() => {
+        trackEvent(event, properties)
+      }, delay)
+
+      return () => clearTimeout(timer)
+    }, [trackEvent, event, properties, delay]),
+  )
+}
```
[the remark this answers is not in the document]
banklesss: fixed

mobile/PlatformShell.tsx
banklesss: i will refactor to avoid that wrapper

banklesss: refactored

mobile/PlatformShell.tsx
```diff
@@ -6,30 +7,66 @@ import {
...
+
+function usePosthogClient(enabled: boolean) {
+  const installationId = React.useMemo(() => initInstallationId(), [])
+
+  const client = React.useMemo(() => {
+    const apiKey = process.env.EXPO_PUBLIC_POSTHOG_KEY
+    const host = process.env.EXPO_PUBLIC_POSTHOG_HOST
+    if (!apiKey || !host) return null
+    const sdk = new PostHog(apiKey, {host, disabled: false})
+    return createPosthogClient({sdk})
```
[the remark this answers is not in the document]
banklesss: fixed

mobile/src/features/Analytics/helpers/createNoopClient.ts
Amurmurmur, line 1: thats some AI cringe ))

mobile/src/features/Analytics/helpers/createPosthogClient.ts
```diff
@@ -0,0 +1,34 @@
...
+export function createPosthogClient({sdk}: {sdk: SDK}): AnalyticsProvider {
+  let currentUserId: string | undefined
+  return {
+    navigate: (to) => sdk.capture('navigate', {to}),
+    capture: (event, properties) => sdk.capture(event, properties),
+    install: (campaign, source) => {
+      if (currentUserId) sdk.identify(currentUserId, {campaign, source})
+    },
+    identify: (userId, traits) => {
+      const hasValidUserId = typeof userId === 'string' && userId.trim() !== ''
```
Amurmurmur: better just check for length straight away rather than for '' after trimming. Pure AI cringe :)

mobile/src/features/Analytics/helpers/createPosthogClient.ts
```diff
@@ -0,0 +1,34 @@
...
+      if (hasValidUserId) {
+        currentUserId = userId
+        sdk.identify(userId, traits)
+        return
+      }
+      currentUserId = undefined
+      if (sdk.reset) sdk.reset()
```
Amurmurmur: why do we want to reset in the identify?

mobile/src/features/Analytics/hooks/usePageViewTracking.ts
```diff
@@ -0,0 +1,57 @@
...
+  React.useEffect(() => {
+    delayRef.current = delay
+  }, [delay])
+
+  useFocusEffect(
+    React.useCallback(() => {
+      const d = delayRef.current ?? 0
+      if (d > 0) {
+        const timer = setTimeout(() => {
+          trackEvent(
```
Amurmurmur: why do you do this through timeout?

mobile/src/features/Discover/useCases/SelectDappFromList/SelectDappFromListScreen.tsx
```diff
@@ -40,6 +41,8 @@ export const SelectDappFromListScreen = () => {
   )
   const [isShowedWelcomeDApp] = useShowWelcomeDApp()
 
+  usePageViewTracking('Discover Page Viewed')
```
Amurmurmur: for page view tracking.
@banklesss just look at the _layout example in the ctrl repo. you just need to subscribe to the global navigation listener and it will let you track every screen, rather than adding this into every screen. Please remove it from every screen.

Amurmurmur (changes requested):

mobile/src/features/Analytics/helpers/createPosthogClient.ts
banklesss: identity cleared if the user is invalid. it returns the SDK to an anonymous state

mobile/PlatformShell.tsx
```diff
@@ -6,30 +7,67 @@ import {
...
+
+function usePosthogClient(enabled: boolean) {
+  const installationId = React.useMemo(() => initInstallationId(), [])
+
+  const client = React.useMemo(() => {
+    const apiKey = process.env.EXPO_PUBLIC_POSTHOG_KEY
+    const host = process.env.EXPO_PUBLIC_POSTHOG_HOST
+    if (!apiKey || !host) throw new Error('Analytics client is not configured')
+    const sdk = new PostHog(apiKey, {host, disabled: !enabled})
+    return createPosthogClient({sdk})
```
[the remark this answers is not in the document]
banklesss: is expected

mobile/src/features/Analytics/hooks/usePageViewTracking.ts
banklesss: removed the hook entirely

mobile/src/features/Discover/useCases/SelectDappFromList/SelectDappFromListScreen.tsx
banklesss: done

mobile/src/features/Analytics/helpers/createPosthogClient.ts
banklesss: done

mobile/src/features/Analytics/helpers/createNoopClient.ts
banklesss: removed

mobile/src/features/Analytics/helpers/createPosthogClient.ts
Amurmurmur: I think it will do that anyway, if you dont provider currentUserId it will come up with a random one byitself

banklesss: i will remove it

mobile/.env.production
```diff
@@ -7,3 +7,6 @@ EXPO_PUBLIC_FRONTEND_FEE_ADDRESS_MAINNET=addr1q9ry6jfdgm0lcrtfpgwrgxg7qfahv80jlg
 EXPO_PUBLIC_FRONTEND_FEE_ADDRESS_PREPROD=addr_test1qrgpjmyy8zk9nuza24a0f4e7mgp9gd6h3uayp0rqnjnkl54v4dlyj0kwfs0x4e38a7047lymzp37tx0y42glslcdtzhqzp57km
 
 EXPO_PUBLIC_UNSTOPPABLE_API_KEY=czsajliz-wxgu6tujd1zqq7hey_pclfqhdjsqolsxjfsurgh
+
```
Amurmurmur: Remove this file all together we dont have it in dev anymore

mobile/src/kernel/navigation/common/helpers.tsx
```diff
@@ -134,7 +134,8 @@ const getFocusedRouteName = (
 export const getCurrentRouteName = (
   state: NavigationState,
 ): string | undefined => {
-  return state.routes[state.index]?.name
+  const names = getFocusedRouteName(state)
```
Amurmurmur: replace this getCurrentRouteName
```
const navigation = useNavigation();
const currentRoute = navigation.getCurrentRoute();
```

mobile/src/features/Analytics/helpers/createPosthogClient.ts
banklesss: done

mobile/src/kernel/navigation/RouterContainer.tsx
```diff
@@ -14,7 +14,11 @@ const prefixes = [...supportedPrefixes]
 const navRef =
   React.createRef<NavigationContainerRef<ReactNavigation.RootParamList>>()
 
-export function RouterContainer({children}: React.PropsWithChildren) {
+type Props = React.PropsWithChildren<{
+  onRouteChange?: (routeName: string | undefined) => void
+}>
+
+export function RouterContainer({children, onRouteChange}: Props) {
```
Amurmurmur, line 21: TrackedRouterContainer - you dont need this.
Just subscribe to router changes here is an example:
https://docs.expo.dev/router/reference/screen-tracking/

You dont need nothing else, just put it in the root layout

mobile/PlatformShell.tsx
```diff
@@ -6,30 +7,97 @@ import {
...
+  }, [enabled])
+
+  React.useEffect(() => {
+    if (installationId && enabled) client.identify(installationId)
+  }, [client, installationId, enabled])
+
+  return client
+}
+
+function TrackedRouterContainer({children}: React.PropsWithChildren) {
```
Amurmurmur, line 77: TrackedRouterContainer - you dont need this.
Just subscribe to router changes here is an example:
https://docs.expo.dev/router/reference/screen-tracking/

You dont need nothing else, just put it in the root layout

Amurmurmur (changes requested):

mobile/src/kernel/navigation/RouterContainer.tsx
banklesss: we use react-navigation @Amurmurmur

mobile/PlatformShell.tsx
banklesss: we use react-navigation @Amurmurmur

mobile/src/kernel/navigation/common/helpers.tsx
banklesss: refactored

mobile/.env.production
banklesss: done

Amurmurmur:
great stuff @banklesss only question is if you need to extend to have another provider like appsflyter with your current setup how would you do it?

Amurmurmur (approved):

banklesss:
> great stuff @banklesss only question is if you need to extend to have another provider like appsflyter with your current setup how would you do it?

@Amurmurmur i would create a `createAppsFlyerClient` similar to `createPosthogClient` to generate a client with the needed interface.

then I would modify the AnalyticsRootProvider to accept multiple clients with that interface and trigger events for all clients.

finallly I would inject Posthog and AppsFlyer clients from `PlatformShell`, (in React Native envarionment in this case)

Of course everything is open to discussion, and I would appreciate your opinion on this when this improvement is implemented in the following sprints.

PR #4303, opened 2025-10-29
feat(AboutScreen): fetch FCM token conditionally for nightly builds

AbdullayevAbdurashid:
https://emurgo.atlassian.net/browse/YV-738?atlOrigin=eyJpIjoiY2MwYmI3OTcxNGZkNDIyNWJlYWExNmU0MjAwMjU5MjUiLCJwIjoiaiJ9

### DIFF
--- a/mobile/src/features/Settings/ui/screens/ChangeApplicationSettingsScreen/AboutScreen/AboutScreen.tsx
+++ b/mobile/src/features/Settings/ui/screens/ChangeApplicationSettingsScreen/AboutScreen/AboutScreen.tsx
@@ -6,7 +6,7 @@ import {Text, TouchableHighlight, View} from 'react-native'
 import {SafeAreaView} from 'react-native-safe-area-context'
 
 import {usePromise} from '~/hooks/usePromise'
-import {appVersion, commit} from '~/kernel/constants'
+import {appVersion, commit, isNightly} from '~/kernel/constants'
 import {useStrings} from '~/kernel/i18n/useStrings'
 import {Copiable} from '~/ui/Copiable/Copiable'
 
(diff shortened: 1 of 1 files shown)


jorbuedo:
I don't think this makes sense.
From what the code looks like now, it should be showing the FCM already, but it's not. So it clearly doesn't work.
Putting it behind isNightly will just make it show even less. The ticket said to put it in nightly because it assumed this wasn't implemented, so it was like "add it just for nightly so we can test it".
But the first thing would be to actually fix it.

Also, if you're adding a nightly flag, add the isDev too, otherwise won't see anything in development.

AbdullayevAbdurashid:
> I don't think this makes sense. From what the code looks like now, it should be showing the FCM already, but it's not. So it clearly doesn't work. Putting it behind isNightly will just make it show even less. The ticket said to put it in nightly because it assumed this wasn't implemented, so it was like "add it just for nightly so we can test it". But the first thing would be to actually fix it.
> 
> Also, if you're adding a nightly flag, add the isDev too, otherwise won't see anything in development.

 The FCMdevice token can’t be obtained on the iOS Simulator (and Android Emulator requires a Google Play image). You need a physical device with notification permission granted to see a real token. Given that, I doubt  this UI will still be usefull

jorbuedo:
> The FCMdevice token can’t be obtained on the iOS Simulator (and Android Emulator requires a Google Play image). You need a physical device with notification permission granted to see a real token. Given that, I doubt this UI will still be usefull

Is that something new? It used to work fine on the emulator

stackchain:
> I don't think this makes sense. From what the code looks like now, it should be showing the FCM already, but it's not. So it clearly doesn't work. Putting it behind isNightly will just make it show even less. The ticket said to put it in nightly because it assumed this wasn't implemented, so it was like "add it just for nightly so we can test it". But the first thing would be to actually fix it.
> 
> Also, if you're adding a nightly flag, add the isDev too, otherwise won't see anything in development.

@AbdullayevAbdurashid I agree with @jorbuedo the fix quite simple here, just display if FCMToken != null and is not production

AbdullayevAbdurashid:
While I agree with that part the issue is now that FCM token is not showing at all because firebase is not configured neither for iOS nor for Android

PR #4333, opened 2025-11-07
feature(wallet): Dev mode improvements

jorbuedo:
Sorry there's a couple of actual minor bugs fixed as well, can move them out if needed

---

> [!NOTE]
> Introduces a toggleable runtime dev mode and gates dev features with it, adds a wallet-name override to the dapp connector and a modal to review custom CBOR transactions, plus small fixes and UI tweaks.
> 
> - **Dev Mode**
>   - Add `isAuthDev` and `toggleDevMode` to `AuthProvider` and expose via context.
>   - Implement hidden tap gesture in `AboutScreen` to toggle dev mode with on-screen feedback.
>   - Gate dev-only routes/features by `isAuthDev` (e.g., `AppNavigator` dev screens, Menu `UtxoList`, `SelectWalletFromList` Dev options, CBOR copy in `ReviewTxNavigator`).
>   - Replace `__DEV__`/feature flag usages with `isAuthDev`/`isDev` where appropriate (Sentry spotlight, trusted apps, deep-link sandbox).
> - **DApp Connector Enhancements**
>   - Add `WalletNameOverrideProvider`/hook and wrap `BrowserProvider` with it.
>   - New `WalletNameOverrideModalContent` and Dev Menu entry to set override.
>   - Pass overridden name to connector via `getWalletConnectorScript`.
> - **Custom Transaction Review**
>   - Add `CborReviewModalContent` and Dev Menu button to navigate to TX review with pasted CBOR, including warning details.
> - **Fixes & UI Tweaks**
>   - Fix unverified dapp flow using a ref to decide opening the main confirmation modal (`useConfirmConnection`).
>   - Properly hide/restore bottom tab bar styling in `BrowseDappScreen`.
>   - Use `ScrollView` in `DevMenu` for long content.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 419dcc4674c5cc4dc815d692706316b74a160524. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


mobile/src/features/Discover/common/useConfirmConnection.ts
```diff
@@ -42,17 +42,17 @@ export const useConfirmConnection = () => {
         }
 
         if (!selectedDapp) {
-          let shouldOpenMainModal = false
+          const shouldOpenMainModalRef = {current: false}
```
stackchain, line 45: I think this should be actually a ref.

mobile/src/features/Discover/common/WalletNameOverrideContext.tsx
stackchain, line 1: does it need a provider for it? it looks like should be a helper (common has a truncateString)

mobile/src/features/ReviewTx/ReviewTxNavigator.tsx
```diff
@@ -32,7 +33,7 @@ export const ReviewTxNavigator = () => {
         options={({route}) => ({
           title: strings.txReview.title,
           headerRight: () =>
-            route.params?.cbor != null && isDev ? (
+            route.params?.cbor != null && isAuthDev ? (
               <Copiable text={route.params.cbor} />
```
stackchain, line 37: i think this should be able in prod, it is a nice feature (the UI is not good though)

stackchain: i meant not only for devs ofc

mobile/src/features/Discover/common/useConfirmConnection.ts
jorbuedo: Sort of, I think the idea here is sending an object with the value (current) inside, so that the closure passed to the modal access the correct reference instead of passing it a value that doesn't update.
But not sure an actual React reference is needed. It would do the same though.

jorbuedo: Basically the bug right now is that `onClose` and `onConfirm` are each receiving a different copy of shouldOpenMainModal because it's just a value. By moving the value inside an object, they both can reference the inside value instead.

mobile/src/features/ReviewTx/ReviewTxNavigator.tsx
jorbuedo: I agree, but until we have a prod ready UI, I think it's good just being able to activate it in prod, that's why I'm adding the isAuthDev mode thing.

mobile/src/features/Discover/common/WalletNameOverrideContext.tsx
jorbuedo: It needs to be accessed in the dapp explorer, the context is basically to hold the state and being able to write to it from the dev menu. If the override was directly inside the dapp explorer area, then it wouldn't use a context. But I couldn't think of a good place there to put the input.

jorbuedo: And the slice is just so that users don't inject there a huge chunk of text, 15 chars should be enough basically. I don't want a truncate that adds ... or anything.

stackchain (approved):
lgt m

PR #4506, opened 2026-04-02
unbiased dRep list

yushih:
> [!NOTE]
> **Medium Risk**
> Adds a new governance API surface and multiple new navigation screens that build and submit delegation transactions, increasing risk of UX/regression and incorrect data mapping from backend metadata.
> 
> **Overview**
> Adds support for browsing and delegating to *active DReps* instead of manually entering an ID. This introduces a new `getActiveDreps(page,pageSize)` governance endpoint (with network config updates) and sanitization logic to normalize CIP-119/JSON-LD metadata into a stable `ActiveDRepEntry` shape.
> 
> Updates governance navigation to include new `staking-gov-drep-list` and `staking-gov-drep-detail` screens: the list supports search/sort and highlights the currently delegated DRep, while the detail screen shows verified metadata, copyable IDs, and links, and both screens can initiate delegation (including staking key registration when needed). Existing flows (`VotingOptions`, `Home`, `ChangeVote`) now route users to the new DRep list instead of opening the “enter DRep ID” modal, and tests/stubs are updated accordingly.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5b8d76d70ab85b6db1ad07fc411f7d1c8e7434f7. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


yushih (commented):
## PR Review: `unbiased dRep list`

**+1,140 / −306** across 10 files. Overall the approach is solid — clean sanitization, proper pagination, well-structured screens. A few things need attention before this is merge-ready.

---

### 🚨 Blockers

**1. Staging URL hardcoded for mainnet (`config.ts:19`)**
```ts
// DO NOT MERGE!!! NOT FOR PRODUCTION!!!
// getActiveDreps: 'https://zero.yoroiwallet.com/dreps/active',
getActiveDreps: 'https://yoroi-backend-zero-mainnet-staging.emurgornd.com/dreps/active',
```
This is a self-documented blocker. The production URL needs to be uncommented and the staging line removed before this can merge.

---

### ⚠️ Important issues

**2. No tests for `ActiveDRepEntry` sanitizer (`api.ts`)**
The sanitizer is non-trivial — it handles two CIP-119 shapes (`metadata.body.X` vs flat `metadata.X`), JSON-LD `{"@value": "..."}` givenName, nested image path, and references array extraction. Zero test coverage was added. The existing `api.test.ts` file has tests for `getStakingKeyState` and `getDRepById` — the same pattern should be applied to `getActiveDreps` / `sanitizeActiveDrep`, particularly for edge cases.

**3. Unbounded pagination loop (`DrepListScreen.tsx:139`)**
```ts
while (true) {
  const response = await api.getActiveDreps(page, PAGE_SIZE)
  ...
  if (entries.length < PAGE_SIZE) break
  page++
}
```
If the API consistently returns exactly 1000 entries (e.g. due to a bug or a list that hits a threshold boundary), this loops forever and the query never resolves. Add a max page guard:
```ts
const MAX_PAGES = 20
while (page <= MAX_PAGES) { ... }
```

**4. `metadataVerification === 'verified'` is a fragile string literal (`DrepDetailScreen.tsx:61`)**
The value `"verified"` is used as the gating condition to show/hide metadata, but it's undocumented — what are the actual API values? If the API returns `"true"`, `"1"`, or changes the casing, everything silently shows as "unverified". Add a constant or at minimum a comment referencing the API contract.

---

### 🔧 Minor issues

**5. Redundant DRep name display in `DrepDetailScreen`**
The screen title is set to `route.params.name` in the navigator, and the content also renders it in a `heading_3_medium`. This shows the name twice — once in the nav bar and once in the body header. Per the Figma design (node 4533:9698), the nav bar carries the title; the body starts directly with the connection details. The avatar-plus-name header row should be removed from the body.

**6. `isCopying` shared across all `IdRow` instances**
`useCopy()` returns a single context-level `isCopying` flag. Copying either ID row will briefly flash the "copied" checkmark on both copy icons simultaneously. This matches the existing behavior in `OtherDrepCard` so it's consistent — just worth awareness.

**7. Social media URL truncated without visual indicator (`DrepDetailScreen:246`)**
`numberOfLines={1}` on the URL will silently clip long URLs. Add `ellipsizeMode="tail"` to make the truncation explicit.

**8. No empty-state when verified DRep has no metadata sections**
If `isVerified === true` but all three of `objectives`, `motivations`, `qualifications` are empty, `MetadataSections` returns `null` and the area below the divider is blank. A short "No additional information provided" fallback would be cleaner.

**9. `random` sort re-shuffles on every memo recomputation**
If the user sorts by `random` and then types in the search box, the list re-randomises on every keystroke. A `useRef` to freeze the shuffled order (reset only when sort method changes) would give more stable UX.

---

### ✅ What's good

- Sanitizer correctly handles both flat and nested CIP-119 metadata shapes, and the JSON-LD `{"@value": "..."}` givenName format
- Delegation logic is self-contained in both `DrepListScreen` and `DrepDetailScreen` — no callback prop drilling
- Currently-delegated DRep is pinned to position 0 regardless of sort, with gradient highlight and disabled DELEGATE button
- `staleTime: 5 * 60 * 1000` on the DRep list query is appropriate — avoids hammering the endpoint on every navigation
- Error state and loading spinner are handled at the list level

stackchain (approved):

